PR #10, opened 2021-10-29
Improve logging, add logging tests for Lock service

hannesg:
This refactoring has become quite lengthy. I've switched the logging library to zap which has some nice perks when it comes to formatting and testing. Namely the formatter can now be switched to gcp optimized form.

Also, I added log tests to the Lock module. Once this MR is approved, I will add them to the remaining service as well.


chart/values.yaml.tpl
```diff
@@ -8,6 +8,11 @@ git:
 
 hub: eu.gcr.io/freiheit-core
 tag: "$VERSION"
+log:
+  # Other possible values are "gcp" for a gcp-optimized format and "default" for json
```
sven-urbanski-freiheit-com, line 12:
```suggestion
  # Possible values are "gcp" for a gcp-optimized format and "default" for json
```
(if there are actual "other" possible values, then please mention them in the comment)

chart/values.yaml.tpl
```diff
@@ -8,6 +8,11 @@ git:
...
+  format: ""
+  # Other possible values are "DEBUG", "INFO", "ERROR"
```
sven-urbanski-freiheit-com, line 14:
```suggestion
  # Possible values are "DEBUG", "INFO", "WARN", "ERROR"
```

services/cd-service/pkg/cmd/server.go
```diff
@@ -62,106 +64,116 @@ func (c *Config) readPgpKeyRing() (openpgp.KeyRing, error) {
...
 
-	lockSrv := &service.LockServiceServer{
-		Repository: repo,
-	}
-	deploySrv := &service.DeployServiceServer{
-		Repository: repo,
-	}
+		repo, err := repository.New(ctx, repository.Config{
+			URL:            c.GitUrl,
+			Path:           "./repository",
```
sven-urbanski-freiheit-com, line 88: I think we should use c.repository here
(I know you didn't introduce it now :) )

services/cd-service/pkg/service/lock_test.go
```diff
@@ -229,24 +238,90 @@ func TestInvalidCreateEnvironmentApplicationLockArguments(t *testing.T) {
...
+	}{
+		{
+			Name: "create an environment lock",
+			Request: &api.CreateEnvironmentApplicationLockRequest{
+				Environment: "bar",
+				Application: "app",
+				LockId:      "lock",
+			},
+			ExpectedCode:    codes.Internal,
+			ExpectedMessage: "internal error",
```
sven-urbanski-freiheit-com, line 281: hm, could we provide a more meaningful error message here?
It's not clear to me, what is actually tested (or: why do we get this error here)

services/cd-service/pkg/cmd/server.go
hannesg: :+1: let's do that in a seperate MR

services/cd-service/pkg/service/lock_test.go
hannesg: :thinking: reading this now, I think that shouldn't be tested in the test case at all.

services/frontend-service/Makefile
```diff
@@ -38,7 +38,7 @@ bin/main: $(GO_FILES) | proto
 GARBAGE += bin/main
 
 run: bin/main
-	KUBERPULT_CDSERVER=localhost:8443 bin/main
+	KUBERPULT_CDSERVER=172.17.0.3:8443 bin/main
```
sven-urbanski-freiheit-com, line 41: please don't commit this

sven-urbanski-freiheit-com (approved):

services/frontend-service/Makefile
hannesg: Good point! I've made it possible to specify this on the command line

PR #29, opened 2021-11-12
Use Batch Api on the frontend

tameremad:
Now that the batch service is implemented in the back-end, the next step is to use it in the front-end and start sending batch requests. 
* create simple confirmation dialogs that show up every time a batch will be processed
* change the create/delete/deploy actions to use the batch instead of the old dedicated endpoint.
* add notifications that appear after successful/failed actions

### DIFF
--- a/services/frontend-service/src/ui/Batch.tsx
+++ b/services/frontend-service/src/ui/Batch.tsx
@@ -5,19 +5,21 @@ import { Button, Dialog, DialogTitle, IconButton, Typography, Snackbar } from '@
 import { useCallback } from 'react';
 import { Close } from '@material-ui/icons';
 
-export const useBatch = (act: BatchAction, fin?: () => void) =>
-    useUnaryCallback(
-        React.useCallback(
-            (api) =>
-                api
-                    .batchService()
-                    .ProcessBatch({
-                        actions: [act],
-                    })
-                    .finally(fin),
-            [act, fin]
-        )
-    );
+export const callbacks = {
+    useBatch: (act: BatchAction, fin?: () => void) =>
+        useUnaryCallback(
+            React.useCallback(
+                (api) =>
+                    api
+                        .batchService()
+                        .ProcessBatch({
+                            actions: [act],
+                        })
+                        .finally(fin),
+                [act, fin]
+            )
+        ),
+};
 
 export interface ConfirmationDialogProviderProps {
     children: React.ReactElement;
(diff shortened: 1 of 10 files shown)


services/frontend-service/src/ui/ReleaseDialog.tsx
```diff
@@ -265,18 +267,18 @@ const ReleaseEnvironment = (props: {
...
-                api.deployService().Deploy({
-                    application: applicationName,
-                    version: version,
-                    environment: environmentName,
-                    lockBehavior: LockBehavior.Ignore,
-                }),
-            [applicationName, version, environmentName]
-        )
-    );
+    const [deployEnv, deployEnvState] = useBatch({
```
hannesg, line 270: This can become tricky because you use `act` as dependency later. React uses object identity when comparing the dependencies and here you are creating a new object on every render.
I think the solution is to wrap the object in a useMemo like this:
```ts
const act = useMemo(() => +{
        action: {
            $case: 'deploy',
            deploy: {
                application: applicationName,
                version: version,
                environment: environmentName,
                ignoreAllLocks: false,
                lockBehavior: LockBehavior.Ignore,
            },
        },
    }, [applicationName, version, environmentName]);
const [deployEnv, deployEnvState] = useBatch(act);
```
Maybe Viktor can also help you.

services/frontend-service/src/ui/Batch.tsx
```diff
@@ -0,0 +1,59 @@
...
+    version: number;
+    currentlyDeployedVersion: number;
+    deployEnv: () => void;
+    state: string;
+    locked: boolean;
+    prefix: string;
+    hasQueue: boolean;
+}
+
+export const SimpleDialog = (props: SimpleDialogProps) => {
```
hannesg, line 32: Not bad, but I would do that in a seperate PR.

services/frontend-service/src/ui/Batch.tsx
```diff
@@ -0,0 +1,144 @@
...
+    switch (action.action?.$case) {
+        case 'deploy':
+            title = 'Are you sure you want to deploy this version?';
+            notMessageSuccess =
+                'Version ' +
+                action.action?.deploy.version +
+                ' was successfully deployed to ' +
+                action.action?.deploy.environment;
+            notMessageFail = 'Deployment failed';
+            break;
```
sven-urbanski-freiheit-com, line 83: not critical, but I would suggest to move this whole switch into it's own function.
If you return an object with three fields, it would not require so many statements (just expressions).
```
function getMessagesForState(state) {
 case 'deploy':
   return { 
   title: ...
   message: ...
   }
```

sven-urbanski-freiheit-com (approved):

services/frontend-service/src/ui/Batch.tsx
```diff
@@ -0,0 +1,144 @@
...
+        )
+    );
+
+export interface ConfirmationDialogProviderProps {
+    children: React.ReactElement;
+    action: BatchAction;
+    fin?: () => void;
+}
+
+export const ConfirmationDialogProvider = (props: ConfirmationDialogProviderProps) => {
```
sven-urbanski-freiheit-com, line 28: please add some unit tests!

sven-urbanski-freiheit-com (changes requested):
please add unit tests

sven-urbanski-freiheit-com (approved):

PR #32, opened 2021-11-22
Add Batch api logic to the frontend

tameremad:
this is part one of [this PR](https://github.com/freiheit-com/kuberpult/pull/29)

### DIFF
--- a/services/frontend-service/pkg/cmd/server.go
+++ b/services/frontend-service/pkg/cmd/server.go
@@ -88,7 +88,7 @@ func RunServer() {
 
 		mux := http.NewServeMux()
 		mux.Handle("/environments/", grpcProxy)
-		mux.Handle("/batches/", grpcProxy)
+		mux.Handle("/batches", grpcProxy)
 		mux.Handle("/", http.FileServer(http.Dir("build")))
 
 		httpSrv := &setup.CORSMiddleware{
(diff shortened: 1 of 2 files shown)


services/frontend-service/pkg/cmd/server.go
```diff
@@ -79,9 +81,14 @@ func RunServer() {
...
 			logger.FromContext(ctx).Fatal("grpc.deployService.register", zap.Error(err))
 		}
+		err = api.RegisterBatchServiceHandlerServer(ctx, grpcProxy, &gproxy)
+		if err != nil {
+			logger.FromContext(ctx).Fatal("grpc.batchService.register", zap.Error(err))
+		}
 
 		mux := http.NewServeMux()
 		mux.Handle("/environments/", grpcProxy)
+		mux.Handle("/batches/", grpcProxy)
```
hannesg, line 91: You are registering the batch endpoint here as "/batches/" but in the proto file it's "/batches" ( https://github.com/freiheit-com/kuberpult/blob/main/pkg/api/api.proto#L38 ). Which one is this correct?

tameremad: hmm. Why didn't this break the code for me? 
I'll change it here

hannesg (approved):

PR #36, opened 2021-11-24
Improve errors/logs + tests

sven-urbanski-freiheit-com:
Refactored a bit, so that the commit message of a transformer can be tested more easily


tameremad (commented):
Great!

PR #40, opened 2021-11-30
added "undeploy" endpoint

sven-urbanski-freiheit-com:
This endpoint creates a new version that we call "undeploy":
* Creating the undeploy version is similar to createApplicationVersion: it also deploys to envs, if the config is set to upstream.
* The undeploy version works like any other version, except it deploys nothing ("the empty set").
* Trying to deploy the undeploy version if it is already deployed, is essentially a no-op, but is allowed
* This change only touches the backend, so there is no button in the UI to create the undeploy version yet.

The goal is to have a process of undeployment (e.g. removing a microservice), similar to how new services are added: first on one environment, then the next, etc.

### DIFF
--- a/pkg/api/api.proto
+++ b/pkg/api/api.proto
@@ -52,6 +52,7 @@ message BatchAction {
     CreateEnvironmentApplicationLockRequest create_environment_application_lock = 3;
     DeleteEnvironmentApplicationLockRequest delete_environment_application_lock = 4;
     DeployRequest deploy = 5;
+    PrepareUndeployRequest prepare_undeploy = 6;
   }
 }
 
@@ -81,7 +82,6 @@ message DeleteEnvironmentApplicationLockRequest {
 
 service DeployService {
   rpc Deploy (DeployRequest) returns (google.protobuf.Empty) {}
-//  rpc ReleaseTrain (ReleaseTrainRequest) returns (google.protobuf.Empty) {}
   rpc ReleaseTrain (ReleaseTrainRequest) returns (google.protobuf.Empty) {
     option (google.api.http) = {
       put: "/environments/{environment}/releasetrain"
@@ -101,6 +101,13 @@ message DeployRequest {
   string application = 2;
   uint64 version = 3;
   bool ignoreAllLocks = 4 [deprecated = true];
+  LockBehavior lockBehavior = 5;
+}
+
+message PrepareUndeployRequest {
+  string environment = 1;
+  string application = 2;
+  uint64 version = 3;
   LockBehavior lockBehavior = 5;
 }
 
@@ -155,6 +162,7 @@ message Environment {
     map<string, Lock> locks = 3;
     uint64 queuedVersion = 4;
     Commit versionCommit = 5;
+    bool undeployVersion = 6;
   }
 
   string name = 1;
@@ -169,6 +177,7 @@ message Release {
   string source_author = 3;
   string source_message = 4;
   Commit commit = 5;
+  bool undeployVersion = 6;
 }
 
 message Application {
--- a/pkg/api/api.proto
+++ b/pkg/api/api.proto
@@ -105,10 +105,7 @@ message DeployRequest {
 }
 
 message PrepareUndeployRequest {
-  string environment = 1;
-  string application = 2;
-  uint64 version = 3;
-  LockBehavior lockBehavior = 5;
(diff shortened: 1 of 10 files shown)


hannesg (commented):
I'm a bit uncertain how we should mark undeploy commits in the filesystem. My first idea t.b.h would have been to have an empty or absent manifest folder in the release. Maybe that's a bit too implicit indeed. But, I still think there is no need to create and fill the manifests folder for undeploy commits.

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -39,6 +39,98 @@ const (
...
+				&CreateApplicationVersion{
+					Application: "app1",
+					Manifests: map[string]string{
+						envProduction: "productionmanifest",
+					},
+				},
+				&CreateUndeployApplicationVersion{
+					Application: "app1",
+				},
+			},
```
tameremad, line 74: what should happen if we create a new version after the UndeployVersion?

services/cd-service/pkg/service/overview.go
```diff
@@ -108,6 +109,14 @@ func (o *OverviewServiceServer) getOverview(
 							app.QueuedVersion = *queuedVersion
 						}
 					}
+					app.UndeployVersion = false
+					if version != nil {
+						if release, err := s.GetApplicationRelease(appName, *version); err != nil && !errors.Is(err, os.ErrNotExist) {
+							return nil, err
+						} else if release != nil {
+							app.UndeployVersion = release.UndeployVersion
```
tameremad, line 117: does this mean that UndeployVersion would still be displayed in the UI like any other version? 
or it will just be returned with the `GetOverviewResponse`?

tameremad (approved):
It looks really good :+1: , I just have a few questions! :smiley:

pkg/api/api.proto
```diff
@@ -52,6 +52,7 @@ message BatchAction {
     CreateEnvironmentApplicationLockRequest create_environment_application_lock = 3;
     DeleteEnvironmentApplicationLockRequest delete_environment_application_lock = 4;
     DeployRequest deploy = 5;
+    PrepareUndeployRequest prepare_undeploy = 6;
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: I think it's unnecessary. My idea is to get rid of all endpoints besides `batch` at some point (although I don't have any concrete plans yet)

services/cd-service/pkg/repository/transformer.go
```diff
@@ -164,7 +164,83 @@ func (c *CreateApplicationVersion) Transform(fs billy.Filesystem) (string, error
...
+	releaseDir := releasesDirectoryWithVersion(fs, c.Application, lastRelease+1)
+	if err = fs.MkdirAll(releaseDir, 0777); err != nil {
+		return "", err
+	}
+
+	configs, err := (&State{Filesystem: fs}).GetEnvironmentConfigs()
+	if err != nil {
+		return "", err
+	}
+	// this is a flag to indicate that this is the specia "undeploy" version
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: already fixed

services/cd-service/pkg/repository/transformer_test.go
sven-urbanski-freiheit-com: I will add a test for this

services/cd-service/pkg/service/overview.go
sven-urbanski-freiheit-com: for this PR it will be displayed like any other version, however in another change, I will add ui changes that make it more explicit

services/cd-service/pkg/repository/transformer_test.go
sven-urbanski-freiheit-com: It should "just work", so the new version would "win". I added a test

sven-urbanski-freiheit-com:
> I'm a bit uncertain how we should mark undeploy commits in the filesystem. My first idea t.b.h would have been to have an empty or absent manifest folder in the release. Maybe that's a bit too implicit indeed. But, I still think there is no need to create and fill the manifests folder for undeploy commits.

@hannesg please take a look again. I added the discussed finalizer, among other things

hannesg (approved):

PR #43, opened 2021-12-09
Added Undeploy Button and Warnings to the UI

sven-urbanski-freiheit-com:
The warnings will show up if there is an "undeploy" version and if the versions are inconsistent (first undeploy, then normal version).

### DIFF
--- a/services/frontend-service/src/ui/ReleaseDialog.tsx
+++ b/services/frontend-service/src/ui/ReleaseDialog.tsx
@@ -270,34 +270,6 @@ const DeployButton = (props: {
     }
 };
 
-const UndeployButton = (props: {
-    openDialog?: () => void; //
-    state?: string; //
-    applicationName: string; //
-}) => {
-    const buttonMsg = 'Prepare to Undeploy';
-    const tooltipMsg =
-        'This will create a new version that is empty. Use this only for services that are not needed anymore.';
-    switch (props.state) {
-        case 'waiting':
-            return (
-                <Tooltip title={tooltipMsg}>
-                    <Button variant="contained" onClick={props.openDialog}>
-                        {buttonMsg}
-                    </Button>
-                </Tooltip>
-            );
-        case 'pending':
-            return <div>..waiting...</div>;
-        case 'resolved':
-            return <div>Resolved</div>;
-        case 'rejected':
-            return <div>Rejected</div>;
-        default:
-            return <div>Unknown</div>;
-    }
-};
-
 export const CreateLockButton = (props: { applicationName?: string; environmentName: string }) => {
     const { applicationName, environmentName } = props;
     const [messageBox, setMessageBox] = React.useState(false);
@@ -639,23 +611,6 @@ const ReleaseDialog = (props: {
         ''
     );
 
-    const undeployAction: BatchAction = useMemo(
-        () => ({
-            action: {
-                $case: 'prepareUndeploy',
-                prepareUndeploy: {
-                    application: applicationName,
-                },
-            },
-        }),
-        [applicationName]
-    );
-    const undeployButton = (
-        <ConfirmationDialogProvider action={undeployAction}>
-            <UndeployButton applicationName={applicationName} />
-        </ConfirmationDialogProvider>
-    );
-
     return (
         <Dialog open fullWidth={true} maxWidth="lg">
             <DialogTitle className={classes.title}>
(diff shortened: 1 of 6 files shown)


services/frontend-service/src/ui/Warnings.tsx
```diff
@@ -0,0 +1,82 @@
...
+    }
+    if (allUndeploy) {
+        return DeployState.Undeploy;
+    }
+    return DeployState.Mixed;
+}
+
+const UndeployRunningWarning: React.FC<any> = (props: { deployState: DeployState }) => {
+    const tooltip = 'This app is in the process of deletion';
+    const undeployHint = (
```
tameremad, line 36: Please use the `<Tooltip />` component for consistency.
https://github.com/freiheit-com/kuberpult/blob/b77ddec016b456d524039f17cdfd1db31fb18d86/services/frontend-service/src/ui/ReleaseDialog.tsx#L69

services/frontend-service/src/ui/Warnings.tsx
```diff
@@ -0,0 +1,82 @@
...
+import DeleteOutlineIcon from '@material-ui/icons/DeleteOutline';
+import type { Environment, Release } from '../api/api';
+
+export enum DeployState {
+    Normal,
+    Undeploy,
+    Mixed,
+}
+
+export function getDeployState(name: string, environments: { [name: string]: Environment }): DeployState {
```
tameremad, line 12: Maybe we should add some unit tests for `Warnings.tsx`?

services/frontend-service/src/ui/Releases.tsx
```diff
@@ -159,10 +160,13 @@ const ApplicationBox: React.FC<any> = (props: {
     }
     const releases = application.releases;
     releases?.sort((a, b) => b.version - a.version);
-
+    const warnings = <Warnings name={name} environments={environments} releases={releases} />;
     return (
         <TableRow className="application">
-            <TableCell className="applicationCard">{name}</TableCell>
+            <TableCell className="applicationCard">
```
tameremad, line 166: Does it make sense to have a new `TableCell` here with the `Warnings` and the `undeployButton`?

services/frontend-service/src/ui/ReleaseDialog.tsx
```diff
@@ -270,6 +270,34 @@ const DeployButton = (props: {
...
+            return (
+                <Tooltip title={tooltipMsg}>
+                    <Button variant="contained" onClick={props.openDialog}>
+                        {buttonMsg}
+                    </Button>
+                </Tooltip>
+            );
+        case 'pending':
+            return <div>..waiting...</div>;
+        case 'resolved':
```
tameremad, line 292: I don't think this is very ideal. maybe we can display a `disabled` button or just `"undeployment ready"` / `"ready for deletion"` 

also it may be useful to prevent doing this if the latest version is already an undeploy version. => also `disabled` button

tameremad (commented):
I just added a few comments.

services/frontend-service/src/ui/ReleaseDialog.tsx
sven-urbanski-freiheit-com: > also it may be useful to prevent doing this if the latest version is already an undeploy version. => also disabled button

I thought about this, and would like to have it, but it seems rather complicated.
Just because we have multiple envs, some can have undeploy version while others do not. So I'd like to keep this simple for now

sven-urbanski-freiheit-com: > I don't think this is very ideal. maybe we can display a disabled button or just "undeployment ready" / "ready for deletion"

Yes you are right, I will change this

services/frontend-service/src/ui/Warnings.tsx
sven-urbanski-freiheit-com: We have unit tests for that already

services/frontend-service/src/ui/Warnings.tsx
sven-urbanski-freiheit-com: done

services/frontend-service/src/ui/Releases.tsx
sven-urbanski-freiheit-com: done. In the long run, we need a better option though, it looks a bit crowded already.

services/frontend-service/src/ui/ReleaseDialog.tsx
sven-urbanski-freiheit-com: done

PR #45, opened 2021-12-13
Add coloring of releases based on freshness

tameremad:


services/cd-service/pkg/cmd/server.go
```diff
@@ -85,7 +85,7 @@ func RunServer() {
 
 		repo, err := repository.New(ctx, repository.Config{
 			URL:            c.GitUrl,
-			Path:           "./repository",
```
sven-urbanski-freiheit-com, line 88: why do we change the path?
Note that this is a seperate path from "repoistory_remote".
They *should* be different from each other, otherwise, working locally is not the same as working remote.

I had an issue that was really hard to detect, because `Path` and `URL` were the same: Essentially, if they are the same, `git commit` is the same as `git push`, which is very confusing, and can only happen locally.

-> Change back to `./repository` please

services/frontend-service/src/ui/Releases.test.tsx
```diff
@@ -72,36 +86,68 @@ describe('Releases', () => {
...
+
+        // when - 10 days old
+        rerender(getNode({ data: getDummyOverview(new Date(now.valueOf() - 10 * 24 * hr)) }));
+        // then - history
+        expect(container.querySelector('.releases .details-history')).toBeTruthy();
+
+        // when - not specified
+        rerender(getNode());
+        // then - history
+        expect(container.querySelector('.releases .details-history')).toBeTruthy();
```
sven-urbanski-freiheit-com, line 122: please transform this into a table test.
Input is something like: `hoursToSubtract`,
expectedOutput: detailsCssClass (or something like that)

sven-urbanski-freiheit-com (changes requested):
.

services/frontend-service/src/ui/Releases.test.tsx
tameremad: Done!

sven-urbanski-freiheit-com (approved):

PR #47, opened 2021-12-15
Reimplement History.Change in a much faster way

hannesg:
The new algorithm calculates much bigger parts of the tree up-front.

```
/build/services/cd-service $ go test -bench . -benchmem -benchtime 1m ./pkg/history/...
goos: linux
goarch: amd64
pkg: github.com/freiheit-com/kuberpult/services/cd-service/pkg/history
cpu: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz
BenchmarkHistoryNew-8                244         302610217 ns/op        44486087 B/op    2501890 allocs/op
BenchmarkHistoryOld-8                 22        2963958233 ns/op        253691023 B/op  17944738 allocs/op
PASS
ok      github.com/freiheit-com/kuberpult/services/cd-service/pkg/history       228.461s
```

### DIFF
--- a/services/cd-service/pkg/history/history.go
+++ b/services/cd-service/pkg/history/history.go
@@ -25,6 +25,8 @@ import (
 
 type History struct {
 	repository *git.Repository
+	commits    map[git.Oid]*CommitHistory
+	useOldAlg  bool
 }
 
 type NotExists struct {
@@ -46,8 +48,20 @@ var (
 
 // Returns the first commit after a certain path was changed.
 func (h *History) Change(from *git.Commit, path []string) (*git.Commit, error) {
-	commit, _, err := h.traverse(from, path)
-	return commit, err
+	if h.useOldAlg {
+		commit, _, err := h.traverse(from, path)
+		return commit, err
+	}
+	var err error
+	ch := h.commits[*from.Id()]
+	if ch == nil {
+		ch, err = NewCommitHistory(h.repository, from)
+		if err != nil {
+			return nil, err
+		}
+		h.commits[*from.Id()] = ch
+	}
+	return ch.Change(path)
 }
 
 // Returns the first commit after a certain path was changed and the oid of the tree node after the change to aid subsequent iterations.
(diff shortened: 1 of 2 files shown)


services/cd-service/pkg/history/history_test.go
```diff
@@ -294,3 +297,88 @@ func assertChangedAtNthCommit(t *testing.T, actualCommit *git.Commit, expectedPo
...
 	t.Errorf("wrong changed commit, expected %d, actually not any known commit", expectedPosition)
 }
+
+func BenchmarkHistoryNew(b *testing.B){
+	benchmarkHistory(b, false)
+}
+
+func BenchmarkHistoryOld(b *testing.B){
+	benchmarkHistory(b, true)
+}
```
sven-urbanski-freiheit-com, line 307: these functions seem unused.

services/cd-service/pkg/history/history.go
```diff
@@ -136,5 +150,166 @@ func NewHistory(repo *git.Repository) *History {
...
+func (t *treeNode) push(commit *git.Commit, entry *git.TreeEntry) {
+	if entry == nil {
+		entry = &git.TreeEntry{Id: &oidZero}
+	}
+	if t.commit == nil {
+		if !t.entry.Id.Equal(entry.Id) {
+			t.commit = commit
+		}
+	}
+	if t.entry.Filemode == git.FilemodeTree {
```
sven-urbanski-freiheit-com, line 302: what happens if it's not a `FilemodeTree`?

services/cd-service/pkg/history/history_test.go
hannesg: These are the Benchmark methods. They are used when you run test with `-bench .`

services/cd-service/pkg/history/history.go
hannesg: Then it cannot have childnodes and then we don't have to propagate the change to them.

services/cd-service/pkg/history/history_test.go
sven-urbanski-freiheit-com: ahh, nice :)

sven-urbanski-freiheit-com (approved):

PR #57, opened 2022-01-04
add test format GitHub actions SRX-169ZVQ

medo-freiheit:


medo-freiheit:
Somehow the second GitHub action stopped triggering!

PR #67, opened 2022-01-14
 add tracer spans SRX-G6KFIX

medo-freiheit:


services/cd-service/pkg/cmd/server.go
```diff
@@ -91,6 +91,11 @@ func RunServer() {
 			defer tracer.Stop()
 		}
 
+		span, ok := tracer.SpanFromContext(ctx)
```
sven-urbanski-freiheit-com, line 94: is it ok to run this when `EnableTracing=false`?

services/cd-service/pkg/repository/repository.go
```diff
@@ -369,10 +370,21 @@ func (r *repository) Apply(ctx context.Context, transformers ...Transformer) err
 }
 
 func (r *repository) Push(ctx context.Context, pushAction func() error) error {
+
+	span, ok := tracer.SpanFromContext(ctx)
```
sven-urbanski-freiheit-com, line 374: I think `StartSpanFromContext` is actually the better option here, as it creates a new Span, if none exists.

sven-urbanski-freiheit-com:
```suggestion
	span, ctx := tracer.StartSpanFromContext(ctx, "Push")
```
Note that it returns a context which we need to use in the rest of the function

services/cd-service/pkg/cmd/server.go
sven-urbanski-freiheit-com, line 94: I think we need to use StartSpanFromContext here as well

medo-freiheit: According to the documentation: `If the tracer is not started, calling this function is a no-op.`

services/cd-service/pkg/repository/repository.go
```diff
@@ -369,10 +370,16 @@ func (r *repository) Apply(ctx context.Context, transformers ...Transformer) err
...
 func (r *repository) Push(ctx context.Context, pushAction func() error) error {
+
+	span, _ := tracer.StartSpanFromContext(ctx, "Apply")
+	defer span.Finish()
+
 	eb := backoff.NewExponentialBackOff()
 	eb.MaxElapsedTime = 7 * time.Second
 	return backoff.Retry(
 		func() error {
+			span, _ := tracer.StartSpanFromContext(ctx, fmt.Sprintf("try to Push at: %v", eb.GetElapsedTime()))
```
hannesg, line 381: The action should never be dynamic. If you want to add dynamic data to the span, please use `SetTag` like this:

```
span, _ := tracer.StartSpanFromContext(ctx, "push")
span.SetTag("elapsedTime", eb.GetElapsedTime())
```

services/cd-service/pkg/cmd/server.go
```diff
@@ -91,6 +91,8 @@ func RunServer() {
 			defer tracer.Stop()
 		}
 
+		span, _ := tracer.StartSpanFromContext(ctx, "Start server")
```
hannesg, line 94: If you discard the ctx return value, then your span will be propagated through the stack. Please use
```
span, ctx := ...
```

hannesg (changes requested):

sven-urbanski-freiheit-com (approved):

hannesg (approved):

PR #69, opened 2022-01-17
Persist Action Authors in Manifest repo

tameremad:


services/cd-service/pkg/repository/repository.go
```diff
@@ -252,6 +253,13 @@ func (r *repository) ApplyTransformers(ctx context.Context, transformers ...Tran
 		Email: r.config.CommitterEmail,
 		When:  time.Now(),
 	}
+	// Add Author to Commit MSG
+	actionAuthor := auth.Extract(ctx)
+	if actionAuthor != nil {
+		authMsg := fmt.Sprintf("Authored by:\nEmail: %s\nUsername: %s\n", actionAuthor.Email, actionAuthor.Username)
```
hannesg, line 259: Please set the actual git author in CreateCommitFromIds ( it's the 2nd parameter ) because that data is actually transfered to the frontend.

hannesg (changes requested):
Please also add some tests.

services/frontend-service/pkg/auth/auth.go
```diff
@@ -0,0 +1,68 @@
...
+		Email:    "local@freiheit.com",
+		Username: "local@freiheit.com",
+	}
+)
+
+// Extract takes the User from middleware.
+// It always returns a User
+func Extract(ctx context.Context) *User {
+	md, _ := metadata.FromIncomingContext(ctx)
+	u := defUser
```
sven-urbanski-freiheit-com, line 23: what does "def" stand for? I guess "default"? Please then call it "defaultUser"

services/cd-service/pkg/repository/repository.go
```diff
@@ -24,6 +24,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/cenkalti/backoff/v4"
+	"github.com/freiheit-com/kuberpult/services/frontend-service/pkg/auth"
```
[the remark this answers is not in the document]
tameremad: Makes sense :+1:

services/cd-service/pkg/repository/repository.go
tameremad: OK, that is much better. thanks!

services/frontend-service/pkg/auth/auth.go
tameremad: Done!

pkg/auth/auth.go
```diff
@@ -0,0 +1,84 @@
...
+			Name:  md.Get("author-username")[0],
+		}
+	}
+	return u
+}
+
+// ToContext adds the User to the context for extraction later.
+// Returning the new context that has been created.
+func ToContext(ctx context.Context, u *User) context.Context {
+	if u == nil || u.Email == "" {
```
tameremad, line 54: Technically, since `ToContext` will only be used in the `ServeHTTP` function in this file I think the user check should be moved somewhere else? or be completely removed because the user is always given? or do we keep it as best practice?

pkg/auth/auth.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (p *Auth) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	c := r.Context()
+	u := getActionAuthor()
+	p.HttpServer.ServeHTTP(w, r.WithContext(ToContext(c, u)))
+}
+
+func getActionAuthor() *User {
+	// Local
```
tameremad, line 82: In my understanding, The IAP check should be around here? 
I mean, this is where we will get the email and username from right?

pkg/auth/auth_test.go
```diff
@@ -0,0 +1,74 @@
+/*This file is part of kuberpult.
```
tameremad, line 1: Also I think this is in a directory that is not tested along with the other tests in `make test` 
how can we fix this?

pkg/auth/auth.go
hannesg: You can't really remove it from the package because the ctxMarkerKey is only available in this package. I don't think it's an issue to have it here

pkg/auth/auth.go
```diff
@@ -0,0 +1,88 @@
...
+	// if no username was specified, use email as username
+	if u.Name == "" {
+		u.Name = u.Email
+	}
+	ctx = metadata.AppendToOutgoingContext(ctx, "author-email", u.Email, "author-username", u.Name)
+	return context.WithValue(ctx, ctxMarkerKey, u)
+}
+
+// splits of grpc-traffic
+type Auth struct {
```
hannesg, line 70: If you want to move something to the frontend service, it would be this type.

tameremad: Yes, that makes more sense i think :+1:

pkg/auth/auth_test.go
sven-urbanski-freiheit-com: It should just work?
Is this directory in the `MAKEDIRS` variable?

sven-urbanski-freiheit-com (approved):

pkg/auth/auth_test.go
tameremad: No it's in `pkg/auth` but `MAKEDIRS` only has `pkg/api` 

also I just checked the test runner here on github and the `setup_test.go` is not running as well. which is in `pkg/setup`

I'll create a new PR with a new makefile in `pkg` and add that to the `MAKEDIRS`

but for now the tests are successful.

hannesg (approved):

PR #72, opened 2022-01-24
add undeploy app btn SRX-M5GOA4

medo-freiheit:


services/frontend-service/src/ui/Warnings.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+        const defaultProps: any = {
+            children: null,
+            state: 'no state', //
+            applicationName: 'app1', //
+        };
+        return <UndeployBtn {...defaultProps} {...overrides} />;
+    };
+    const getWrapper = (overrides?: { state?: string }) => render(getNode(overrides));
+
+    it('renders the UndeployBtn component', () => {
```
tameremad, line 33: I think it's best to use table driven tests here since they all render the same container only with different states

https://github.com/freiheit-com/kuberpult/blob/76802b82f8c5ebf6065f524567cc1b8ca14f5f36/services/frontend-service/src/ui/Releases.test.tsx#L89-L133

medo-freiheit: I tried to do that at first then I changed it because I need to use different selectors but the fact that I use the same validation function can help here and I can use it

services/cd-service/pkg/service/batch.go
```diff
@@ -139,6 +139,14 @@ func (d *BatchServer) processAction(
 		return &repository.CreateUndeployApplicationVersion{
 			Application: act.Application,
 		}, nil
+	case *api.BatchAction_Undeploy:
+		act := action.Undeploy
+		if err := ValidatePrepareUndeploy(act.Application); err != nil {
```
sven-urbanski-freiheit-com, line 144:
```suggestion
		if err := ValidatePrepareUndeploy(act.Application); err != nil {
```
Either you rename this to `ValidateApplication`, or you make a seperate function that is called `ValidateUndeploy`.
As it is now, this is a bit misleading.

sven-urbanski-freiheit-com (approved):

PR #73, opened 2022-01-25
 fix un-deploy endpoint SRX-M5GOA4

medo-freiheit:


hannesg (approved):
Looks good. Can you add a test that checks that the application is really gone after deleting it?

PR #76, opened 2022-01-31
health check improvement SRX-V33NGL

medo-freiheit:


services/cd-service/pkg/service/service.go
```diff
@@ -253,15 +253,6 @@ func ArgocdLogin(host string, username string, password string) (string, error)
...
-	if ok, err := s.Repository.IsReady(); ok && err != nil {
-		return err
-	}
-	if s.ArgoCdPass != "" {
-		_, er := ArgocdLogin(s.ArgoCdHost, s.ArgoCdUser, s.ArgoCdPass)
-		if er != nil {
-			return er
-		}
-	}
 	return nil
```
sven-urbanski-freiheit-com, line 256: if it just returns `nil`, then consider removing this function

sven-urbanski-freiheit-com (approved):

PR #78, opened 2022-02-03
get rid of "get go" SRX-71OYF2

medo-freiheit:
go get github.com/ghodss/yaml@v1.0.0

is still existing because of the fact that it's ok to use "go get" with non-executable packages, which "go install" cannot build.

this warning is only showing when you are trying to install executable package with "go get".
`go get: installing executables with 'go get' in module mode is deprecated.
	To adjust and download dependencies of the current module, use 'go get -d'.
	To install using requirements of the current module, use 'go install'.
	To install ignoring the current module, use 'go install' with a version,
	like 'go install example.com/cmd@latest'.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.`

### DIFF
--- a/.github/workflows/test-linter-build-push.yml
+++ b/.github/workflows/test-linter-build-push.yml
@@ -30,16 +30,7 @@ jobs:
           sudo make install
 
       - name: Install dependencies
-        # the reason why hardcoded the dependencies instead of running "make .install" is that make .install throws an error related to go mod
-        run: |
-          go install -modfile=go.tools.mod github.com/golang/protobuf/protoc-gen-go google.golang.org/grpc/cmd/protoc-gen-go-grpc && \
-          go get github.com/ghodss/yaml@v1.0.0 && \
-          go install github.com/libgit2/git2go/v33 && \
-          go install \
-            github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway \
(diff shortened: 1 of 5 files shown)


.github/workflows/test-linter-build-push.yml
```diff
@@ -32,16 +32,13 @@ jobs:
...
           go get github.com/ghodss/yaml@v1.0.0 && \
-          go install google.golang.org/grpc/cmd/protoc-gen-go-grpc && \
-          go get github.com/libgit2/git2go/v33 && \
+          go install github.com/libgit2/git2go/v33 && \
           go install \
             github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway \
             github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2 \
             google.golang.org/protobuf/cmd/protoc-gen-go \
-            google.golang.org/grpc/cmd/protoc-gen-go-grpc \
             github.com/grpc-ecosystem/grpc-gateway/v2/internal/descriptor
```
sven-urbanski-freiheit-com, line 42: we need to run this locally anyway, right? Please try to move this to the makefile. Or if that is not possible, explain more why (line 33)

medo-freiheit: Yes, I think that makes more sense

sven-urbanski-freiheit-com (approved):

PR #79, opened 2022-02-04
 add GH action test helm chart SRX-27FCEG

medo-freiheit:
Apparently, KinD recourses limits is 1m for CPU ( in Github action) , so I decreased  the  CPU for Kuberpult to: 
cd-service: 
`limits:
            cpu: "500m"
            memory: "1Gi"
          requests:
            cpu: "500m"
            memory: "1Gi"`
            
frontend-service: * 2
`limits:
            cpu: "250m"
            memory: "100Mi"
          requests:
            cpu: "250m"
            memory: "100Mi"`


.github/workflows/test-helm-chart-release.yml
```diff
@@ -0,0 +1,50 @@
+name: Lint and Test Charts
```
sven-urbanski-freiheit-com, line 1: NOTE to self: Before merging this, make sure that all projects set the values that are NOT required anymore

charts/kuberpult/values.yaml
```diff
@@ -0,0 +1,51 @@
+# Default values for ..
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+git:
+  url: "/repository"  # git@github.com/.../...
```
sven-urbanski-freiheit-com, line 6: please add a comment, eg.
"the url should not have a default value. We only give it a value here, because `kind` in combination with `helm install` does not allow to set values (without crashing at teardown)"

charts/kuberpult/values.yaml.tpl
```diff
@@ -17,13 +17,13 @@ log:
...
-    create: false # Add backend config for health checks on GKE only
+    create: false  # Add backend config for health checks on GKE only
 frontend:
   image: kuberpult-frontend-service
 ingress:
   annotations: {}
-  domainName: null
-  exposeReleaseEndpoint: false
+  domainName: kuberpult.example.com
+  exposeReleaseEndpoint: true
```
sven-urbanski-freiheit-com, line 26: I don't think we need to change the release endpoint.

medo-freiheit: Yes that's true, I just tried it, I will move it back to default

PR #83, opened 2022-02-14
add locks datadog metrics SRX-9KE1DS

medo-freiheit:


services/cd-service/pkg/repository/repository.go
```diff
@@ -118,6 +123,16 @@ func openOrCreate(path string) (*git.Repository, error) {
 // Opens a repository. The repository is initialized and updated in the background.
 func New(ctx context.Context, cfg Config) (Repository, error) {
 	logger := logger.FromContext(ctx)
+
+	isMetricsEnabeld := ctx.Value("EnableMetrics")
+	if isMetricsEnabeld != nil && isMetricsEnabeld.(bool) {
+		var err error = nil
+		ddMetrics, err = statsd.New("127.0.0.1:8125", statsd.WithNamespace("Kuperbult"))
```
sven-urbanski-freiheit-com, line 130: Kuberpult

services/cd-service/pkg/repository/transformer.go
```diff
@@ -465,18 +520,21 @@ type DeleteEnvironmentApplicationLock struct {
...
 			Filesystem: fs,
 		}
-		queueMessage, err := s.ProcessQueue(fs, c.Environment, c.Application)
+		queueMessage, err := s.ProcessQueue(ctx, fs, c.Environment, c.Application)
 		if err != nil {
 			return "", err
 		}
+		if ddMetrics != nil {
+			ddMetrics.Set("app_lock_count", fmt.Sprintf("%d", GetEnvironmentApplicationLocksCount(fs, c.Environment, c.Application)), []string{"app:" + c.Application, "env:" + c.Environment}, 1)
+		}
```
sven-urbanski-freiheit-com, line 537: please extract a function from these 3 lines (including the `if`)

services/cd-service/pkg/repository/transformer.go
```diff
@@ -453,6 +505,9 @@ func (c *CreateEnvironmentApplicationLock) Transform(fs billy.Filesystem) (strin
 			if err := createLock(chroot, c.LockId, c.Message); err != nil {
 				return "", err
 			} else {
+				if ddMetrics != nil {
+					ddMetrics.Set("app_lock_count", fmt.Sprintf("%d", GetEnvironmentApplicationLocksCount(fs, c.Environment, c.Application)), []string{"app:" + c.Application, "env:" + c.Environment}, 1)
+				}
```
sven-urbanski-freiheit-com, line 510: please extract a function from these 3 lines (including the if)

services/cd-service/pkg/repository/repository.go
```diff
@@ -221,10 +236,17 @@ func (r *repository) ApplyTransformersInternal(transformers ...Transformer) ([]s
 }
 
 func (r *repository) ApplyTransformers(ctx context.Context, transformers ...Transformer) error {
-	commitMsg, state, err := r.ApplyTransformersInternal(transformers...)
+	commitMsg, state, err := r.ApplyTransformersInternal(ctx, transformers...)
+	//
 	if err != nil {
 		return err
 	}
+	if ddMetrics != nil {
```
sven-urbanski-freiheit-com, line 244: this line does not need to be here, `UpdateDatadogMetrics` takes care of the check

services/cd-service/pkg/repository/repository.go
```diff
@@ -118,6 +123,16 @@ func openOrCreate(path string) (*git.Repository, error) {
 // Opens a repository. The repository is initialized and updated in the background.
 func New(ctx context.Context, cfg Config) (Repository, error) {
 	logger := logger.FromContext(ctx)
+
+	isMetricsEnabeld := ctx.Value("EnableMetrics")
+	if isMetricsEnabeld != nil && isMetricsEnabeld.(bool) {
```
sven-urbanski-freiheit-com, line 128: can't we read `cfg.EnableMetrics` here? Then we don't need to read it from the context.

services/cd-service/pkg/repository/repository.go
```diff
@@ -118,6 +123,16 @@ func openOrCreate(path string) (*git.Repository, error) {
...
+		var err error = nil
+		ddMetrics, err = statsd.New("127.0.0.1:8125", statsd.WithNamespace("Kuperbult"))
+		if err != nil {
+			logger.Fatal("datadog.metrics.error", zap.Error(err))
+		}
+	}
+
```
sven-urbanski-freiheit-com, line 135: the host/port `127.0.0.1:8125` should come from an environment variable. The value is good as a default though.

sven-urbanski-freiheit-com: Actually, I think if you put `nil`, statsd will read it from an env var, but I'm not sure about this.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -58,18 +63,58 @@ func releasesDirectoryWithVersion(fs billy.Filesystem, application string, versi
...
+	envAppLocksCount := 0
+	appDir := environmentApplicationDirectory(fs, environment, application)
+	locksDir := fs.Join(appDir, "locks")
+	if entries, _ := fs.ReadDir(locksDir); entries != nil {
+		envAppLocksCount += len(entries)
+	}
+	return envAppLocksCount
+}
+
+func UpdateDatadogMetrics(fs billy.Filesystem) error {
```
sven-urbanski-freiheit-com, line 86: even if it's trivial, I would like to have one tests that checks that this function does not crash when ddMetrics is nil.

sven-urbanski-freiheit-com (changes requested):
.

services/cd-service/pkg/repository/repository.go
medo-freiheit: `cfg Config` here is the repo configuration, if I add EnableMetrics to repo configuration then it will work

services/cd-service/pkg/repository/transformer.go
medo-freiheit: I believe it will return immediately with nil if ddMetrics is nil

services/cd-service/pkg/repository/repository.go
medo-freiheit: According to [DDStatsD](https://docs.datadoghq.com/developers/dogstatsd/?tab=hostagent#setup) DogStatsD is enabled by default over UDP port 8125 for Agent v6+, but it can be configured so I will add an env variable for it

services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: Yes. I want to ensure that nobody removes that line by accident.

sven-urbanski-freiheit-com: (but I'm also ok if you do that later)

sven-urbanski-freiheit-com (approved):

PR #84, opened 2022-02-14
 new release 0.4.4

medo-freiheit:


charts/kuberpult/values.yaml
```diff
@@ -61,6 +61,9 @@ argocd:
 datadogTracing:
   enabled: false
   debugging: false
-  metrics: false
+
+dogstatsdMetrics:
+  enabled: false
+  address: "127.0.0.1:8125"
```
fdcds, line 67: I would add a link to the documentation here, to explain why port 8125. Explaining that here in text would be nice for someone like me, but I would consider it not critical.

```suggestion
  # dogstatsD listens on port udp:8125 by default.
  # https://docs.datadoghq.com/developers/dogstatsd/?tab=hostagent#agent
  address: "127.0.0.1:8125"
```

charts/kuberpult/values.yaml
```diff
@@ -61,6 +61,9 @@ argocd:
 datadogTracing:
   enabled: false
   debugging: false
-  metrics: false
```
fdcds, line 64: Q: This was unused before?

fdcds (commented):
LGTM, but I am not familiar with the code yet, so I feel not comfortable to actually approve this.

I have one question and a small suggestion.

charts/kuberpult/values.yaml
medo-freiheit: it's related to my new story, I added it but didn’t adapt the helm chart with it, I did that in this PR so I just reorganized it

charts/kuberpult/values.yaml
medo-freiheit: Fair enough, I will add this

PR #87, opened 2022-02-17
Fix import order in cmd package

fdcds:


services/frontend-service/pkg/cmd/server.go
```diff
@@ -18,6 +18,8 @@ package cmd
 
 import (
 	"context"
+	"net/http"
```
medo-freiheit, line 21: why we care about import order ?

medo-freiheit (approved):

services/frontend-service/pkg/cmd/server.go
sven-urbanski-freiheit-com: Can we get a lint/autoformatter that fixes this automatically?

fdcds: In sre-standards we even have a linter for that.

There is no technical difference between the previous and the order in this PR, as far as I know. The difference is just to the reader: Golang standard library imports are in the top block, everything else is in the bottom block, everything is alphabetically ordered.

sven-urbanski-freiheit-com: There's one more thing to it:
if the order is fixed, then we can't have meaningless diffs, were the order is changed accidentally, or because of different preferences/settings. This keeps the history clean and meaningful.

PR #99, opened 2022-02-24
SRX-1QG56B - Add warning box when deploying with locks

tameremad:


services/frontend-service/src/ui/Batch.tsx
```diff
@@ -99,6 +100,25 @@ export const ConfirmationDialogProvider = (props: ConfirmationDialogProviderProp
         </IconButton>
     );
 
+    let locksWarning = '';
+    locks?.forEach((lock) => {
+        locksWarning += '- Lock ID: ' + lock[0] + ' | Message: ' + lock[1].message + '\n';
```
sven-urbanski-freiheit-com, line 105: this is not very "html" like.
I'd rather have an array here and render an array of textfields.
Something like
`locks.map(lock) => <TextField ....`

however, this is a bit nitpicky, I'm also fine with merging it as it is.

sven-urbanski-freiheit-com (approved):

services/frontend-service/src/ui/Batch.tsx
tameremad: done!

PR #101, opened 2022-02-25
SRX-1AZDH9 - Add docker image for kuberpult dependencies

tameremad:


mouhsen-ibrahim (approved):
Remove the non-docker workflow

medo-freiheit (changes requested):
.

.github/workflows/docker-run-tests.yml
```diff
@@ -0,0 +1,22 @@
...
+    container: europe-west3-docker.pkg.dev/fdc-public-docker-registry/kuberpult/kuberpult-dependencies:0.1.1
+    env:
+      LD_LIBRARY_PATH: '/usr/local/lib'
+      CI: true
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install go dependencies and run tests (charts are not tested here)
+        env:
+          HOME: /root # set the HOME variable to access the real GOPATH. See: https://github.com/actions/runner/issues/1146
+        run: make .install && make test
```
medo-freiheit, line 22:
```suggestion
        run: make test
```
I like it!  but the main proposal was to not make .install every time, especially libgit2 it takes a long time, all go deps should be installed in the docker image, if that is not possible for some of them you can add a step for them here

tameremad: yes Libgit is installed in the image. these are only go dependencies sadly they are still large. 

https://github.com/freiheit-com/kuberpult/blob/a29afb79ae4c7ac22622497dcb4f9827bdeaf243/Makefile#L27-L36

mouhsen-ibrahim: `make .install` installs the go libraries needed for kuberpult to run, these could change at any time. The image only contains libgit2 and buf with the tools needed to compile and test kuberpult.

medo-freiheit: I was miss-leaded, the other action was renamed and had 'docker' at the end its name *o*, 
Yeah, you're right it's what we want, sorry for the confusion!

medo-freiheit (approved):

PR #103, opened 2022-03-02
SRX-K20BI8 - Add conflict detection and warning

tameremad:


services/frontend-service/src/ui/Batch.tsx
```diff
@@ -133,9 +206,14 @@ export const ConfirmationDialogProvider = (props: ConfirmationDialogProviderProp
...
                 <div style={{ margin: '16px 24px' }}>{GetActionDetails(action).description}</div>
                 {deployLocks}
+                {conflictMessage}
                 <span style={{ alignSelf: 'end' }}>
                     <Button onClick={handleClose}>Cancel</Button>
-                    <Button onClick={closeWhenDone}>Add to cart</Button>
+                    {!conflict ? (
+                        <Button onClick={closeWhenDone}>Add to cart</Button>
+                    ) : (
+                        <Button onClick={closeWhenDone}>Add anyway</Button>
```
medo-freiheit, line 215: Maybe ??

```suggestion
                        <Button onClick={closeWhenDone}>{ conflict ? "Add anyway" : "Add to cart"  }</Button>
```

services/frontend-service/src/ui/Batch.tsx
```diff
@@ -56,9 +68,55 @@ export interface ConfirmationDialogProviderProps {
...
+                } else {
+                    // either one or both are Un-deploying the same app
+                    return true;
+                }
+            }
+        }
+
+        if (newAct.type === ActionTypes.CreateEnvironmentLock && act.type === newAct.type) {
+            if (newAct.environment === act.environment) {
+                // conflict, locking the same env twice
```
medo-freiheit, line 97: Why we can't lock the same environment more than once?

services/frontend-service/src/ui/Batch.tsx
```diff
@@ -56,9 +68,55 @@ export interface ConfirmationDialogProviderProps {
...
+                return true;
+            }
+        }
+
+        if (newAct.type === ActionTypes.CreateApplicationLock && act.type === newAct.type) {
+            if (newAct.environment === act.environment && newAct.application === act.application) {
+                // conflict, locking the same app/env twice
```
medo-freiheit, line 104: same question for application locks

services/frontend-service/src/ui/Batch.tsx
```diff
@@ -56,9 +68,55 @@ export interface ConfirmationDialogProviderProps {
...
+                return true;
+            }
+        }
+    }
+    return false;
+};
+
+export const exportedForTesting = {
```
medo-freiheit, line 112: will it be better if you just export `IsConflictingAction`?

services/frontend-service/src/ui/Batch.test.tsx
```diff
@@ -193,6 +193,31 @@ describe('Confirmation Dialog Provider', () => {
...
+                createEnvironmentApplicationLock: {
+                    environment: 'dummy environment',
+                    application: 'dummy application',
+                    lockId: '1234',
+                    message: 'hello',
+                },
+            },
+        },
+    ];
+
```
medo-freiheit, line 220: I have a question about the logic but it may be trivial, what happened if a user/developer does the following:
`deployApp : app1 -> createApplock : lock1 -> undeployApp: app1 -> some other actions`
Assuming that he will click ` add anyway ` in the third step, then he checkout, 
the third action will fail with its original error ( because we can't un-deploy locked app) right? are the remaining actions will be ignored?

services/frontend-service/src/ui/Batch.tsx
tameremad: This applies for actions in a single batch. so adding two locks at the same time won't be useful. but you can still add as many but in more batches.

services/frontend-service/src/ui/Batch.tsx
tameremad: Good Idea! :+1:

services/frontend-service/src/ui/Batch.tsx
tameremad: I looked into this, It could work but I thought since we don't really need to export it (other than to test it), maybe we shouldn't export it (security risks, best practices)? But I'm not sure.

services/frontend-service/src/ui/Batch.tsx
tameremad: Yes. That's what I have in the story. I think it is not useful to lock the same thing twice in one batch at least. you can still lock it in another batch. and you can still apply it anyway.

services/frontend-service/src/ui/Batch.test.tsx
tameremad: Yeah, I'm actually not sure. this logic is handled in the backend.  In line `225` the transformer fails and then propagates the error all the way back to the service I think. This will not push any actions though. so nothing will change in the repo and you'll just get a notification error in the UI.
https://github.com/freiheit-com/kuberpult/blob/14c145b81934018f3c95ee3904974699e64bcdae/services/cd-service/pkg/repository/repository.go#L218-L232

services/frontend-service/src/ui/Batch.tsx
medo-freiheit: @sven-urbanski-freiheit-com what do you think about that?

hannesg (commented):
Maybe also offer to remove the conflicting actions right away?

services/frontend-service/src/ui/Batch.tsx
```diff
@@ -56,9 +68,55 @@ export interface ConfirmationDialogProviderProps {
 const InCart = (actions: BatchAction[], action: BatchAction) =>
     actions ? actions.find((act) => JSON.stringify(act.action) === JSON.stringify(action.action)) : false;
 
+const IsDeployment = (t: ActionTypes) =>
+    t === ActionTypes.Deploy || t === ActionTypes.PrepareUndeploy || t === ActionTypes.Undeploy;
+
+const IsConflictingAction = (cartActions: BatchAction[], newAction: BatchAction) => {
```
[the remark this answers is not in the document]
tameremad: Do you mean that the user can `replace` the actions in the cart? 
I think that's a good idea, though I think it could take some time to be done. because I'm not sure yet how to delete the actions and if there are multiple conflicts. and also if you can return them as references to the actual actions or indices? 

I can create a new story. would that work?

services/frontend-service/src/ui/Batch.tsx
```diff
@@ -116,6 +175,20 @@ export const ConfirmationDialogProvider = (props: ConfirmationDialogProviderProp
             />
         )) ?? null;
 
+    const conflictMessage = conflict ? (
+        <TextField
```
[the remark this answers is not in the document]
tameremad: I chose it for the warning and the outline properties. Simply listing text could get really messy. esp. with the locks being there too.

services/frontend-service/src/ui/Batch.tsx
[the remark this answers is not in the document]
hannesg: Yeah sure!

services/frontend-service/src/ui/Batch.tsx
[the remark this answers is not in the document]
hannesg: Material UI has some other elements that create a structure. I guess an alert message would be the right thing in this case https://mui.com/components/alert/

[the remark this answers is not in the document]
tameremad: i'm not sure why, but i can't get it in the red color. (I set the prop to error but the component didn't change.)
(my current theory is because some different library version. because I encountered other variations between what I can actually do and what's on the documentation)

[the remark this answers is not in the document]
tameremad: @hannesg

medo-freiheit (approved):

PR #106, opened 2022-03-08
SRX-K20BI8 - Alert user on page reload and prevent data loss

tameremad:


services/frontend-service/src/ui/AppBar/ActionsCart.tsx
```diff
@@ -70,6 +70,15 @@ const ActionsList = (props: { openNotification: (msg: string) => void }) => {
     }, [openNotification]);
     const [doActions, doActionsState] = callbacks.useBatch(actions, actionsSucceeded, actionsFailed);
 
+    useEffect(() => {
+        window.onbeforeunload = (e: BeforeUnloadEvent) => {
```
medo-freiheit, line 74: Isn’t this triggered every time? Shouldn’t be only when the cart not empty ?

tameremad: Yes!! :sweat_smile: 
I changed it.

services/frontend-service/src/ui/AppBar/ActionsCart.tsx
```diff
@@ -70,6 +70,17 @@ const ActionsList = (props: { openNotification: (msg: string) => void }) => {
     }, [openNotification]);
     const [doActions, doActionsState] = callbacks.useBatch(actions, actionsSucceeded, actionsFailed);
 
+    useEffect(() => {
+        if (actions.length) {
+            window.onbeforeunload = (e: BeforeUnloadEvent) => {
```
hannesg, line 75: This onbeforeunload never gets removed from the window. Does that mean that once you have an action list you cannot leave anymore?
This library does it correct btw: https://www.npmjs.com/package/react-beforeunload

medo-freiheit (approved):

PR #107, opened 2022-03-08
 adapt helm chart & move getActionAuthor to frontend-service

medo-freiheit:


services/frontend-service/pkg/cmd/server.go
```diff
@@ -127,8 +131,39 @@ type Auth struct {
 	HttpServer http.Handler
 }
 
+func getActionAuthor(r *http.Request) *auth.User {
```
tameremad, line 134: I suggest a better name, `getRequestAuthor`

services/frontend-service/pkg/cmd/server.go
```diff
@@ -127,8 +131,39 @@ type Auth struct {
...
+	ctx := r.Context()
+
+	var c Config
+
+	err := envconfig.Process("kuberpult", &c)
```
tameremad, line 139: does this read the env variables multiple times? as in line https://github.com/freiheit-com/kuberpult/blob/c7ccf6ba104b638ff358d351b90dfd7e06567a8c/services/frontend-service/pkg/cmd/server.go#L44

tameremad (approved):

PR #110, opened 2022-03-10
make: Run `ct` in container

fdcds:
Their documentation recommends to use the container image, which has all
necessary tools installed [^1] and simplifies running the tool greatly.

[^1]: https://github.com/helm/chart-testing#installation

### DIFF
--- a/charts/kuberpult/Makefile
+++ b/charts/kuberpult/Makefile
@@ -25,7 +25,7 @@ REL_DIR := $(subst $(TOP_DIR)/,,$(CURDIR))
 
 # Run from container image, as recommended by https://github.com/helm/chart-testing#installation
 # `ct` analyses the Git history, so we need to mount the whole repository.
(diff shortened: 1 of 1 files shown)


charts/kuberpult/Makefile
```diff
@@ -16,6 +16,11 @@
 #Copyright 2021 freiheit.com
 MAKEFLAGS += --no-builtin-rules
 
+# Container image as recommended by https://github.com/helm/chart-testing#installation
+#CT := podman run -ti quay.io/helmpack/chart-testing:v3.5.0 ct
+# Manual / local installation
+CT := ct
```
sven-urbanski-freiheit-com, line 22: This change does not reflect what is written in the PR description.
Shouldn't running ct with the docker image be the default (as it has always the correct version)? I think manually running is an exception.

sven-urbanski-freiheit-com (approved):

PR #115, opened 2022-03-18
Optimize history calculation to reuse previously computed results

hannesg:
The history algorithm currently recalculates the whole history for the repository on every run. However, most of the things in a repo do not change on each commit. I introduced a cache that memorizes the last ten calculation results so that we can reuse them later. This bring the history calculation in most cases down from O(N) to O(1) for N = number of commits in repo.

Benchmark:

```
/build/services/cd-service/pkg/history $ go test -benchmem -bench=.
goos: linux
goarch: amd64
pkg: github.com/freiheit-com/kuberpult/services/cd-service/pkg/history
cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
BenchmarkHistoryNoCache-12             4         289058522 ns/op        36413094 B/op    2330716 allocs/op
BenchmarkHistoryCache-12             465           2560525 ns/op          596726 B/op      23879 allocs/op
PASS
ok      github.com/freiheit-com/kuberpult/services/cd-service/pkg/history       45.740s
```

> Those who do not --learn-from-- cache history are doomed to repeat it.
>
> *Lenin*

### DIFF
--- a/services/cd-service/pkg/history/history.go
+++ b/services/cd-service/pkg/history/history.go
@@ -19,32 +19,36 @@ package history
 import (
 	"fmt"
 	"strings"
+	"sync"
 
+	"github.com/hashicorp/golang-lru"
 	git "github.com/libgit2/git2go/v33"
 )
 
 type cacheNode struct {
+	mx        sync.Mutex
 	children  map[string]*cacheNode
 	changedAt *git.Commit
 	cost      uint
-	reads     uint
-	hits      uint
 }
 
 func (c *cacheNode) get(name string) *cacheNode {
 	if c == nil {
 		return nil
 	}
+	c.mx.Lock()
+	defer c.mx.Unlock()
 	child := c.children[name]
 	if child == nil {
 		child = newCacheNode()
 		c.children[name] = child
 	}
-	child.reads = child.reads + 1
 	return child
 }
 
 func (c *cacheNode) found(commit *git.Commit, cost uint) {
+	c.mx.Lock()
+	defer c.mx.Unlock()
 	c.changedAt = commit
 	c.cost = cost
 }
@@ -53,9 +57,8 @@ func (c *cacheNode) changed() *git.Commit {
 	if c == nil {
 		return nil
 	}
-	if c.changedAt != nil {
-		c.hits = c.hits + 1
-	}
+	c.mx.Lock()
+	defer c.mx.Unlock()
 	return c.changedAt
 }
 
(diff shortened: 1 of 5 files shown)


services/cd-service/pkg/history/history.go
```diff
@@ -19,14 +19,86 @@ package history
...
+	root, ok := c.roots.Get(commitId)
+	if ok {
+		return root.(*cacheNode)
+	}
+	return nil
+}
+
 type History struct {
 	repository *git.Repository
 	commits    map[git.Oid]*CommitHistory
```
hannesg, line 100: TODO(HVG): Remove this map. It would grow unbounded with this usage.

services/cd-service/pkg/history/history.go
```diff
@@ -212,16 +215,18 @@ type treeEntry struct {
...
 type treeNode struct {
-	entry    *git.TreeEntry
-	children map[string]*treeNode
-	queue    []queueEntry
-	commit   *git.Commit
+	entry      *git.TreeEntry
+	children   map[string]*treeNode
+	queue      []queueEntry
+	writeCache *cacheNode
+	cost       uint
```
hannesg, line 226: TODO(HVG): meter the costs somehow

services/cd-service/pkg/history/history.go
```diff
@@ -19,14 +19,86 @@ package history
...
 )
 
+type cacheNode struct {
+	mx        sync.Mutex
+	children  map[string]*cacheNode
+	changedAt *git.Commit
+	cost      uint
+}
+
+func (c *cacheNode) get(name string) *cacheNode {
```
medo-freiheit, line 35: it seems more like 'put' 🤔

medo-freiheit (commented):
I like the idea and the new LRU part looks great, but I got confused with history structure, so maybe @sven-urbanski-freiheit-com could review this

services/cd-service/pkg/history/history.go
```diff
@@ -19,14 +19,86 @@ package history
...
 
+	"github.com/hashicorp/golang-lru"
 	git "github.com/libgit2/git2go/v33"
 )
 
+type cacheNode struct {
+	mx        sync.Mutex
+	children  map[string]*cacheNode
+	changedAt *git.Commit
+	cost      uint
```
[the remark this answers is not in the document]
hannesg: `cost` is the number of commits that needed to be examined to find this result. I think, I'll remove that for now.

services/cd-service/pkg/history/history.go
hannesg: Yes. Now with two days after writing it, I think the wording is not perfect. Also it's called `cacheNode` but it's actually a `resultNode` as we are not storing the result anywhere else.

hannesg: I'll change some wordings here.

medo-freiheit (approved):

PR #116, opened 2022-03-18
Add an option to customize the PVC storage request

mouhsen-ibrahim:


tameremad:
I think we should also add it the the template so it's saved for later. 
https://github.com/freiheit-com/kuberpult/blob/b333db18b9da0027fe954213b4795f0cb2c7a9dd/charts/kuberpult/values.yaml.tpl#L25-L27

tameremad (approved):

charts/kuberpult/values.yaml.tpl
```diff
@@ -25,6 +25,7 @@ cd:
     requests:
       cpu: 1
       memory: 2Gi
+  pvc_storage: 5Gi
```
hannesg, line 28: Shouldn't this rather look like this:

```
pvc:
  storage: 5Gi
```

We probably need to support annotations or storage classes eventually.

mouhsen-ibrahim: Makes sense, I modified the code, thanks for the hint

sven-urbanski-freiheit-com (approved):

PR #118, opened 2022-03-25
SRX-MB6QUU - Add locks table

tameremad:


services/frontend-service/src/ui/AppBar/LocksDrawer.tsx
```diff
@@ -116,75 +242,60 @@ export const LocksDrawer = (props: { data: GetOverviewResponse }) => {
...
-                    <strong>all locks</strong>
-                </Badge>
+                {outdated && <WarningRounded color="error" />}
+                <strong>Locks</strong>
                 <ExpandMoreRounded />
             </Button>
-            <Drawer anchor={'top'} open={state['isOpen']} onClose={toggleDrawer(false)}>
+            <Drawer anchor={'top'} open={drawerOpen} onClose={toggleDrawer(false)}>
                 <Box sx={{ width: 'auto' }} role="presentation">
-                    <AllLocks locks={locks} onClick={toggleDrawer(false)} />
```
medo-freiheit, line 187: Assume we have a lot of looks, and the drawer takes the whole screen length, how we can close the drawer? did you handle this situation?

tameremad: It can be closed with `Esc` button. I think it could be useful to have some close button somewhere in the ui, but I haven't found a good way to do it yet. Do you think a ui close button would work?

medo-freiheit: maybe create a small story for that?

medo-freiheit (approved):

services/frontend-service/src/ui/AppBar/LocksDrawer.tsx
tameremad: Yeah! Good idea. Thanks :)

PR #126, opened 2022-04-19
Generate api.ts with proto

tameremad:


medo-freiheit (approved):

medo-freiheit:
Why this is necessary?

services/frontend-service/Makefile
```diff
@@ -30,7 +30,7 @@ IMAGENAME?=ghcr.io/freiheit-com/kuberpult-frontend-service:${VERSION}
 
 GARBAGE := *.sentinel
 
-proto:
+proto: src/api/api.ts
 	$(MAKE) -C../../pkg/api all
```
fdcds, line 34: This declares a dependency: "To build `proto` you first need to create or update `src/api/api.ts`, after you have done that, you can create `proto` using the following command(s): ...". It seems surprising that `src/api/api.ts` is required for the command to succeed, since it is not even executed in the same directory (but in `../../pkg/api`). Is it a dependency, because of some include rules that include it from a different directory?

tameremad: You are correct, it's not required for the command to succeed. But it is required to generate all proto related files. 
there are the `*.pb.go` files and then also `api.ts` is needed for the frontend to run. and there's another target at:
https://github.com/freiheit-com/kuberpult/blob/8fd38a90af163d65f97b76eeb22514a69e4e2490/services/frontend-service/Makefile#L75-L77

I thought this is the best place to have such a dependency. 

I think is also possible to create a new target that runs both proto and generates `api.ts`. Do you have any suggestions?

fdcds: I would distinguish between "real" targets that actually build something from other files (like the `api.ts` target) and targets that are purely "virtual" and "collect" the "real" targets (like the `all` target).

fdcds: > executing `make run` alone will not work and you still need to do a `make all run` But in reality what's missing is the `api.ts` file

In this case it is good to make the dependency explicit. Especially since targets given on the command line could also be built in parallel. Should maybe the `run` target depend on `api.ts`? Does that make sense?

tameremad: Yes That makes more sense, I added a new make target with this dependency included there. Now I think it's a little better.

fdcds (approved):

services/frontend-service/Makefile
```diff
@@ -42,6 +42,9 @@ export KUBERPULT_CDSERVER?=localhost:8443
 run: bin/main
 	./bin/main
 
+start: src/api/api.ts
```
fdcds, line 45: Should this also depend on another target, or is just having the `api.ts` file enough to run the app?

tameremad: It is enough to start the ui although you need to run the frontend and backend services for KP to run. but it doesn't depend on them. / it doesn't break. :+1:

PR #128, opened 2022-04-22
Move lock message inputs from inline to actions cart

tameremad:
The aim of this PR is to make locking an application or environment simpler. Most use cases include adding multiple locks at the same time (with the same message). So by moving the input field to the cart, the locking process is simpler for the users.

The `BatchAction` type is generated from the API proto file directly. This type includes a `create...(env/app)...Lock`. which was used to represent lock actions in the Cart. 

After moving the input field, the lock button has to add lock actions to Cart without any lock message. which creates a problem because the message field is required for `BatchAction` but it's not yet available.

To solve it, This PR will:
* Create a new Type Union of all the different actions, namely, `CartAction`, and then use it to represent different actions in the frontend.
* Remove the input fields from `ReleaseDialog/LockButtonGroup` and from `ReleaseDialog/CreateLockButton`
* Use `CartAction` instead of `BatchAction` in `Warnings`, `Releases`, `ReleaseDialog`, `ActionsCart`, and `index.tsx`.
* Create a new file `ActionDetails` that contains the `CartAction` type, and the function to transform `CartActions` to `BatchActions` before sending the API request.
* Move `getActionDetails`, `ActionTypes` from `ConfirmationDialog` to `ActionDetails`
* Add an input field in `CheckoutDialog` that appears only when the cart contains create-lock actions. use this input as the lock message.
* Add tests in `ActionDetails.test` and modify the tests of `CheckoutDialog.test` and `ConfirmationDialog.test` to use `CartAction`
* Create an issue for more improvement points from this PR: https://github.com/freiheit-com/kuberpult/issues/195


services/frontend-service/src/ui/ReleaseDialog.tsx
```diff
@@ -256,34 +209,24 @@ export const CreateLockButton = (props: { applicationName?: string; environmentN
...
+                          message: 'default-lock',
                       },
                   }
                 : {
                       $case: 'createEnvironmentLock',
                       createEnvironmentLock: {
                           environment: environmentName,
                           lockId: randomLockId(),
-                          message: message,
+                          message: 'default-lock',
```
fdcds, line 220: Maybe like this?
```suggestion
                          message: 'unspecified reason',
```

`default-lock` looks a bit like we forgot to replace an internal identifier with a human-readable text.

services/frontend-service/src/ui/ActionsCart/CheckoutDialog.tsx
```diff
@@ -39,22 +49,9 @@ export const callbacks = {
...
-    }, [setActions, openNotification, closeDialog]);
-    const actionsFailed = useCallback(() => {
-        closeDialog();
-        openNotification('Actions were not applied. Please try again!');
-    }, [openNotification, closeDialog]);
-    const [doActions, doActionsState] = callbacks.useBatch(actions, actionsSucceeded, actionsFailed);
-
-    switch (doActionsState.state) {
+const ApplyButton = (props: { doActions: () => void; state: string }) => {
+    const { doActions, state } = props;
```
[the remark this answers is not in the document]
fdcds: Sorry, I misremembered. It should have been:

```suggestion
const ApplyButton: VFC<{doActions: () => void; state: string}> = ({doActions, state}) => {
    const { doActions, state } = props;
```

VFC is the React type for a functional component that does not take children as an argument. In Revolution we did it this way to have slightly stricter types, to ensure our components actually behave like React expects, and still have a somewhat concise variant of the argument list in `= ({...}) =>`. Your choice whether you want to do it this way.

services/frontend-service/src/ui/ActionsCart/CheckoutDialog.tsx
```diff
@@ -78,11 +75,75 @@ const ApplyButton = (props: { openNotification: (msg: string) => void; closeDial
...
-    const [notifyMessage, setNotifyMessage] = React.useState('');
+const showTextField = (actions: BatchAction[]) => {
+    for (const act of actions) {
+        const t = GetActionDetails(act).type;
+        if (t === ActionTypes.CreateEnvironmentLock || t === ActionTypes.CreateApplicationLock) {
+            return true;
+        }
+    }
+    return false;
+};
```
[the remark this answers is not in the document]
tameremad: Thank you for suggesting these functions and for including the links, I hadn't used them before! 

It works, but I still had to extract the type from `lockActions` because the `.includes` function also complained :(

services/frontend-service/src/ui/ActionsCart/CheckoutDialog.tsx
```diff
@@ -78,11 +75,75 @@ const ApplyButton = (props: { openNotification: (msg: string) => void; closeDial
...
+                    placeholder="default-lock"
+                    onChange={updateInput}
+                    className="actions-cart__lock-message"
+                />
+            )}
+            <Button sx={{ display: 'flex' }} onClick={openDialog} variant={'contained'} disabled={disabled}>
+                <Typography variant="h6">
+                    <strong>Apply</strong>
+                </Typography>
+            </Button>
```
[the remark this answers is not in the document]
tameremad: Yeah, I will leave this for the next story because I don't think doing it here will be a good idea.

services/frontend-service/src/ui/ActionsCart/CheckoutDialog.tsx
```diff
@@ -78,11 +75,75 @@ const ApplyButton = (props: { openNotification: (msg: string) => void; closeDial
...
+        <>
+            {showTextField(actions) && (
+                <TextField
+                    label="Lock Message"
+                    variant="outlined"
+                    sx={{ m: 1 }}
+                    placeholder="default-lock"
+                    onChange={updateInput}
+                    className="actions-cart__lock-message"
+                />
```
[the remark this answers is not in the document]
tameremad: I'm sorry, I don't quite understand the question. This is the input that will change the `lockMessage` so it should have the `@latest` value for `lockMessage`. right? 

at the start the `lockMessage` is empty and it's showing the placeholder and after that the `lockMessage` isn't changed anywhere else. so I think It works, unless there is some corner case or something?

services/frontend-service/src/ui/ReleaseDialog.tsx
tameremad: technically this value doesn't matter at all. because this value will be overwritten by the `actionsWithMessages` when we try to apply the actions.
https://github.com/freiheit-com/kuberpult/blob/a090c499230c8f4dbef61c495a2d384c00b8fa71/services/frontend-service/src/ui/ActionsCart/CheckoutDialog.tsx#L184-L188

I think I can use an empty string here, but I thought to keep this as this could be helpful somehow although it doesn't affect anything.

I can change the placeholder for example? or the default value in the text field (because the initial state for `lockMessage` in the input is empty  | and the placeholder is just 'default-lock' ) ?

services/frontend-service/src/ui/ActionDetails.tsx
```diff
@@ -0,0 +1,197 @@
...
+    icon: React.ReactElement;
+
+    // action details optional
+    environment?: string;
+    application?: string;
+    lockId?: string;
+    lockMessage?: string;
+    version?: number;
+};
+export const GetActionDetails = (action: BatchAction): ActionDetails => {
```
hannesg, line 56: Please use lowercase for functions. Uppercase is reserved for classes and react components.

services/frontend-service/src/ui/ActionsCart/CheckoutDialog.tsx
```diff
@@ -78,11 +74,66 @@ const ApplyButton = (props: { openNotification: (msg: string) => void; closeDial
     }
 };
 
+const lockActions = [ActionTypes.CreateEnvironmentLock, ActionTypes.CreateApplicationLock] as const;
+type lockAction = typeof lockActions[number];
+const hasLockAction = (actions: BatchAction[]) =>
+    actions.some((act) => lockActions.includes(GetActionDetails(act).type as lockAction));
+
+const updateMessage = (act: BatchAction, m: string): BatchAction => {
```
hannesg, line 82: :arrow_up: this should probably not have the same input and output types.

services/frontend-service/src/ui/ReleaseDialog.tsx
hannesg: The correct thing (tm) would be to have different types here.

services/frontend-service/src/ui/ActionDetails.tsx
tameremad: I changed the name in the last commit!

services/frontend-service/src/ui/ActionsCart/CheckoutDialog.tsx
tameremad: I tried to do this, I added new types for the lock details with no message. and then a general CartAction type as a superposition of Batch Action and the Lock details. I think this is better than creating new types for every action. but also I'm not totally sure about it. I still need to fix the tests. but I thought to check first.

services/frontend-service/src/ui/ActionDetails.tsx
```diff
@@ -82,104 +118,181 @@ type ActionDetails = {
...
+                ' of "' +
+                action.deploy.application +
+                '" to ' +
+                action.deploy.environment,
+            icon: <MoveToInboxRounded />,
+            environment: action.deploy.environment,
+            application: action.deploy.application,
+            version: action.deploy.version,
+        };
+    if ('createEnvironmentLock' in action)
```
fdcds, line 138: Since `switch` in JavaScript cannot have conditions in `case` (which is what I would have used in Golang in this case), I would use `else if` here to signal that this is one large block that belongs together and examines `action`:
```suggestion
    else if ('createEnvironmentLock' in action)
```

services/frontend-service/src/ui/ActionDetails.tsx
```diff
@@ -82,104 +118,181 @@ type ActionDetails = {
...
+            type: ActionTypes.Deploy,
+            name: 'Deploy',
+            dialogTitle: 'Are you sure you want to deploy this version?',
+            summary:
+                'Deploy version ' +
+                action.deploy.version +
+                ' of "' +
+                action.deploy.application +
+                '" to ' +
+                action.deploy.environment,
```
fdcds, line 132: In a follow-up refactoring PR you could use template literals here to make this a bit more readable: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals
(Since it was already like this before and this is not the aim of your PR, I would not change it now, to make it easy for the reader to see what you changed to achieve the goal of the PR.)

services/frontend-service/src/ui/ActionDetails.tsx
```diff
@@ -82,104 +118,181 @@ type ActionDetails = {
...
+        return {
+            type: ActionTypes.UNKNOWN,
+            name: 'invalid',
+            dialogTitle: 'invalid',
+            summary: 'invalid',
+            icon: <Error />,
+        };
+};
+
+const randomLockId = () => 'ui-' + Math.random().toString(36).substring(7);
```
fdcds, line 219: I had to look up what `Math.random().toString(36).substring(7)` does. Having a named `const randBase36 = () => Math.random().toString(36).substring(7)` function, with a hint that the result will match `[0-9A-Z]{7}` and a link to https://en.wikipedia.org/wiki/Base36, might have helped.

services/frontend-service/src/ui/ActionDetails.tsx
```diff
@@ -82,104 +118,181 @@ type ActionDetails = {
...
+            action: {
+                $case: 'createEnvironmentLock',
+                createEnvironmentLock: {
+                    environment: act.createEnvironmentLock.environment,
+                    lockId: randomLockId(),
+                    message: m,
+                },
+            },
+        };
+    if ('createApplicationLock' in act)
```
fdcds, line 233: I would use `else if`, like above, to more strongly emphasise that these blocks handle different cases for the same variable.

services/frontend-service/src/ui/ActionDetails.tsx
```diff
@@ -82,104 +118,181 @@ type ActionDetails = {
...
+            name: 'invalid',
+            dialogTitle: 'invalid',
+            summary: 'invalid',
+            icon: <Error />,
+        };
+};
+
+const randomLockId = () => 'ui-' + Math.random().toString(36).substring(7);
+
+export const transformToBatch = (act: CartAction, m: string): BatchAction => {
```
fdcds, line 221: Does this function do two things, transform `CartAction` -> `BatchAction` and add a message `m` for some of them? I would split it into two functions, expecting the result to be easier to understand.

services/frontend-service/src/ui/ActionDetails.tsx
```diff
@@ -82,104 +118,181 @@ type ActionDetails = {
...
+    if ('undeploy' in act)
+        return {
+            action: {
+                $case: 'undeploy',
+                undeploy: {
+                    application: act.undeploy.application,
+                },
+            },
+        };
+    else return {};
```
fdcds, line 297: This is essentially an error, right? In that case I would:
```suggestion
    else return null;
```

You could then, for example, `.filter(isNonNullable)` to get rid of these invalid values. With:
```typescript
export const isNonNullable = <T>(value: T): value is NonNullable<T> => {
  return value !== undefined && value !== null;
};
```

services/frontend-service/src/ui/ActionsCart/CheckoutDialog.tsx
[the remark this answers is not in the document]
fdcds: Since this calls `setLockMessage`, I would have expect for it to use the value of `lockMessage` as its content. In my own code I always (?) passed in a pair of value and setter, when I needed to set something. You did not run into it not displaying the `lockMessage`, because `TextField` implicitly also has a local copy of it?

services/frontend-service/src/ui/ActionsCart/CheckoutDialog.tsx
```diff
@@ -78,11 +74,34 @@ const ApplyButton = (props: { openNotification: (msg: string) => void; closeDial
     }
 };
 
+const LockMessageInput: VFC<{ setLockMessage: (m: string) => void }> = ({ setLockMessage }) => {
+    const updateInput = useCallback((e) => setLockMessage(e.target.value), [setLockMessage]);
```
fdcds, line 78: I think this post might apply here: https://reactjs.org/docs/hooks-faq.html#are-hooks-slow-because-of-creating-functions-in-render

With some more explanation here: https://kentcdodds.com/blog/usememo-and-usecallback
Or maybe this one, too: https://dev.to/milu_franz/demystifying-react-hooks-usecallback-and-usememo-1a8j

services/frontend-service/src/ui/ConfirmationDialog.tsx
```diff
@@ -14,45 +14,24 @@ You should have received a copy of the GNU General Public License
...
+    actions ? actions.find((act) => JSON.stringify(act) === JSON.stringify(action)) : false;
 
-const isDeployment = (t: ActionTypes) =>
-    t === ActionTypes.Deploy || t === ActionTypes.PrepareUndeploy || t === ActionTypes.Undeploy;
-
-const getCartConflicts = (cartActions: BatchAction[], newAction: BatchAction) => {
-    const conflicts = new Set<BatchAction>();
+const getCartConflicts = (cartActions: CartAction[], newAction: CartAction) => {
+    const conflicts = new Set<CartAction>();
     for (const action of cartActions) {
```
fdcds, line 30: Note to self: I think here we are reducing over `cartActions` to produce `conflicts`. (i.e. the whole section could be simplified a lot by writing it in a functional way)

services/frontend-service/src/ui/ConfirmationDialog.tsx
```diff
@@ -67,20 +46,6 @@ const getCartConflicts = (cartActions: BatchAction[], newAction: BatchAction) =>
...
-                conflicts.add(action);
-            }
-        }
-
-        if (newAct.type === ActionTypes.CreateApplicationLock && act.type === newAct.type) {
-            if (newAct.environment === act.environment && newAct.application === act.application) {
-                // conflict, locking the same app/env twice
-                conflicts.add(action);
-            }
-        }
```
fdcds, line 83: Why can this go?

services/frontend-service/src/ui/ActionDetails.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+export type DeleteApplicationLock = {
+    deleteApplicationLock: {
+        environment: string;
+        application: string;
+        lockId: string;
+    };
+};
+
+export type CartAction =
```
fdcds, line 87: `CartAction` is the representation within the frontend, while `BatchAction` is the wire format, right? Please explain that in a comment for `CartAction` and ideally for each of the actions create a link to the wire format it represents.

fdcds (commented):
Some more comments.

services/frontend-service/src/ui/ActionsCart/CheckoutDialog.tsx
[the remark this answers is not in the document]
tameremad: Yes, I believe that's how MUI designed the `TextField`

services/frontend-service/src/ui/ActionDetails.tsx
tameremad: That is correct I'll add a comment

services/frontend-service/src/ui/ConfirmationDialog.tsx
tameremad: yeah that's a good idea, i'll leave that for the next PR

services/frontend-service/src/ui/ActionDetails.tsx
tameremad: Done!

services/frontend-service/src/ui/ActionsCart/CheckoutDialog.tsx
```diff
@@ -120,27 +126,26 @@ export const SyncWindowsWarning: VFC<{
...
+        [setNotify]
     );
 
     const closeNotification = useCallback(
         (event: React.SyntheticEvent | React.MouseEvent, reason?: string) => {
-            if (reason === 'clickaway') {
-                return;
-            }
-            setOpenNotify(false);
+            if (reason === 'clickaway') return;
```
sven-urbanski-freiheit-com, line 145: not necessary in this commit, but in the future, i'd like to have a linter that makes sure every block is wrapped in `{]`, even if it's just one line

PR #134, opened 2022-04-26
Use buf plugins instead of manually installing protoc-gen-* binaries

fdcds:
Not having to manually install the `protoc-gen-*` binaries enables
simplifications of the `start.sh` scripts and makes the `docker-compose` start
more robust.  Additionally all tool dependencies required to generate protobuf
code is now in `buf.gen.yaml` instead of split between that file and
`go.tools.mod`.

The versions used in `go.mod` and `buf.gen.yaml` have to be kept in sync!
(Previously, the versions used in `go.mod` and `go.tools.mod` would have
to be kept in sync.)

This also removes manual installation of libgit2 and a yaml library.
libgit2 is already a dependency in `go.mod` and the yaml library was not used.
I included both removals in this commit (even though they are unrelated),
because that allowed some simplification of the makefile code in
the `.install` target.

This also updates protoc-gen-go from 1.26.0 to 1.28.0, protoc-gen-go-grpc
from 1.1.0 to 1.2.0, protoc-gen-grpc-gateway from 2.4.0 to 2.10.0.

Story: SRX-Y9BRXK
Story: https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/STORY/-Mwb00UWgoLeu2OucGwW/


hannesg:
Uh nice, when I started using buf it didn't have those!

hannesg (approved):

PR #136, opened 2022-04-26
Remove unused proxies

fdcds:
Instead of using `google.api.annotations` we now implement the 5 HTTP handlers
we need ourselves.  This removes the dependency on grpc-gateway (which
implements `google.api.annotations`), which we used in two different versions
before.

We cannot replace these handlers with grpc-web (yet), because that requires
changing jenkins-shared-libraries and updating all teams to use the new
version.

For the web UI we still use grpc-web.

Story: SRX-Y9BRXK
Story: https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/STORY/-Mwb00UWgoLeu2OucGwW/


sven-urbanski-freiheit-com (approved):
looks good to me, nice structure.
Things to check (we can check it together)
* http: health endpoint
* grpc: ui (esp batch endpoint)
* http: rest calls from jenkins, e.g. deploy, lock/unlock and releasetrain

services/frontend-service/pkg/cmd/server.go
```diff
@@ -69,37 +75,42 @@ func RunServer() {
...
-		mux.Handle("/environments/", grpcProxy)
-		mux.Handle("/batches", grpcProxy)
+		mux.Handle("/environments", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			defer readAllAndClose(req.Body, 1024)
+			httpHandler.Handle(w, req)
+		}))
 		mux.Handle("/", http.FileServer(http.Dir("build")))
-
-		httpSrv := &setup.CORSMiddleware{
+		splitGrpcHandler := http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
```
fdcds, line 103:
```suggestion
		// Split HTTP REST from gRPC Web requests, as suggested in the documentation:
		// https://pkg.go.dev/github.com/improbable-eng/grpc-web@v0.15.0/go/grpcweb
		splitGrpcHandler := http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
```

PR #143, opened 2022-05-04
SRX-63BV32 If config.json file is invalid, give a fatal error during repository initialization

mnishamk-freiheit:
https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/FILESYSTEM/FOLDER/ROOT/STORY/-Mp6i5lzSJREYmTNHy6p

When the config.json file in any of the environments is an invalid json file (missing file is ok) then do not start kuberpult cd-service, throw a valid error message and quit, do the same in /health check as well.


services/cd-service/pkg/service/service.go
```diff
@@ -72,8 +72,13 @@ func (s *Service) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *Service) ServeHTTPHealth(w http.ResponseWriter, r *http.Request) {
-	w.WriteHeader(200)
-	fmt.Fprintf(w, "ok\n")
+	if err := s.Repository.State().ValidateConfigs(); err != nil {
```
mnishamk-freiheit, line 75: This seems a bit expensive to do in every health check

mouhsen-ibrahim:
@mnishamk-freiheit please add tests

services/cd-service/pkg/repository/repository.go
```diff
@@ -891,6 +912,18 @@ func names(fs billy.Filesystem, path string) ([]string, error) {
 	return result, nil
 }
 
+func validateJsonFile(fs billy.Filesystem, path string) (error) {
+	if buf, err := readFile(fs, path); err != nil {
+		return wrapFileError(err, path, "could not read json file")
+	} else {
+		valid := json.Valid(buf)
+		if (!valid) {
+			return errors.New("JSON file " + path + " is invalid.")
```
sven-urbanski-freiheit-com, line 921:
```suggestion
			return errors.New("JSON file " + path + " is not valid json.")
```

services/cd-service/pkg/service/service.go
sven-urbanski-freiheit-com: I agree. Let's remove this from the health check.

services/cd-service/pkg/repository/repository.go
```diff
@@ -208,7 +224,12 @@ func New(ctx context.Context, cfg Config) (Repository, error) {
 				}
 			}
 			// check that we can build the current state
-			if _, err := result.buildState(); err != nil {
+			state, err := result.buildState();
+			if err != nil {
+				return nil, err
+			}
+			
+			if err := state.validateConfigs(); err != nil {
```
hannesg, line 232: you could just use `state` and check whether `GetEnvironmentConfigs` returns an error. Then you would also check that the configs are semantically valid.

mnishamk-freiheit: Had originally discussed this with @sven-urbanski-freiheit-com , and he mentioned to keep validity check without semantics which is why the valid function was added

sven-urbanski-freiheit-com: `semantically valid` is a bit of a strong word for the checks we do there ;) But yes, we check a bit more.

But yes, I would agree with Hannes, you can basically just call `GetEnvironmentConfigs` and get a better result.
Just make sure that the error returned is actually readible (but that should already be the case) @mnishamk-freiheit

mnishamk-freiheit: ,"msg":"repository.new.error","error":"invalid character ':' after object key:value pair",

this is the error if you use GetEnvironmnetConfigs, it comes from json.decoder.decode method. Might want to wrap this error in something readable pointing to the actual file

services/cd-service/pkg/repository/repository_test.go
```diff
@@ -246,6 +246,141 @@ func TestNew(t *testing.T) {
 					t.Errorf("mismatched file content, expected %s, got %s", "hello", content)
 				}
 			},
+		}, {
+			Name:   "Old with valid config initializes correctly",
+			Branch: "master",
+			Setup:  func(t *testing.T, remoteDir, localDir string) {
```
sven-urbanski-freiheit-com, line 252: please make the 2 new tests into a seperate table driven test, as discussed.

PR #147, opened 2022-05-13
SRX-4JC9OG allow setting the release number when creating new commits

hannesg:
This change allows a new parameter for the deploy endpoint `version` to make the endpoint (nearly) idempotent. The parameter is optional. If it's omitted, kuberpult will use the existing behaviour and use the latest release+1.
 
This also slightly changes the http status code. Until now, the status code 204 ("No Content") is used to indicate success. However, we have now two different success outcomes. To reflect this, kuberpult will respond with 201 ("Created") when a new release was created and 200 ("Okay") when the release already exists.

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -204,39 +204,45 @@ func (c *CreateApplicationVersion) Transform(ctx context.Context, fs billy.Files
 		}
 	}
 	result := ""
-	for env, man := range c.Manifests {
-		envDir := fs.Join(releaseDir, "environments", env)
+	isLatest, err := isLatestsVersion(fs, c.Application, version)
+	if err != nil {
+		return "", err
+	}
+	if isLatest {
+		for env, man := range c.Manifests {
+			envDir := fs.Join(releaseDir, "environments", env)
 
-		config, found := configs[env]
-		hasUpstream := false
-		if found {
-			hasUpstream = config.Upstream != nil
-		}
-
-		if err = fs.MkdirAll(envDir, 0777); err != nil {
-			return "", err
-		}
-		if err := util.WriteFile(fs, fs.Join(envDir, "manifests.yaml"), []byte(man), 0666); err != nil {
-			return "", err
-		}
-
-		if hasUpstream && config.Upstream.Latest {
-			d := &DeployApplicationVersion{
-				Environment:   env,
-				Application:   c.Application,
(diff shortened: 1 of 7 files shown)


mnishamk-freiheit:
do we need to make sure that there was no existing code (outside of kuberpult) calling /release api and expecting 204 (and just just 20x)

services/cd-service/pkg/repository/transformer.go
```diff
@@ -129,6 +135,7 @@ func (t TransformerFunc) Transform(ctx context.Context, fs billy.Filesystem) (st
 var _ Transformer = TransformerFunc(func(_ context.Context, _ billy.Filesystem) (string, error) { return "", nil })
 
 type CreateApplicationVersion struct {
+	Version        uint64
```
sven-urbanski-freiheit-com, line 138: How does the `publish.sh` script know what version to put here?

services/cd-service/pkg/repository/transformer.go
```diff
@@ -356,68 +416,67 @@ type CleanupOldApplicationVersions struct {
...
+	return versions[0 : positionOfOldestVersion-(keptVersionsOnCleanup-1)], err
+}
 
-	// 3) remove all releases older than x
-	versions, err := state.GetApplicationReleases(c.Application)
+func (c *CleanupOldApplicationVersions) Transform(ctx context.Context, fs billy.Filesystem) (string, error) {
+	oldVersions, err := findOldApplicationVersions(fs, c.Application)
 	if err != nil {
-		return "", fmt.Errorf("cleanup: could not get application releases for app '%s': %w", c.Application, err)
+		return "", err
```
sven-urbanski-freiheit-com, line 463: Why remove this error message?
I would suggest to keep it.
```suggestion
		return "", fmt.Errorf("cleanup: could not get application releases for app '%s': %w", c.Application, err)
```

services/cd-service/pkg/repository/transformer.go
hannesg: > do we need to make sure that there was no existing code (outside of kuberpult) calling /release api and expecting 204 (and just just 20x)

Yes. I actually checked that. The existing scripts take any 200 as success.

hannesg:
> do we need to make sure that there was no existing code (outside of kuberpult) calling /release api and expecting 204 (and just just 20x)

Yes. I actually checked that. The existing scripts take any 200 as success.

PR #153, opened 2022-05-20
Execution plan executor for github-actions

mnishamk-freiheit:


Makefile
```diff
@@ -51,6 +51,7 @@ plan:
 
 all: $(addsuffix /all,$(MAKEDIRS))
 
+init: ;
```
sven-urbanski-freiheit-com, line 54:
```suggestion
init:
```
I think it's fine to just leave it empty

services/cd-service/Makefile
```diff
@@ -90,4 +90,6 @@ release: docker
 
 all: test docker
 
+publish: ;
```
sven-urbanski-freiheit-com, line 93:
```suggestion
publish:
```

.github/workflows/execution-plan.yml
```diff
@@ -29,5 +29,94 @@ jobs:
...
+        with:
+          credentials_json: ${{ secrets.FDC_CORE_CI_IMAGE_READER }} 
+      - name: gcloud authorize
+        run: |
+          gcloud auth configure-docker
+      - name: Planner
+        id: plan
+        run: |
+          init="$(cat artifacts/execution-plan.json | docker run -i eu.gcr.io/freiheit-core/services/github-convertor/github-convertor:0.0.2 init | tr -d '\n')"
+          echo "::set-output name=init::${init}"
```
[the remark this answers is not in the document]
mnishamk-freiheit: I'll try this out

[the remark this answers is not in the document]
sven-urbanski-freiheit-com: we discussed that we won't do this, as it's more complicated. (converter wouldn't return json anymore for example)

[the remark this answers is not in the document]
mnishamk-freiheit: - json output would need to be changed to custom string of prefix + unindented json for this
- would be harder to use this output for any other processes in the future

keeping it as json itself for now

.github/workflows/execution-plan.yml
```diff
@@ -29,5 +29,94 @@ jobs:
...
+          build="$(cat artifacts/execution-plan.json | docker run -i eu.gcr.io/freiheit-core/services/github-convertor/github-convertor:0.0.2 build | tr -d '\n')"
```
sven-urbanski-freiheit-com, line 55: why do you need to delete the newline in the end? Is that really necessary?

sven-urbanski-freiheit-com (approved):

.github/workflows/execution-plan.yml
mnishamk-freiheit: output of github-convertor is indented, this is to convert it to one line json so that set-output works properly (set output doesn't work with multiline strings out of the box

PR #159, opened 2022-05-27
Fix for compile error in queue diff

sven-urbanski-freiheit-com:
This PR fixes a compile error that was accidentally merged.

### DIFF
--- a/services/frontend-service/src/ui/ReleaseDialog.tsx
+++ b/services/frontend-service/src/ui/ReleaseDialog.tsx
@@ -91,9 +91,9 @@ const VersionDiff = (props: { current: number | undefined; target: number; relea
     }
 };
 
-const QueueDiff = (props: { queued: number; current: number; releases: Release[] }) => {
+const QueueDiff = (props: { queued: number; target: number; current: number; releases: Release[] }) => {
     const prefix = 'queued: ';
-    const { current, queued, releases } = props;
+    const { current, queued, releases, target } = props;
     if (queued === 0) {
         // no queue
         return (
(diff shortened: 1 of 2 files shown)


services/frontend-service/src/ui/ReleaseDialog.tsx
```diff
@@ -120,15 +125,21 @@ const QueueDiff = (props: { queued: number; current: number; releases: Release[]
...
             </Tooltip>
         );
     }
     return (
         <Tooltip title={prefix + diff + ' behind'}>
             <span>
-                <span className="ahead">{'+' + diff}</span>
+                &nbsp;
+                {' queued: '}
+                <span className="ahead" data-testid="queue-diff">
```
tameremad, line 140: Is the className correct here? or should it be `behind`?

services/frontend-service/src/ui/ReleaseDialog.tsx
```diff
@@ -91,26 +91,31 @@ const VersionDiff = (props: { current: number | undefined; target: number; relea
     }
 };
 
-const QueueDiff = (props: { queued: number; current: number; releases: Release[] }) => {
+// target is the version we are looking at currently:
+const QueueDiff = (props: { queued: number; target: number; current: number; releases: Release[] }) => {
```
tameremad, line 95: I think we can remove the `current` prop from here?

tameremad (approved):
:+1:

services/frontend-service/src/ui/ReleaseDialog.tsx
sven-urbanski-freiheit-com: done

services/frontend-service/src/ui/ReleaseDialog.tsx
sven-urbanski-freiheit-com: yes! I totally overlooked the colors.
Done

PR #161, opened 2022-05-27
Makefile improvements

sven-urbanski-freiheit-com:
* deleted files Chart.yaml and values.yaml. They are generated! No need to check them in.
* Improved makefile dependencies


charts/kuberpult/Chart.yaml
```diff
@@ -1,29 +0,0 @@
-apiVersion: v2
```
mouhsen-ibrahim, line 1: Why removing this???

mouhsen-ibrahim: In all helm charts that we use `Chart.yaml` and `values.yaml` are defined there

charts/kuberpult/Makefile
```diff
@@ -25,40 +25,42 @@ REL_DIR := $(subst $(TOP_DIR)/,,$(CURDIR))
...
 
 values.yaml: values.yaml.tpl ../../version
 	envsubst < values.yaml.tpl > $@
 
-kuberpult-$(VERSION).tgz: Chart.yaml values.yaml templates/*
+$(TGZ_FILE): Chart.yaml values.yaml templates/*
 	helm package .
 
-ct-test:
+ct-test: Chart.yaml values.yaml
```
mouhsen-ibrahim, line 46: You did not add targets to build `Chart.yaml` and `values.yaml` so I think this will fail

mouhsen-ibrahim (approved):
Not sure if these changes are good on the long run but for now I can approve them

charts/kuberpult/Makefile
sven-urbanski-freiheit-com: I didn't add them, because they are already there.

charts/kuberpult/Chart.yaml
sven-urbanski-freiheit-com: yes, but they are generated automatically, so we don't need to check them in.

PR #174, opened 2022-06-13
Fix main branch build

mnishamk-freiheit:


medo-freiheit (approved):

infrastructure/scripts/execution-plan/container.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+set -ueo pipefail
+echo "eu.gcr.io/freiheit-core/images/execution-plan:2.0-scratch-NG-8"
```
medo-freiheit, line 3: Looks good to me, but why you’re putting the container name ina separate script?

medo-freiheit (approved):

infrastructure/scripts/execution-plan/container.sh
mnishamk-freiheit: refactored code to have common code, so that if container is updated in one place, it gets updated in other script as well.

infrastructure/scripts/execution-plan/plan-pr.sh
```diff
@@ -1,10 +1,11 @@
 #!/bin/bash
 set -ueo pipefail
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]:-$0}"; )" &> /dev/null && pwd 2> /dev/null; )";
```
fdcds, line 3: I would write this as:
```bash
script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
```

Since you know you use bash, you do not need the `:-$0` fallback. Also the output redirects seem a bit much. And the `--` as first argument was surprising to me -- it works without and the `--` argument is not documented in `bash_builtins(1)` for my version of bash (5.1.8).

infrastructure/scripts/execution-plan/plan-pr.sh
```diff
@@ -1,10 +1,11 @@
...
 
 main_branch="main"
 current_branch="$(git rev-parse --abbrev-ref HEAD)"
 # When called locally with Makefile, the default parameters are used
 base="${1:-$(git merge-base "${main_branch}" "${current_branch}")}"
 head="${2:-$(git rev-parse HEAD)}"
 
-git diff --diff-filter=ACMRT --name-only "$base" "$head" | docker run -i -v "$(pwd)":/repo eu.gcr.io/freiheit-core/images/execution-plan:2.0-scratch-NG-8 build-pr
+git diff --diff-filter=ACMRT --name-only "$base" "$head" | docker run -i -v "$(pwd)":/repo $($SCRIPT_DIR/container.sh) build-pr
```
fdcds, line 11: You would need to quote `$SCRIPT_DIR` in case it contains whitespace. And for improved readability you could use the `${SCRIPT_DIR}` form.

fdcds, line 11: Instead of `"$("${SCRIPT_DIR}"/container.sh)"` with `container.sh` outputting the image via `echo` to stdout, I would instead have:
```suggestion
source "${SCRIPT_DIR}"/container.inc.sh
git diff --diff-filter=ACMRT --name-only "$base" "$head" | docker run -i -v "$(pwd)":/repo "${BUILDER_IMAGE}" build-pr
 ```
 
 And `container.inc.sh` would be:
 ```bash
BUILDER_IMAGE=eu.gcr.io/freiheit-core/images/execution-plan:2.0-scratch-NG-8
```

PR #175, opened 2022-06-13
Fix main branch build

mnishamk-freiheit:
optimization comments on previous pull request


infrastructure/scripts/execution-plan/container.inc.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+set -ueo pipefail
```
fdcds, line 2: Not needed, because we just set variables and do not execute commands. In any case, this file is executed in the context of the script that calls `source`, so all the `set` options from there still apply here.

fdcds (approved):

PR #181, opened 2022-06-14
Add libc6-compat to cd-service docker image

mouhsen-ibrahim:


hannesg (changes requested):
I have some good doubt that this is necessary. This would only be necessary if your binary was linked to libc which we don't do.

PR #191, opened 2022-06-16
Apply extra optional annotation to the kuberpult ingress

mouhsen-ibrahim:
These are needed in case we want to have a static IP for the ingress on GKE we need to add this annotation
```
kubernetes.io/ingress.global-static-ip-name: <name of global static IP in GCP>
```

### DIFF
--- a/charts/kuberpult/templates/ingress.yaml
+++ b/charts/kuberpult/templates/ingress.yaml
@@ -20,6 +20,9 @@ metadata:
   annotations:
     cert-manager.io/acme-challenge-type: dns01
     cert-manager.io/cluster-issuer: letsencrypt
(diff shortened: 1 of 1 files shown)


charts/kuberpult/templates/ingress.yaml
```diff
@@ -20,6 +20,9 @@ metadata:
   annotations:
     cert-manager.io/acme-challenge-type: dns01
     cert-manager.io/cluster-issuer: letsencrypt
+{{- range $key, $value := .Values.ingress.annotations }}
```
sven-urbanski-freiheit-com, line 23: please also adapt the release notes

sven-urbanski-freiheit-com (approved):

PR #196, opened 2022-06-21
Fix lint and test helm charts

mnishamk-freiheit:
As of now, linting is checking old charts. 
Make it so that it checks the chart that is built
[SRX-CLM9SO](https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/STORY/-N4gw438l8Dk1N98F6u-/)


charts/kuberpult/ci/test-values-override.yaml
```diff
@@ -0,0 +1,19 @@
+git:
```
medo-freiheit, line 1: What is the use of this file?

mnishamk-freiheit: some values are required in the chart, these cannot be added in values.yaml.tpl since those need to be provided by the end user. Other values like memory and cpu are changed because the kind cluster in the pipeline is underpowered, so memory and cpu is reduced in pipeline.

So the overrides are applied on top of values.yaml to create test-values.yaml.

medo-freiheit (approved):

PR #207, opened 2022-06-24
Add check license as separate build

mnishamk-freiheit:
Remove check license workflow and add it as part of execution plan
Story: [SRX-57TKD2](https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/STORY/-N3djfe-fMevnyrNqHzj/)


medo-freiheit:
If it’s only one target why we are adding it as a separate build?

medo-freiheit (approved):

PR #222, opened 2022-07-03
SRX-XX2WFP - Filter apps by search parameters and team name

tameremad:


services/frontend-service/src/ui/Releases.tsx
```diff
@@ -314,9 +316,25 @@ export const sortEnvironmentsByUpstream = (envs: Environment[], distance: EnvSor
 export const Releases: React.FC<any> = (props: { data: GetOverviewResponse }) => {
     const { data } = props;
     const classes = useStyles(data.environments);
+    const [searchParams] = useSearchParams();
+    let apps = Object.values(data.applications);
 
-    const keys = Object.keys(data.applications);
-    keys.sort();
+    if (searchParams.has('app')) {
+        apps = apps.filter((k) => k.name.includes(searchParams.get('app')!));
```
hannesg, line 323: Wouldn't it be better to have an actual route to the app? Something like "/applications/<app>" maybe?

services/frontend-service/src/ui/Releases.tsx
```diff
@@ -314,9 +316,25 @@ export const sortEnvironmentsByUpstream = (envs: Environment[], distance: EnvSor
...
+    }
+    if (searchParams.has('team')) {
+        apps = apps.filter((k) => k.team.includes(searchParams.get('team')!));
+    }
+
+    apps.sort((a, b) => {
+        if (a.name < b.name) {
```
hannesg, line 330: Does it make sense to sort by name or should we switch to sort by team and then by name?

services/frontend-service/src/ui/Releases.tsx
```diff
@@ -230,6 +231,7 @@ const ApplicationBox: React.FC<any> = (props: {
                 {warnings}
                 {name}
                 {undeployButton}
+                {application.team !== '' ? 'Owner: ' + application.team : null}
```
hannesg, line 234: Can you give the team name a different style. Maybe a lighter gray or something?

services/frontend-service/src/ui/Releases.tsx
mnishamk-freiheit: This is just a filter, since we currently treat / as the list of all applications. So it doesn't make sense to add it as a route. 

When we have details page for applications (v.1.0) it might make sense

services/frontend-service/src/ui/Releases.tsx
sven-urbanski-freiheit-com: I would prefer to sort by team first and then by name

mnishamk-freiheit: updated

services/frontend-service/src/ui/Releases.tsx
mnishamk-freiheit: updated

services/frontend-service/src/ui/Releases.tsx
```diff
@@ -326,15 +326,8 @@ export const Releases: React.FC<any> = (props: { data: GetOverviewResponse }) =>
...
-        if (a.name < b.name) {
-            return -1;
-        } else if (a.name === b.name) {
-            return 0;
-        } else {
-            return 1;
-        }
-    });
+    // Sort by team, within team sort by name
+    apps.sort((a, b) => a.team?.localeCompare(b.team) || a.name?.localeCompare(b.name) || 0);
```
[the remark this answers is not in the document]
mnishamk-freiheit: This was based on hanne's comment for previous cr

medo-freiheit (approved):

PR #234, opened 2022-07-08
SRX-GUPKQB - argocd baseurl argument

medo-freiheit:
* clean up unused Argocd arguments
* add Argocd baseurl argument
* create getConfig rpc

### DIFF
--- a/charts/kuberpult/templates/cd-service.yaml
+++ b/charts/kuberpult/templates/cd-service.yaml
@@ -80,6 +80,8 @@ spec:
           value: {{ required ".Values.git.url is required" .Values.git.url | quote }}
         - name: KUBERPULT_GIT_BRANCH
           value: {{ .Values.git.branch | quote }}
+        - name: KUBERPULT_ARGOCD_BASE_URL
+          value: {{ .Values.argocd.baceUrl | quote }}
         - name: LOG_FORMAT
           value: {{ .Values.log.format | quote }}
         - name: LOG_LEVEL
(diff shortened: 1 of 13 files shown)


mouhsen-ibrahim:
@medo-freiheit please remove installing the ArgoCD client in Dockerfile

pkg/api/api.proto
```diff
@@ -109,10 +109,15 @@ message CreateEnvironmentRequest {
 }
 
 service OverviewService {
+  rpc GetConfig (google.protobuf.Empty) returns (GetConfigResponse) {}
   rpc GetOverview (GetOverviewRequest) returns (GetOverviewResponse) {}
   rpc StreamOverview (GetOverviewRequest) returns (stream GetOverviewResponse) {}
 }
 
+message GetConfigResponse {
+  string argocdBaseUrl = 1;
```
hannesg, line 118: I would suggest that we make an argoCd sub-message. Then we can group all argoCd related configs.

charts/kuberpult/templates/cd-service.yaml
```diff
@@ -80,6 +80,8 @@ spec:
           value: {{ required ".Values.git.url is required" .Values.git.url | quote }}
         - name: KUBERPULT_GIT_BRANCH
           value: {{ .Values.git.branch | quote }}
+        - name: KUBERPULT_ARGOCD_BASE_URL
```
hannesg, line 83: I would let the frontend-service handle this request as it is strictly for the frontend.

medo-freiheit:
> @medo-freiheit please remove installing the ArgoCD client in Dockerfile



@mouhsen-ibrahim @hannesg  I think we still need it for syncing, right? [here](https://github.com/freiheit-com/kuberpult/blob/main/services/cd-service/pkg/service/service.go#L241)

sven-urbanski-freiheit-com:
> > @medo-freiheit please remove installing the ArgoCD client in Dockerfile
> 
> @mouhsen-ibrahim @hannesg I think we still need it for syncing, right? [here](https://github.com/freiheit-com/kuberpult/blob/main/services/cd-service/pkg/service/service.go#L241)

We don't use the sync endpoint in any team, please remove it. @medo-freiheit

sven-urbanski-freiheit-com (approved):

PR #238, opened 2022-07-13
Read config from file in bootstrap mode

mnishamk-freiheit:
If `KUBERPULT_BOOTSTRAP_MODE=true` env variable is provided, then read kuberpult configuration from `./environment_configs.json` and not from manifest repo. 

If bootstrap mode is true, then updating configurations is not allowed. But creating environment without configuration is still allowed.

This file will be created in a subsequent pr from configmap

[SRX-NXORS9](https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/STORY/-N4m28IVhuVJW1-LvRHi/)

### DIFF
--- a/services/cd-service/pkg/repository/repository_test.go
+++ b/services/cd-service/pkg/repository/repository_test.go
@@ -35,10 +35,11 @@ import (
 
 func TestNew(t *testing.T) {
 	tcs := []struct {
-		Name   string
-		Branch string
-		Setup  func(t *testing.T, remoteDir, localDir string)
-		Test   func(t *testing.T, repo Repository, remoteDir string)
+		Name          string
+		Branch        string
+		BootstrapMode bool
+		Setup         func(t *testing.T, remoteDir, localDir string)
+		Test          func(t *testing.T, repo Repository, remoteDir string)
 	}{
 		{
 			Name:  "new in empty directory works",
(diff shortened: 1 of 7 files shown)


services/cd-service/pkg/repository/repository_test.go
```diff
@@ -251,6 +252,81 @@ func TestNew(t *testing.T) {
...
+			BootstrapMode: true,
+			Setup: func(t *testing.T, remoteDir, localDir string) {
+				EnvironmentConfigsPath := filepath.Join(remoteDir, "..", "environment_configs.json")
+				configContent := `{"uniqueEnv": {"upstream": {"latest": true }}}`
+				err := os.WriteFile(EnvironmentConfigsPath, []byte(configContent), 0644)
+				if err != nil {
+					t.Fatal(err)
+				}
+			},
+			Test: func(t *testing.T, repo Repository, remoteDir string) {
```
sven-urbanski-freiheit-com, line 306: I think these should be different test-array entirely.
If it uses different "Setup" and "Test" functions, that's an indicator to make it a seperate test-array.

But maybe you can generalize a bit (but since it's so many different things that are happening here, I think making it seperate test-arrays would be simpler)

services/cd-service/pkg/repository/repository_test.go
```diff
@@ -282,6 +360,50 @@ func TestNew(t *testing.T) {
...
 }
 
+func TestBootstrapError(t *testing.T) {
+	tcs := []struct {
+		Name          string
+		ConfigContent string
+	}{
+		{
+			Name:          "Invalid json in bootstrap configuration",
+			ConfigContent: `{"development": "upstream": {"latest": true}}}`,
```
sven-urbanski-freiheit-com, line 370: maybe add another test with a json syntax error?

services/cd-service/pkg/repository/repository_test.go
sven-urbanski-freiheit-com: (I know you did invent this pattern. We shouldn't continue using it though)

sven-urbanski-freiheit-com (changes requested):
.

services/cd-service/pkg/repository/repository_test.go
mnishamk-freiheit: how is json syntax one different from this one?

sven-urbanski-freiheit-com: yes, you are right.

services/cd-service/pkg/repository/repository_test.go
sven-urbanski-freiheit-com: meant to say "did NOT invent"

sven-urbanski-freiheit-com (approved):

PR #245, opened 2022-07-20
Add v2 route for kuberpult UI overhaul

mnishamk-freiheit:
Add /v2 route for new ui in kuberpult

Since v2 app would have proper routing, added it as a seperate routes file


services/frontend-service/src/v2-ui/App/App.test.tsx
```diff
@@ -0,0 +1,28 @@
+/*This file is part of kuberpult.
```
sven-urbanski-freiheit-com, line 1: looks fine, just one suggestion:
Can we move the old code into a subdirectory like `legacy-ui` and have the new v2 things in the normal `ui` directory?
This way it would make it easier in the future.

sven-urbanski-freiheit-com (approved):

services/frontend-service/src/v2-ui/App/App.test.tsx
mnishamk-freiheit: Was originally planning to have this when switchover was made to have v2 be the default option. 
But yes, will update that

PR #252, opened 2022-07-22
Restart when disk is full

sven-urbanski-freiheit-com:
Adds healthcheck function to each "service" (go file).
Each service has the possibility to declare the health as broken.
The emptyDir storage is used because it should evict the pod automatically as decribed here: https://github.com/kubernetes/kubernetes/issues/63126#issuecomment-387817170


sven-urbanski-freiheit-com:
NOTE:
As of now, this doesn't seem necessary.
Keeping it here, in case we need a way to handle healthchecks from various services.

PR #272, opened 2022-08-03
fix buf installing in docker images

medo-freiheit:
Buf caused an issue in Leandro machine ( Mac m1 chip ) when running docker images


CallMeLeopoldina (approved):

CallMeLeopoldina (approved):

tameremad (approved):
:+1:

PR #280, opened 2022-08-08
Use pnpm as package manager instead of yarn

leandro-freiheit:


services/frontend-service/Makefile
```diff
@@ -74,8 +74,10 @@ test-go: $(GO_FILES) proto
 	$(GO) test ./...
 
 test-ts: src/api/api.ts
-	yarn eslint
-	yarn test --watchAll=false
+	pnpm eslint
+	#pnpm is run on powershell. Without the extra --, it will be `pnpm test watchAll`.
```
sven-urbanski-freiheit-com, line 78: what do you mean "is run on powershell"?
We use bash or sh.

sven-urbanski-freiheit-com: Please either explain why a powershell-related bug is relevant for us, or adapt the comment.

medo-freiheit, line 78:
```suggestion
	# The extra '--' was added because of pnpm, Usage: pnpm test [-- <args>...]
```

medo-freiheit (approved):

services/frontend-service/Makefile
medo-freiheit: It was a misunderstanding, the actual reason is:
The extra '--' was added because of pnpm, Usage: pnpm test [-- <args>...]

infrastructure/docker/install-dependencies/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM alpine:3.15
 
-RUN apk --update add --no-cache libgit2 libgit2-dev go protoc make pkgconfig git cmake yarn gettext
+RUN apk --update add --no-cache libgit2 libgit2-dev go protoc make pkgconfig git cmake pnpm gettext
```
medo-freiheit, line 3: To do: fix this too, installing pnpm using npm

medo-freiheit (changes requested):

infrastructure/docker/install-dependencies/Dockerfile
leandro-freiheit: Since the pnpm is installed globally on the builder Dockerfile, there is no need to install it in this Dockerfile. However, that pnpm add must be removed.

services/frontend-service/Makefile
```diff
@@ -74,8 +74,10 @@ test-go: $(GO_FILES) proto
 	$(GO) test ./...
 
 test-ts: src/api/api.ts
-	yarn eslint
-	yarn test --watchAll=false
+	pnpm eslint
+	# The extra '--' was added because of pnpm, Usage: pnpm test [-- <args>...]
+	#To read more, follow this : [link](https://github.com/pnpm/pnpm/issues/2134)
```
medo-freiheit, line 79:
```suggestion
```
this line is no longer needed, please remove it

medo-freiheit (approved):

PR #282, opened 2022-08-08
SRX-OWCUUV - Removing cache

sven-urbanski-freiheit-com:
### Comparison between localhost and live
the manifest repo was cloned and then used to populate the local kuberpult version. the migration scripts were also applied for this version of kuberpult to run with no errors. 
image


pkg/api/api.proto
```diff
@@ -199,9 +201,3 @@ message Application {
   repeated Release releases = 2;
   string team = 3;
 }
-
-message Commit {
```
hannesg, line 203: :thinking: In my mental model, it really makes sense to have these information always together ( when was something changed and by whom ) and so having this type seems like a good idea. On the other hand, the author for a release will always be a bot user but in that case we could just keep the author name and email empty. Do you think it makes sense to keep this type around under a different name?

tameremad: I tried to keep it at first and have a different name for this structure. I decided to remove it because to me it felt more simpler to have the information there without this extra layer. I would agree with you if it was used in more places. but it only makes sense as the `lock_commit` or in the context of locks. 

I actually hadn't thought of keeping the name and email empty. I can keep this under a different name, e.g. `Metadata`?

i am only against writing the name and email of release author (which is as you said, bots) in the manifest repo. (as part of the scripts)

hannesg (approved):

PR #285, opened 2022-08-09
Add azure auth in UI

mnishamk-freiheit:


services/frontend-service/src/legacy-ui/App/index.tsx
```diff
@@ -82,16 +83,20 @@ const Main = () => {
...
-                    {(overview) => (
-                        <Box sx={{ display: 'flex', marginRight: '14%' }}>
-                            <Header overview={overview} />
-                            <Box component="main" className={classes.main}>
-                                <Releases data={overview} />
-                            </Box>
-                        </Box>
-                    )}
-                </GetOverview>
+                {config.state !== 'pending' && (
```
medo-freiheit, line 86:
```suggestion
                { !!configs && (
```

services/frontend-service/src/legacy-ui/App/AuthContext.tsx
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export function AuthProvider({ children }: { children: React.ReactNode }): JSX.Element {
+    const { configs } = React.useContext(ConfigsContext);
+    const useAzureAuth = configs?.authConfig?.azureAuth?.enabled || false;
+    const msalConfig = React.useMemo(() => getMsalConfig(configs), [configs]);
+    const msalInstance = new PublicClientApplication(msalConfig);
+
+    return (
+        <AuthContext.Provider value={{}}>
```
medo-freiheit, line 65: Why is this context empty? will we assign a value to it and use it somewhere?

mnishamk-freiheit: updated

services/frontend-service/src/legacy-ui/App/index.tsx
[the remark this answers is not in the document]
mnishamk-freiheit: moved logic to auth context, added one more condition so that it waits till config is returned

pkg/api/api.proto
```diff
@@ -119,7 +119,18 @@ message GetFrontendConfigResponse {
   message ArgoCD {
     string baseUrl = 1;
   }
+  message AzureAuthConfig {
```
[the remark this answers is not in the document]
mnishamk-freiheit: updated

charts/kuberpult/templates/frontend-service.yaml
```diff
@@ -82,6 +82,18 @@ spec:
         - name: KUBERPULT_ENABLE_TRACING
           value: "{{ .Values.datadogTracing.enabled }}"
 {{- end }}
+{{- if .Values.auth.azureAuth.enabled }}
+        - name: KUBERPULT_AZURE_ENABLE_AUTH
+          value: "{{ .Values.auth.azureAuth.enabled }}"
```
sven-urbanski-freiheit-com, line 87:
```suggestion
        - name: KUBERPULT_AZURE_ENABLE_AUTH
          value: "{{ .Values.auth.azureAuth.enabled }}"
{{- if .Values.auth.azureAuth.enabled }}
```

sven-urbanski-freiheit-com (approved):

services/frontend-service/src/legacy-ui/App/AuthContext.tsx
```diff
@@ -0,0 +1,126 @@
...
+                        <MsalProvider instance={msalInstance}>
+                            <AuthenticatedTemplate>
+                                <AzureAuthTokenProvider>{children}</AzureAuthTokenProvider>
+                            </AuthenticatedTemplate>
+                            <UnauthenticatedTemplate>
+                                <AzureAutoSignIn />
+                            </UnauthenticatedTemplate>
+                        </MsalProvider>
+                    ) : (
+                        <AuthTokenContext.Provider value={{ token: '', authHeader: {} }}>
```
medo-freiheit, line 118: Why wrapping `{children}` with Azure Auth, did you consider use cases that we will not use it?

medo-freiheit (changes requested):

services/frontend-service/src/legacy-ui/App/AuthContext.tsx
mnishamk-freiheit: When auth is not used, the children will still need to use the hook (as hook shouldn't be called conditionally)
```
    const { authHeader } = React.useContext(AuthTokenContext);
```

So this removes the need for the children to conditionally call the hook (which isn't good). This way an empty hook is provided when there is no auth and there is no issue.

services/frontend-service/src/legacy-ui/App/AuthContext.tsx
```diff
@@ -0,0 +1,126 @@
...
+    }
+    return <></>;
+};
+
+export function AuthProvider({ children }: { children: React.ReactNode }): JSX.Element {
+    const { configs } = React.useContext(ConfigsContext);
+    const useAzureAuth = configs?.authConfig?.azureAuth?.enabled || false;
+    const useAuth = useAzureAuth;
+    const msalConfig = React.useMemo(() => getMsalConfig(configs), [configs]);
+    const msalInstance = new PublicClientApplication(msalConfig);
```
medo-freiheit, line 102: Is It ok to send ` auth: {
        clientId: null,
        authority:  null,
        redirectUri: null,
    },
    cache: {
        cacheLocation: 'sessionStorage',
        storeAuthStateInCookie: false,
    },` to `new PublicClientApplication()` 
 Please add some tests for `AuthProvider`

services/frontend-service/src/legacy-ui/ActionsCart/CheckoutDialog.tsx
```diff
@@ -32,21 +32,25 @@ import { Close } from '@material-ui/icons';
...
+    useBatch: (acts: BatchAction[], success?: () => void, fail?: () => void, authHeader: any = {}) =>
         useUnaryCallback(
             useCallback(
                 (api) =>
                     api
                         .batchService()
-                        .ProcessBatch({
-                            actions: acts,
-                        })
+                        .ProcessBatch(
```
medo-freiheit, line 45: Why do we just add the header to `overview & batch` services, what about the other services? I'm not sure if this the best way to send the Auth header I mean why do we need to use the context and then add it to the request, we should have something like middleware that added this header to all requests by default 🤔

services/frontend-service/src/legacy-ui/App/AuthContext.tsx
medo-freiheit, line 102: Is It ok to send ` auth: {
        clientId: null,
        authority:  null,
        redirectUri: null,
    },
    cache: {
        cacheLocation: 'sessionStorage',
        storeAuthStateInCookie: false,
    },` to `new PublicClientApplication()` 
 Please add some tests for `AuthProvider`

services/frontend-service/src/legacy-ui/ActionsCart/CheckoutDialog.tsx
mnishamk-freiheit: We don't have a middleware right now, and as of now only two services are used. (everything else would be piped through batch service) 

checked with sven on this, unless there is more services incoming, he suggested keeping it like so

services/frontend-service/src/legacy-ui/App/AuthContext.tsx
mnishamk-freiheit: How would you suggest adding tests for AuthProvider. All the data given to authprovider comes from other contexts, and it uses azure components and functions to function properly.

medo-freiheit: I answered you before 
`
You can import the contexts with mocked values, I'm not sure how @azure/msal-react works but I guess it should redirect you or open a popup window to authenticate, so the test should make sure if you give the right input @azure/msal-react should display the right output
`
`@azure/msal-react` is a third party library we can't just say ` it uses azure components and functions to function properly` we can't trust them blindly, it's better to have some tests to simulate all possible use cases.

mnishamk-freiheit: for 1, No. the function doesn't accept empty values its given by microsoft and dont have access to it.  So updated the structure so that the function will be called only when azure auth is enabled.

services/frontend-service/src/legacy-ui/ActionsCart/CheckoutDialog.tsx
medo-freiheit: > We don't have a middleware right now, and as of now only two services are used. (everything else would be piped through batch service)
> 
Hmm, this is not quite correct, you can use `useUnaryCallback` & `useObservable` to achieve that.

services/frontend-service/src/legacy-ui/App/index.tsx
```diff
@@ -51,7 +52,8 @@ export const Spinner: React.FC<any> = (props: any) => {
 };
 
 const GetOverview = (props: { children: (r: api.GetOverviewResponse) => JSX.Element }): JSX.Element | null => {
-    const getOverview = React.useCallback((api) => api.overviewService().StreamOverview({}), []);
+    const { authHeader } = React.useContext(AuthTokenContext);
```
medo-freiheit, line 55:
```suggestion
    const getOverview = React.useCallback((api,authHeader) => api.overviewService().StreamOverview({}, authHeader), []);
```

medo-freiheit: Same goes for `useUnaryCallback`, 
in `useUnaryCallback` & `useObservable` add this 
```
const { authHeader } = React.useContext(AuthTokenContext);
```
then use it like this:
```
callback(api, authHeader)
```
then please delete all the const { authHeader } = React.useContext(AuthTokenContext); from the requests

medo-freiheit (changes requested):

services/frontend-service/src/legacy-ui/ActionsCart/CheckoutDialog.tsx
medo-freiheit: Please see the changes here: https://github.com/freiheit-com/kuberpult/pull/285#discussion_r942546720

mnishamk-freiheit: use unarycallback doesn't have access to which api is called right now. it only has access to the api level, not the api.service.function level

services/frontend-service/src/legacy-ui/App/AuthContext.tsx
mnishamk-freiheit: even for testing msalprovider, you'd need an azure ad app client id.

https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-react/test/TestConstants.ts 
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-react/test/components/AuthenticatedTemplate.spec.tsx

So would still be dependant on someone elses azure ad app, which isn't ideal either.

services/frontend-service/src/legacy-ui/App/index.tsx
mnishamk-freiheit: seems like indirection for its own sake, but sure

services/frontend-service/src/legacy-ui/ActionsCart/CheckoutDialog.tsx
medo-freiheit: Yes, they don't need to now what is calling but they can pass more arguments to the callback they calling, please take a look at the changes, if you didn't get it we can have a call

services/frontend-service/src/legacy-ui/App/index.tsx
medo-freiheit: > 



> seems like indirection for its own sake, but sure

I didn’t get this, please can you clarify ?

mnishamk-freiheit: Needed additional type casting because of browser headers being required

services/frontend-service/src/legacy-ui/ActionsCart/CheckoutDialog.tsx
mnishamk-freiheit: updated

services/frontend-service/src/legacy-ui/App/index.tsx
mnishamk-freiheit: https://github.com/freiheit-com/kuberpult/pull/285/commits/cdde5751fb07ac239868e0f1af5485197576ce40

services/frontend-service/src/legacy-ui/App/AuthContext.tsx
mnishamk-freiheit: added tests

medo-freiheit (approved):

PR #287, opened 2022-08-10
Add Material Components package

tameremad:


services/frontend-service/package.json
```diff
@@ -9,6 +9,7 @@
     "@material-ui/core": "5.0.0-alpha.34",
     "@material-ui/icons": "5.0.0-alpha.34",
     "@types/react-beforeunload": "^2.1.1",
+    "material-components-web": "^14.0.0",
```
medo-freiheit, line 12: Why not `https://github.com/mui/material-ui`?

tameremad: material-ui is the one already installed. 
https://material.io/develop/web/getting-started
this is the new one.

medo-freiheit: Yes, is a different 'material-components-web' from https://www.npmjs.com/package/@mui/material, we are spoused to migrate from '@material-ui/core' to the newer version '@mui/material', right?

medo-freiheit: If we have `@mui/material` why would we need `material-components-web`

tameremad: No, I believe that we want to not use Mui at all. and use the material design ones. because in mui the components for anything were ready and cannot be changed according to the use case. 
see: https://freiheit.slack.com/archives/C01HK6J6AQ1/p1651685697627859
but for MDC, all the components are made with CSS. and it's still possible to add individual changes to the design.

tameremad: Because we don't want to use any mui. we need it to get the CSS styles. 
```
app-v2.scss: 

- @use '@material/button';
```

in the react app we use only classNames to style components
```
<button className="mdc-button mdc-top-app-bar__action-item">
    <span className="mdc-button__ripple"></span>
    Planned Actions
</button>
```

medo-freiheit: Ohh, ok I didn’t see this discussion before, then yes let's use material-components-web

medo-freiheit (approved):

hannesg (approved):

PR #292, opened 2022-08-12
Show Kuberpult version UI

leandro-freiheit:


services/frontend-service/pkg/cmd/server.go
```diff
@@ -52,6 +54,7 @@ type Config struct {
 	AzureClientId       string `default:"" split_words:"true"`
 	AzureTenantId       string `default:"" split_words:"true"`
 	AzureRedirectUrl    string `default:"" split_words:"true"`
+	Version             string `default:"testing"`
```
mnishamk-freiheit, line 57:
```suggestion
	Version             string `default:""`
```

pkg/api/api.proto
```diff
@@ -130,7 +130,11 @@ message GetFrontendConfigResponse {
     AzureAuthConfig azureAuth= 1;
   }
   ArgoCD argoCd = 1;
+
```
medo-freiheit, line 133: Please remove the additional lines here

services/frontend-service/pkg/config/config.go
```diff
@@ -14,11 +14,30 @@ You should have received a copy of the GNU General Public License
 along with kuberpult.  If not, see <http://www.gnu.org/licenses/>.
 
 Copyright 2021 freiheit.com*/
+/*
```
medo-freiheit, line 17: Seems to be a duplicated section, please remove it

services/frontend-service/pkg/service/config.go
```diff
@@ -14,6 +14,24 @@ You should have received a copy of the GNU General Public License
 along with kuberpult.  If not, see <http://www.gnu.org/licenses/>.
 
 Copyright 2021 freiheit.com*/
+/*
```
medo-freiheit, line 17: Same here, Duplicated part 😄

medo-freiheit (changes requested):

services/frontend-service/pkg/config/config.go
mnishamk-freiheit: probably the comment ending got modified somehow? and automation caught it. removing the entire comment (above one as well) and adding it back via check.sh should fix it

pkg/api/api.proto
```diff
@@ -130,7 +130,11 @@ message GetFrontendConfigResponse {
...
   Auth authConfig = 2;
+
+  string version = 3;
```
sven-urbanski-freiheit-com, line 136: please rename
```suggestion
  string kuberpultVersion = 3;
```

services/frontend-service/pkg/cmd/server.go
```diff
@@ -52,6 +70,7 @@ type Config struct {
 	AzureClientId       string `default:"" split_words:"true"`
 	AzureTenantId       string `default:"" split_words:"true"`
 	AzureRedirectUrl    string `default:"" split_words:"true"`
+	Version             string `default:""`
```
sven-urbanski-freiheit-com, line 73: here too

services/frontend-service/pkg/config/config.go
mnishamk-freiheit: maybe some goland prettier function that says comment ending should be in same line?

services/frontend-service/src/legacy-ui/AppBar/Header.tsx
```diff
@@ -22,15 +22,21 @@ import { AppDrawer } from './AppDrawer';
...
+    configs: api.GetFrontendConfigResponse;
+}) => {
+    const { overview, configs } = props;
     return (
         <AppBar>
             <Box sx={{ display: 'flex' }}>
                 <Typography component="h1" variant="h6" color="inherit" noWrap sx={{ flexGrow: 1, width: '24rem' }}>
                     <strong>
-                        <code>KUBERPULT UI</code>
+                        <Tooltip title={`Kuberpult ${configs.version || ''}`}>
```
mnishamk-freiheit, line 37: would this show up as an empty tooltip? if so might want to make the default value (if no version is supplied) the same as the text
```suggestion
                        <Tooltip title={`Kuberpult ${configs.version || 'KUBERPULT UI'}`}>
```

medo-freiheit: NO! it will not be an empty tooltip if there is no version supplied it will be `Kuberpult`

services/frontend-service/pkg/cmd/server.go
leandro-freiheit: It is fixed now

pkg/api/api.proto
leandro-freiheit: Fixed

services/frontend-service/pkg/service/config.go
leandro-freiheit: Deleted the duplicate

services/frontend-service/pkg/config/config.go
leandro-freiheit: Already deleted

pkg/api/api.proto
leandro-freiheit: The additional lines were deleted!

services/frontend-service/pkg/cmd/server.go
medo-freiheit: We will keep it as "Version" here because it will be in helm "KUBERPULT_VERSION", because we adding the "KUERPULT" prefix by default, if we make it "KuberpultVersion" it will be "KUBERPULT_KUBERPULT_VERSION"
but in all other places, it will be "KuberpultVersion"

medo-freiheit (approved):

PR #299, opened 2022-08-17
Add authentication to grpc api calls in frontend

mnishamk-freiheit:
If azure auth is enabled, validate that the grpc api calls have authorization metadata, they are signed by microsoft and that they come from the client and tenant provided in the values.yaml

coverage
image


pkg/auth/azure.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/MicahParks/keyfunc"
+	jwt "github.com/golang-jwt/jwt/v4"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+)
+
+var jwks *keyfunc.JWKS = nil
```
sven-urbanski-freiheit-com, line 33: please put this global state into a struct.

pkg/auth/azure.go
```diff
@@ -0,0 +1,138 @@
...
+var clientId string
+var tenantId string
+
```
sven-urbanski-freiheit-com, line 36: Ideally, there should be no global variables.
Could you make it so that they are passed in from outside? then server.go would keep the state.

mnishamk-freiheit: moved state to server.go

pkg/auth/azure_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	for _, tc := range tcs {
+		tc := tc
+		t.Run(tc.Name, func(t *testing.T) {
+			t.Parallel()
+			ctx := context.Background()
+			var jwks *keyfunc.JWKS
+			var err error
+			if !(tc.noInit) {
+				jwks, err = JWKSInitAzure(ctx)
```
sven-urbanski-freiheit-com, line 73: This means we are doing a several requests to `https://login.microsoftonline.com/common/discovery/v2.0/keys` right?
We might run into some sort of rate limiting. That could be dangerous, as our build would randomly fail.

sven-urbanski-freiheit-com: I would suggest that we store the keys, instead of requesting them again on each test.

sven-urbanski-freiheit-com (approved):

pkg/auth/azure_test.go
mnishamk-freiheit: updating in follow up pr

PR #302, opened 2022-08-18
Remove token validation from root path

mnishamk-freiheit:
without this, when loading the url for kuberpult, the react ui would not be displayed. Instead it would throw an error that invalid auth header specified. So to get access to react ui, this path shouldn't have authorization

### DIFF
--- a/pkg/auth/azure.go
+++ b/pkg/auth/azure.go
@@ -140,10 +140,13 @@ func StreamInterceptor(
 
 func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfunc.JWKS, clientId string, tenantId string) error {
 	token := req.Header.Get("authorization")
-	err := ValidateToken(token, jwks, clientId, tenantId)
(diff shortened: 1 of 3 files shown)


pkg/auth/azure.go
```diff
@@ -140,10 +140,13 @@ func StreamInterceptor(
 
 func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfunc.JWKS, clientId string, tenantId string) error {
 	token := req.Header.Get("authorization")
-	err := ValidateToken(token, jwks, clientId, tenantId)
-	if err != nil {
-		resp.WriteHeader(http.StatusUnauthorized)
-		resp.Write([]byte("Invalid authorization header provided"))
+	if req.URL.Path != "/" {
```
medo-freiheit, line 143: Why removing it from the root path, can you add more description please

mnishamk-freiheit: updated

PR #305, opened 2022-08-18
Remove auth for /static folders as well

mnishamk-freiheit:
react ui is still unable to load as static js files return 401.
only add auth for actual apis

### DIFF
--- a/pkg/auth/azure.go
+++ b/pkg/auth/azure.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/MicahParks/keyfunc"
@@ -140,7 +141,7 @@ func StreamInterceptor(
 
 func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfunc.JWKS, clientId string, tenantId string) error {
 	token := req.Header.Get("authorization")
-	if req.URL.Path != "/" {
+	if strings.HasPrefix(req.URL.Path, "/environments") {
 		err := ValidateToken(token, jwks, clientId, tenantId)
 		if err != nil {
 			resp.WriteHeader(http.StatusUnauthorized)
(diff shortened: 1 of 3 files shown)


pkg/auth/azure.go
```diff
@@ -140,7 +141,7 @@ func StreamInterceptor(
 
 func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfunc.JWKS, clientId string, tenantId string) error {
 	token := req.Header.Get("authorization")
-	if req.URL.Path != "/" {
+	if strings.HasPrefix(req.URL.Path, "/environments") {
```
sven-urbanski-freiheit-com, line 144: please extract this into a constant and give it a name or comment
```
// all paths mentioned here DO NOT require authentication:
const ignoredAuthoPrefixes = ["..."]
```

mnishamk-freiheit: updated and moved the ignored paths to server.go so that paths are known there instead of here

pkg/auth/azure.go
```diff
@@ -138,15 +139,23 @@ func StreamInterceptor(
 	return handler(srv, stream)
 }
 
-func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfunc.JWKS, clientId string, tenantId string) error {
+func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfunc.JWKS, clientId string, tenantId string, allowedPaths []string, allowedPrefixes []string) error {
```
sven-urbanski-freiheit-com, line 142: please add a test for the paths/prefixes

sven-urbanski-freiheit-com (approved):

pkg/auth/azure.go
mnishamk-freiheit: updated

sven-urbanski-freiheit-com (approved):

PR #307, opened 2022-08-19
Add scripts to store git history for releases, locks, deployments

tameremad:


infrastructure/scripts/metadata-migration/create-metadata-app-version.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -eu # protection
```
hannesg, line 3: Please use this line:

```
set -eou pipefail
```

infrastructure/scripts/metadata-migration/create-metadata-locks.sh
```diff
@@ -0,0 +1,49 @@
...
+  echo -----------------Application Locks------------------------
+
+  echo -----------------Environment Locks------------------------
+  # If $env has locks
+  if [ -d "$env"/locks ]; then
+    echo Adding metadata to "$(ls "$env"/locks | wc -l)" locks in "$env"
+    find "$env"/locks  -maxdepth 1 -mindepth 1 -type f -print0 | while IFS= read -r -d '' lock
+    do
+      echo Lock ID: "$(basename "$lock")" - Lock Message: "$(cat "$lock")"
+      LOCK_DIR="$(dirname "$lock")"/lock_"$(basename "$lock")"
```
hannesg, line 39: Why not just turn the file into a directory? The prefix `lock_` shouldn't be necessary.

infrastructure/scripts/metadata-migration/create-metadata-app-version.sh
```diff
@@ -0,0 +1,18 @@
...
+
+echo --------------------Versions------------------------------
+for env in environments/* # $(find environments  -maxdepth 1 -mindepth 1 -type d)
+do
+  for app in "$env"/applications/*
+  do
+    # If $app has version
+    echo Application Version: "$app"
+    if [ -d "$app"/version ]; then
+      git log -1 --date=iso-strict --format="%ad" -- "$app"/version > "$app"/deploy_date
```
hannesg, line 14: I recall that we discussed removing this date entirely. Do we still need it?

infrastructure/scripts/metadata-migration/create-metadata-app-version.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
hannesg, line 1: Please use this:

```
#!/usr/bin/env bash
```

hannesg (approved):

infrastructure/scripts/metadata-migration/create-metadata-app-version.sh
tameremad: I thought we could use it in the future as the deploy date. and as this is a one-time migration, creating this now would be better? 

also, the backend can write the deploy date for newer deployments before we find a good way to show it in the UI. 

I think the feature is nice but it's not urgent. 
in the future:
* we can uncomment this field in the API and read the date from the repository. 
* then it's available in the frontend. 
* then we find a good way to show this in the UI.

PR #310, opened 2022-08-23
 do not show delete queue btn if there is no deployed version

medo-freiheit:


services/frontend-service/src/legacy-ui/ReleaseDialog.tsx
```diff
@@ -427,6 +427,11 @@ const ReleaseEnvironment: VFC<{
         return 1;
     });
 
+    const envHasAppDeployed = overview.environments[environmentName].applications[applicationName];
+    if (!envHasAppDeployed) {
```
tameremad, line 431: When is this condition true?
From what I see here, it think it always has a non-zero value.
https://github.com/freiheit-com/kuberpult/blob/main/services/cd-service/pkg/service/overview.go#L92-L95

medo-freiheit: When creating a new env, at first it will contain no apps, in this case, it will be true

tameremad (approved):

PR #311, opened 2022-08-23
SRX-EN3BAY Add indicator on the active Page in the left side bar

leandro-freiheit:


leandro-freiheit:
I noticed that in the navListItem.tsx when `pathname.startsWith` is used, the result will always return `false`. This happens because `useLocation()` returns the entire the path `/v2/...`

services/frontend-service/src/ui/components/navigation/navListItem.tsx
```diff
@@ -33,21 +44,24 @@ export const NavListItem = (props: { className?: string; to: string; icon?: JSX.
...
-            ref={control}
-            to={to}
-            tabIndex={pathname.startsWith(to) ? 0 : undefined}>
-            <div className="mdc-list-item__ripple" />
-            {icon &&
-                cloneElement(icon, {
-                    key: 'icon',
-                })}
-        </Link>
+        <div className="mdc-list-item-container" style={{ display: 'flex' }}>
```
medo-freiheit, line 47: Please do not use inline styling move it to scss file
```suggestion
        <div className="mdc-list-item-container" >
```

medo-freiheit (changes requested):

services/frontend-service/src/ui/components/navigation/navListItem.tsx
```diff
@@ -19,6 +19,17 @@ import { cloneElement, useEffect, useRef } from 'react';
 import { MDCRipple } from '@material/ripple';
 import { Link, useLocation } from 'react-router-dom';
 
+export const NavbarIndicator = (props: { pathname: string; to: string }) => {
+    const { pathname, to } = props;
+    return (
+        <div
+            className={classNames(
+                'mdc-list-item-indicator',
+                pathname.startsWith('/v2/' + to) ? 'mdc-list-item-indicator--activated' : ''
```
medo-freiheit, line 28:
```suggestion
                pathname.startsWith(`/v2/${to}`) ? 'mdc-list-item-indicator--activated' : ''
```

services/frontend-service/src/ui/components/navigation/navListItem.tsx
```diff
@@ -33,21 +44,24 @@ export const NavListItem = (props: { className?: string; to: string; icon?: JSX.
...
-                    key: 'icon',
-                })}
-        </Link>
+        <div className="mdc-list-item-container" style={{ display: 'flex' }}>
+            <NavbarIndicator pathname={pathname} to={to} />
+            <Link
+                aria-label={to}
+                className={classNames(
+                    'mdc-list-item',
+                    pathname.startsWith(to) ? 'mdc-list-item--activated' : '',
```
medo-freiheit, line 53:
```suggestion
                    pathname.startsWith(`/v2/${to}`) ? 'mdc-list-item--activated' : '',
```

medo-freiheit: please fix it here too

services/frontend-service/src/ui/components/navigation/navListItem.tsx
medo-freiheit: please use ES6 template literal {``}.

services/frontend-service/src/ui/components/navigation/navListItem.tsx
leandro-freiheit: It is fixed

services/frontend-service/src/ui/components/navigation/navListItem.tsx
leandro-freiheit: It is corrected

services/frontend-service/src/ui/NavigationBar/NavigationBar.scss
```diff
@@ -13,6 +13,20 @@
   }
 
   .mdc-drawer__content {
-    height: 100%;
+    height: 95%;
```
tameremad, line 16: Why is this at `95%`? is it because of the top bar? or?
Can you move it into a variable and define it in https://github.com/freiheit-com/kuberpult/blob/main/services/frontend-service/src/assets/_variables.scss?

services/frontend-service/src/ui/NavigationBar/NavigationBar.scss
```diff
@@ -13,6 +13,20 @@
...
+  }
+
+  .mdc-list-item-indicator {
+    background-color: transparent;
+    width: 5%;
```
tameremad, line 21: In general, anything that's more specific or not intuitive like `100%` should go into a variable. 
Please take a look here: https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/WORKSPACE_DOCUMENT/-Mn5HLuHv2wjoq0QKPax/

services/frontend-service/src/ui/NavigationBar/NavigationBar.scss
```diff
@@ -13,6 +13,20 @@
...
+  }
+
+  .mdc-list-item-indicator--activated {
+    background-color: orange;
+    transition: background-color 0.2s ease-in-out;
+  }
+
+  .mdc-list-item-container {
+    display: flex
```
tameremad, line 30: Please add:
`height: 100%;`
so that the container can also take the full size of the parent (the three navigation buttons fill up the space of the drawer instead of just the height of the content which is just the icons.

services/frontend-service/src/ui/NavigationBar/NavigationBar.scss
tameremad: Also Please Move the css for the indicator and for the list-item-container into a new scss file in the navigation directory: 
https://github.com/freiheit-com/kuberpult/tree/main/services/frontend-service/src/ui/components/navigation
`/navListItem.scss` for example.

leandro-freiheit: I placed that value at 95% because of the the width of the indicator. This to avoid that any overflow would occur

services/frontend-service/src/ui/NavigationBar/NavigationBar.scss
leandro-freiheit: Okay, fixed!

services/frontend-service/src/ui/NavigationBar/NavigationBar.scss
leandro-freiheit: I fixed it according to your suggestion!

services/frontend-service/src/ui/NavigationBar/NavigationBar.scss
leandro-freiheit: Totally agreed! Already done it.

PR #316, opened 2022-08-26
SRX-APRASR Provide PR number in Overview Services and GitUrl configuration

leandro-freiheit:
Added GitUrl variable in frontend config, modified the GetFrontEnd Config to accept the variable, added PR number extraction and modified services to send the the PR number


hannesg:
Given that you are just providing one url and a pr number, how do you want to construct the merge request url in the end?

We want to support people not using github and teams sharing one kuberpult instance with multiple source repositories. I guess you rather need something like an url template per app.

leandro-freiheit:
The PR number is associated with each release, for each application. Since we have the base git url for the repo (https://github.com/freiheit-com/kuberpult), with each PR number, we could obtain the git url for the pull request. But yeah, this requires that the source repository is a git repo.

services/frontend-service/pkg/cmd/server.go
```diff
@@ -14,6 +14,24 @@ You should have received a copy of the GNU General Public License
 along with kuberpult.  If not, see <http://www.gnu.org/licenses/>.
 
 Copyright 2021 freiheit.com*/
+/*
```
mnishamk-freiheit, line 17: why does this keep getting added again?

leandro-freiheit: As far as I know, it is the ESLint formatting that always does that

services/cd-service/pkg/service/overview.go
```diff
@@ -307,3 +320,22 @@ func transformSyncWindows(syncWindows []config.ArgoCdSyncWindow, appName string)
 	}
 	return envAppSyncWindows, nil
 }
+
+func extractPrNumber(sourceMessage string) *string {
```
mnishamk-freiheit, line 324: is the source message added by git? or is this from some script (thats written by us). Does it have pr number already.

Is this specific to github, or would all have it?

Also couldn't this be done with just the regex
```
	re := regexp.MustCompile("\\((#(([0-9])+))\\)")
	str := "something (#123)"
	res := re.FindAllStringSubmatch(str, 1)
	fmt.Println("Message :", res[0][2])
```

mnishamk-freiheit:
is the part that uses urltemplate to make the url to be done in future?

services/cd-service/pkg/service/overview.go
leandro-freiheit: The source message is added by git, yes. The PR number is present in the commit header for each commit.

Only the regex would not be enough because the commit header may contain '(', '(' or '#' in the message itself. There must be a filtering for those cases.

leandro-freiheit:
Yes, it is. More specifically to support multiple source repos

services/cd-service/pkg/service/overview.go
mnishamk-freiheit: yes but if there are multiple items of <left-paranthesis><hashtag>numbers<right-paranthesis>, do you do anything about it?

because thats the only thing the above regex looks for

leandro-freiheit: @mnishamk-freiheit the option you suggested has some problems. First of all, it does not work when there are multiple PR numbers for one commit. Additionally, it returns the first PR number that appears when in truth we want the last PR number.

leandro-freiheit: > yes but if there are multiple items of <left-paranthesis><hashtag>numbers<right-paranthesis>, do you do anything about it?
> 
> because thats the only thing the above regex looks for

Yes, the option that I inserted does that, precisely

mnishamk-freiheit: in above res is an array with all the pr numbers, res[0] is the first pr number.

leandro-freiheit: Your option works if the the PR numbers are within brackets for each one. For example, "Commit header (#PRNUMBER1) (#PRNUMBER2)". But can't the PR numbers be stored like "Commit header (#PRNUMBER1 #PRNUMBER2)". Because my implementation was done taking both cases into account.

mnishamk-freiheit: I'm still not sure if regex is the way to go or not, but since you'd used regex, figured could use regex for the whole thing. check with @hannesg  once

leandro-freiheit: Will do!

leandro-freiheit: @mnishamk-freiheit There is one case that my solution does not take into account. And after looking better the expression use for the regex, I think your approach is better and simpler

charts/kuberpult/values.yaml.tpl
```diff
@@ -9,6 +9,8 @@ git:
 hub: europe-west3-docker.pkg.dev/fdc-public-docker-registry/kuberpult
 tag: "$VERSION"
 
+sourceRepoUrl: ""
```
medo-freiheit, line 12: Can you move it under git

charts/kuberpult/templates/frontend-service.yaml
```diff
@@ -56,6 +56,8 @@ spec:
           value: {{ .Values.argocd.baseUrl | quote }}
         - name: KUBERPULT_VERSION
           value: {{ .Values.tag | quote}}
+        - name: KUBERPULT_SOURCE_REPO_URL
+          value: {{ .Values.sourceRepoUrl | quote}}
```
medo-freiheit, line 60:
```suggestion
          value: {{ .Values.git.sourceRepoUrl | quote}}
```

services/cd-service/pkg/config/config.go
```diff
@@ -14,6 +14,24 @@ You should have received a copy of the GNU General Public License
 along with kuberpult.  If not, see <http://www.gnu.org/licenses/>.
 
 Copyright 2021 freiheit.com*/
```
medo-freiheit, line 16: Please fix this duplication

services/cd-service/pkg/repository/repository.go
```diff
@@ -1040,6 +1040,21 @@ func (s *State) GetApplicationTeamOwner(application string) (string, error) {
...
 
+func (s *State) GetApplicationSourceRepoUrl(application string) (string, error) {
+	appDir := applicationDirectory(s.Filesystem, application)
+	appSourceRepoUrl := s.Filesystem.Join(appDir, "sourceRepoUrl")
+
+	if url, err := readFile(s.Filesystem, appSourceRepoUrl); err != nil {
+		if os.IsNotExist(err) {
+			return "", nil
+		} else {
+			return "", fmt.Errorf("error while reading team owner file for application %v found: %w", application, err)
```
medo-freiheit, line 1051:
```suggestion
			return "", fmt.Errorf("error while reading sourceRepoUrl file for application %v found: %w", application, err)
```

services/cd-service/pkg/repository/time.go
```diff
@@ -0,0 +1,44 @@
...
+Kuberpult is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with kuberpult.  If not, see <http://www.gnu.org/licenses/>.
+
+Copyright 2021 freiheit.com*/
+package repository
```
medo-freiheit, line 17: What is the use of `time.go` ?

medo-freiheit (changes requested):

services/cd-service/pkg/repository/time.go
leandro-freiheit: No usage, it was a leftover file that was added accidentally

services/cd-service/pkg/config/config.go
leandro-freiheit: Fixed!

charts/kuberpult/templates/frontend-service.yaml
leandro-freiheit: Fixed!

charts/kuberpult/values.yaml.tpl
leandro-freiheit: Yes, of course!

medo-freiheit (approved):

PR #318, opened 2022-08-29
Add documentation for scripts and return nil when locks are not found

tameremad:
quote from the story: 
> Do not read old data. If data is not found, just return nil/"". 
> Since we have the migration script, this should be fine.

### DIFF
--- a/infrastructure/scripts/metadata-migration/Migration.md
+++ b/infrastructure/scripts/metadata-migration/Migration.md
@@ -5,15 +5,18 @@ We can improve this, by storing data explicitly, instead of relying on the git-h
 We do this anyway for some things like author.
 * The releases script goes over all releases for all applications, and writes the commit date
   (the date when this release was created) to a new file `created_at`
-* The locks script goes over all env and app locks, removes the lock **file** and creates a **directory** 
+* The locks script goes over all env and app locks, removes the lock **file** and creates a **directory**
   with the same name instead. Then it writes the commit date (the date when this lock was created)
   as well as the author name and email and the message into files in the new lock directory.
-  
-  
-  
+* The scripts only create the files and do not add or push to GitHub.
+
 The new kuberpult will only read the data in the new format and will ignore old data which is normally deleted by the script.
 
 ## Deployment with Downtime
 1) Shut down kuberpult: Delete StatefulSet
 2) Write data to manifest repo, this could happen locally during downtime
-3) Deploy new kuberpult version that reads new data
+   * clone the manifests repository depends on where Kuberpult is deployed.
+   * cd and run the scripts in the top dir.
+   * Add the created files. and commit the changes to the manifests
+     repo (use a different branch to get approval first).
+3) Merge and then Deploy new kuberpult version that reads new data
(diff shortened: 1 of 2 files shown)


infrastructure/scripts/metadata-migration/Migration.md
```diff
@@ -0,0 +1,19 @@
...
+  with the same name instead. Then it writes the commit date (the date when this lock was created)
+  as well as the author name and email and the message into files in the new lock directory.
+  
+  
+  
+The new kuberpult will only read the data in the new format and will ignore old data which is normally deleted by the script.
+
+## Deployment with Downtime
+1) Shut down kuberpult: Delete StatefulSet
+2) Write data to manifest repo, this could happen locally during downtime
```
mnishamk-freiheit, line 18: could you add how to as well

git clone
script commands
git push

or whatever set of commands are required to do this.

services/cd-service/pkg/repository/repository.go
```diff
@@ -741,35 +741,45 @@ type Lock struct {
...
-		return nil, err
-	}
-	email, err := readFile(fs, fs.Join(lockDir, "created_by_email"))
-	if err != nil {
-		return nil, err
+	lock := &Lock{}
+
+	if cnt, err := readFile(fs, fs.Join(lockDir, "message")); err != nil {
+		if !os.IsNotExist(err) {
+			return nil, err
```
mnishamk-freiheit, line 748: is there any way to identify between the cases
- lock doesn't exist
and
- migration not done yet

infrastructure/scripts/metadata-migration/Migration.md
```diff
@@ -5,15 +5,18 @@ We can improve this, by storing data explicitly, instead of relying on the git-h
...
 The new kuberpult will only read the data in the new format and will ignore old data which is normally deleted by the script.
 
 ## Deployment with Downtime
 1) Shut down kuberpult: Delete StatefulSet
 2) Write data to manifest repo, this could happen locally during downtime
-3) Deploy new kuberpult version that reads new data
+   * clone the manifests repository depends on where Kuberpult is deployed.
+   * cd and run the scripts in the top dir.
+   * Add the created files. and commit the changes to the manifests
+     repo (use a different branch to get approval first).
```
mnishamk-freiheit, line 21:
```suggestion
     repo (use a different branch to get approval first).
     ```bash
     git clone $MANIFEST_REPO
     git checkout -b kuberpult_migrate
     ./create-metadata-locks.sh
     ./create-metadata-releases.sh
     git add .
     git commit -m "migration to new kuberpult"
     git push
     ```
```

services/cd-service/pkg/repository/repository.go
tameremad: I don't think this is very necessary. We can't really differentiate between the two errors in any meaningful way. What comes to mind is checking if the old lock is there. but that is not applicable. (the old lock is a file and the new lock is a directory with the same name. )

mnishamk-freiheit: okay, so as of now if migration is not done you'll lose locks. that seems a bit dangerous. Which is the only reason I ask

mnishamk-freiheit (approved):

PR #320, opened 2022-08-30
Remove persistent volume and use empty dir

tameremad:


tameremad:
This is already done and merged in another PR

PR #329, opened 2022-09-02
Add Release Card Component

tameremad:
image
image


services/frontend-service/src/ui/ReleaseCard/ReleaseCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+            MDComponent.current = new MDCRipple(control.current);
+        }
+        return () => MDComponent.current?.destroy();
+    }, []);
+
+    return (
+        <div className={classNames('mdc-card release-card', className)}>
+            <div className="release-card__header">
+                <div className="release__title mdc-typography--headline6">{title}</div>
+                <Button className="release__hash" label={hash} />
```
hannesg, line 48: The hash can be empty. We should probably hide this button then.

services/frontend-service/src/ui/ReleaseCard/ReleaseCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+        if (control.current) {
+            MDComponent.current = new MDCRipple(control.current);
+        }
+        return () => MDComponent.current?.destroy();
+    }, []);
+
+    return (
+        <div className={classNames('mdc-card release-card', className)}>
+            <div className="release-card__header">
+                <div className="release__title mdc-typography--headline6">{title}</div>
```
hannesg, line 47: What will we do when the title is very long?

hannesg (approved):

services/frontend-service/src/ui/ReleaseCard/ReleaseCard.tsx
tameremad: I'm curious when this could happen? 
for the undeploy versions?

Medo-fdc (approved):
I think we should move ReleaseCard directory inside Components directory

services/frontend-service/src/ui/ReleaseCard/ReleaseCard.tsx
tameremad: image

medo-freiheit: try just to add tails and only one row e.g ( something something ...)

tameremad: I think it's important for devs to see the entire title.

PR #335, opened 2022-09-05
Use material themes

tameremad:


tameremad:
This is not necessary for kuberpult now. it can work with setting the variables directly in _variables.scss

PR #343, opened 2022-09-09
SRX-381AY8 Kuberpult configuration to use login from Azure authentication

leandro-freiheit:


pkg/auth/azure.go
```diff
@@ -152,7 +152,13 @@ func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfu
 		}
 	}
 
-	err := ValidateToken(token, jwks, clientId, tenantId)
+	claims, err := ValidateToken(token, jwks, clientId, tenantId)
+
+	if _, ok := claims["tid"]; ok {
+		req.Header.Set("username", claims["name"].(string))
+		req.Header.Set("email", claims["email"].(string))
```
mnishamk-freiheit, line 159: email isn't a valid field. use preferred_username instead

pkg/auth/azure.go
```diff
@@ -152,7 +152,13 @@ func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfu
 		}
 	}
 
-	err := ValidateToken(token, jwks, clientId, tenantId)
+	claims, err := ValidateToken(token, jwks, clientId, tenantId)
+
+	if _, ok := claims["tid"]; ok {
```
mnishamk-freiheit, line 157: tid will be present whether or not its an access token. check !appid instead.

appid is given if access token

https://github.com/freiheit-com/kuberpult/commit/14251afe7e133ad5c61549ac3cc44b48ec45f3a0

services/frontend-service/pkg/cmd/server.go
```diff
@@ -264,9 +264,25 @@ func getRequestAuthor(ctx context.Context, r *http.Request) *auth.User {
 	return u
 }
 
+func getRequestAuthorFromAzure(r *http.Request) *auth.User {
+	username := r.Header.Get("username")
```
mnishamk-freiheit, line 268: add empty case as well

mnishamk-freiheit (changes requested):

services/frontend-service/pkg/cmd/server.go
leandro-freiheit: Fixed!

pkg/auth/azure.go
leandro-freiheit: Fixed, as now the `aud` claim is verified since it is used to check the App ID, according to https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens#payload-claims

pkg/auth/azure.go
leandro-freiheit: It is a valid field according to https://docs.microsoft.com/en-us/azure/active-directory/develop/id-tokens#payload-claims but it is true that preferred_username is a better option to use. Fixed!

mnishamk-freiheit (approved):

PR #347, opened 2022-09-12
Add service lanes and show the new kuberpult homepage

tameremad:
image

* subscribing to the overview service at `index.tsx` is enough to track all changes and stream updates
* the first release cards belong to undeploy versions created locally.

### DIFF
--- a/services/frontend-service/src/ui/App/index.tsx
+++ b/services/frontend-service/src/ui/App/index.tsx
@@ -18,13 +18,40 @@ import { NavigationBar } from '../components/NavigationBar/NavigationBar';
 import { TopAppBar } from '../components/TopAppBar/TopAppBar';
 import { PageRoutes } from './PageRoutes';
 import '../../assets/app-v2.scss';
+import * as React from 'react';
+import { PanicOverview, UpdateOverview } from '../utils/store';
+import { useApi } from '../utils/GrpcApi';
 
-export const App: React.FC = () => (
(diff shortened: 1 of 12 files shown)


hannesg (approved):

services/frontend-service/src/ui/App/index.tsx
```diff
@@ -18,13 +18,40 @@ import { NavigationBar } from '../components/NavigationBar/NavigationBar';
...
-        <NavigationBar />
-        <div className="mdc-drawer-app-content">
-            <TopAppBar />
-            <PageRoutes />
+export const App: React.FC = () => {
+    const api = useApi;
+    React.useEffect(() => {
+        const subscription = api
+            .overviewService()
+            .StreamOverview({}) // TODO TE: add auth header
```
sven-urbanski-freiheit-com, line 30: please add the ID of the story, where this `TODO` will be done @tameremad

sven-urbanski-freiheit-com: (as a comment in the code)

tameremad: I actually looked into this. I saw that we have a little complicated contexts setup for authentication. as I didn't want to move anything that could break I didn't move them. I will meet with Nisham to discuss them. I think some could be simplified.

tameremad: if that proved too difficult I will refine a story for it

PR #355, opened 2022-09-19
Force using gpg when on Azure authentication

tameremad:


services/cd-service/pkg/cmd/server.go
```diff
@@ -78,6 +79,9 @@ func RunServer() {
 		if err != nil {
 			logger.FromContext(ctx).Fatal("pgp.read.error", zap.Error(err))
 		}
+		if c.AzureEnableAuth && pgpKeyRing == nil {
+			logger.FromContext(ctx).Fatal("azure.auth.error: pgpKeyRing is required to authenticate manifests")
```
sven-urbanski-freiheit-com, line 83:
```suggestion
			logger.FromContext(ctx).Fatal("azure.auth.error: pgpKeyRing is required to authenticate manifests when 'KUBERPULT_AZURE_ENABLE_AUTH" is true")
```

sven-urbanski-freiheit-com (approved):

PR #365, opened 2022-09-27
Add pgp signature verification for put/delete environment lock requests

tameremad:


pkg/auth/azure.go
```diff
@@ -161,12 +161,16 @@ func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfu
 			return nil
 		}
 	}
-	// Skip azure authentication with ID for `/release` and `/releasetrain` endpoints. The requests will be validated with pgp signature
+	// Skip azure authentication with ID for `/release`, `/releasetrain` and `/locks`  endpoints. The requests will be validated with pgp signature
 	// usage in requests from outside the cluster (e.g. by GitHub Actions and the publish.sh script).
-	releaseTrainRx := regexp.MustCompile("/environments/[^/]*/releasetrain")
+	releaseTrainRx := regexp.MustCompile("/environments/[^/]+/releasetrain")
```
hannesg, line 166: The regular expressions are not anchored. That means, it also matches '/foobar/environments/a/releasetrain/barbaz'

hannesg (approved):

pkg/auth/azure.go
tameremad: hmm okay. I'll fix it.

PR #371, opened 2022-10-05
do a different reload (polling)

sven-urbanski-freiheit-com:


services/frontend-service/src/legacy-ui/App/index.tsx
```diff
@@ -51,17 +51,53 @@ export const Spinner: React.FC<any> = (props: any) => {
...
-        []
+        (api: any) => api.overviewService().GetOverview({}),
+        // eslint-disable-next-line
+        [count]
     );
-    const overview = useObservable<api.GetOverviewResponse>(getOverview);
+    const overview = useUnary<api.GetOverviewResponse>(getOverview);
+    useInterval(() => {
+        setCount(count + 1);
+    }, 10 * 1000);
```
sven-urbanski-freiheit-com: 10 sec is pretty often.
I think we should increase this to 20 or 30 and ensure that after a request (cart submit) GetOverview is called again.

services/frontend-service/src/legacy-ui/App/index.tsx
```diff
@@ -52,16 +52,37 @@ export const Spinner: React.FC<any> = (props: any) => {
...
+    const { authHeader } = React.useContext(AuthTokenContext);
+
+    const updateOverview = React.useCallback(() => {
+        api.overviewService()
+            .GetOverview({}, authHeader)
+            .then(
+                (result) => setOverview({ result, state: 'resolved' }),
+                (error) => setOverview({ error, state: 'rejected' })
+            );
+    }, [api, authHeader]);
```
tameremad, line 67: Any `useUnary` can be replaced with this callback. 
previously, we had tons of API calls. so a template was useful. now, not so much.

services/frontend-service/src/legacy-ui/App/index.tsx
```diff
@@ -52,16 +52,37 @@ export const Spinner: React.FC<any> = (props: any) => {
 };
 
 const GetOverview = (props: { children: (r: api.GetOverviewResponse) => JSX.Element }): JSX.Element | null => {
-    const getOverview = React.useCallback(
-        (api: any, authHeader: AuthHeaderType) => api.overviewService().StreamOverview({}, authHeader),
-        []
-    );
-    const overview = useObservable<api.GetOverviewResponse>(getOverview);
+    const [overview, setOverview] = useState<UnaryState<api.GetOverviewResponse>>({ state: 'pending' });
+    const Tdelay = 10 * 1000;
```
sven-urbanski-freiheit-com, line 56:
```suggestion
    const reloadDelayMillis = 10 * 1000;
```

services/frontend-service/src/legacy-ui/App/index.tsx
```diff
@@ -52,16 +52,37 @@ export const Spinner: React.FC<any> = (props: any) => {
...
     switch (overview.state) {
         case 'resolved':
-            return props.children(overview.result);
+            backupState.current = overview.result;
+            return props.children(backupState.current);
         case 'rejected':
-            return <div>Error: {JSON.stringify(overview.error)}</div>;
+            // eslint-disable-next-line no-console
+            console.log('2 restarting streamoverview due to error: ', overview.error);
+            // setRestart(!restart);
```
sven-urbanski-freiheit-com, line 84:
```suggestion
```

services/cd-service/pkg/service/overview.go
```diff
@@ -210,6 +210,7 @@ func (o *OverviewServiceServer) getOverview(
 		Environments: map[string]*api.Environment{},
 		Applications: map[string]*api.Application{},
 	}
+	//return nil, internalError(ctx, fmt.Errorf("Hello world error."))
```
sven-urbanski-freiheit-com, line 213:
```suggestion
```

services/frontend-service/src/legacy-ui/App/index.tsx
```diff
@@ -52,16 +52,37 @@ export const Spinner: React.FC<any> = (props: any) => {
...
+
     switch (overview.state) {
         case 'resolved':
-            return props.children(overview.result);
+            backupState.current = overview.result;
+            return props.children(backupState.current);
         case 'rejected':
-            return <div>Error: {JSON.stringify(overview.error)}</div>;
+            // eslint-disable-next-line no-console
+            console.log('2 restarting streamoverview due to error: ', overview.error);
```
sven-urbanski-freiheit-com, line 83: I'm fine with the code as it is. I still think we need to make sure that the state is reloaded when a cart submit is successful.

tameremad: reload on submit is a more difficult than it seems. the overview currently is always a passed along local variable. but to have anything like that it would require us to use some sort of context or store or any other similar solution. and by then it seems too tricky. we can talk about this

hannesg:
Do we really want to make polling the default? I think switching to websocket transport could actually help because for websockets, it's expected to stay open for a long time.

hannesg:
> > Do we really want to make polling the default? I think switching to websocket transport could actually help because for websockets, it's expected to stay open for a long time.
> 
> I'm more than happy to implement websockets for v2 (Edit: typo!). But for now, we need a simple quick solution that does the trick. We put all the polling code in the `legacy-ui` directory, so it won't just switch over to the new ui.

Okay. Well, then let's do this.

services/frontend-service/src/legacy-ui/App/RefreshStore.ts
```diff
@@ -0,0 +1,36 @@
...
+        this.doRefresh = newState;
+    }
+
+    shouldRefresh() {
+        return this.doRefresh;
+    }
+}
+
+const refreshStore = new RefreshStore();
+export default refreshStore;
```
hannesg, line 36: Global variables are global :(

hannesg (approved):
Please, somehow make sure that code gets really removed soon.

PR #377, opened 2022-10-11
Refresh overview on login

tameremad:
#### First:
when az auth is on, the authorization token might take a little time to be ready (basically till we login) and so we need to refresh the overview again after getting the token.

#### Also:
fix a bug with `react-error-overlay` where react sometimes fails with `Uncaught ReferenceError: process is not defined` after using an `iframe`. 

source: 
* https://github.com/facebook/create-react-app/issues/11773
* https://stackoverflow.com/questions/70368760/react-uncaught-referenceerror-process-is-not-defined

sorry about the typos will fix before merging.

### DIFF
--- a/services/frontend-service/pnpm-lock.yaml
+++ b/services/frontend-service/pnpm-lock.yaml
@@ -1,5 +1,8 @@
 lockfileVersion: 5.4
 
+overrides:
+  react-error-overlay: 6.0.9
+
 importers:
 
   .:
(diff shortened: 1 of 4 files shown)


services/frontend-service/pnpm-lock.yaml
```diff
@@ -11530,17 +11539,17 @@ packages:
       open: 7.4.2
       pkg-up: 3.1.0
       prompts: 2.4.0
-      react-error-overlay: 6.0.11
+      react-error-overlay: 6.0.9
```
mouhsen-ibrahim, line 11542: Why the downgrade?

mouhsen-ibrahim (approved):

services/frontend-service/pnpm-lock.yaml
tameremad: The `error-overlay` pacakge was updated to support webpack v5, and is now incompatible with react-scripts v4 which we still need.

sven-urbanski-freiheit-com: Fine for now, but next time, please put such a comment into the code and git commit message

PR #379, opened 2022-10-16
Add Azure Auth to V2

tameremad:


services/frontend-service/src/ui/utils/AzureAuthProvider.tsx
```diff
@@ -0,0 +1,123 @@
...
+    useIsAuthenticated,
+    UnauthenticatedTemplate,
+    useMsal,
+} from '@azure/msal-react';
+import { GetFrontendConfigResponse } from '../../api/api';
+import { createStore } from 'react-use-sub';
+import { grpc } from '@improbable-eng/grpc-web';
+
+export const [useFrontendConfig, UpdateFrontendConfig] = createStore({} as GetFrontendConfigResponse);
+export const [useReady, UpdateReady] = createStore({ config: 'waiting', auth: 'waiting' });
```
sven-urbanski-freiheit-com, line 32: can we use an enum here instead of just strings? (for both `config` and `waiting`)

sven-urbanski-freiheit-com (approved):

services/frontend-service/src/ui/utils/AzureAuthProvider.tsx
tameremad: I'd say probably yes. I was actually interested in getting feedback about this. Does this `Ready` thing make any sense? is it best practices? any better way to do it? 

the goal of this is to only subscribe to the overview when we got everything ready.

my other idea is to add a flag to the other store, e.g. (useFrontendConfig)?

services/frontend-service/src/ui/App/index.tsx
```diff
@@ -21,38 +21,59 @@ import '../../assets/app-v2.scss';
...
+            .then(
                 (result) => {
-                    UpdateOverview.set(result);
-                    PanicOverview.set({ error: '' });
+                    UpdateFrontendConfig.set({ configs: result, configReady: true });
                 },
-                (error) => PanicOverview.set({ error: JSON.stringify(error) })
+                (error) => {
+                    // eslint-disable-next-line no-console
+                    alert('Error: Cannot connect to server!\n' + error);
```
sven-urbanski-freiheit-com, line 39: can't we use the normal way to display errors? An `alert` is quite intrusive

services/frontend-service/src/ui/utils/AzureAuthProvider.tsx
sven-urbanski-freiheit-com: > the goal of this is to only subscribe to the overview when we got everything ready.

is there a state where only half of it is ready? If yes, it might make sense to have a flag/state like this. 
But what is this in-between state? I'm not sure if I get that.

services/frontend-service/src/ui/App/index.tsx
tameremad: Added a [story](https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/STORY/-NEkhGYEXWbhiSAJ216v/)

PR #396, opened 2022-11-09
SRX-XYKGST Don't deploy queued version when unlocking

mouhsen-ibrahim:
* rename Queue Lock Behavior to Record
* Remove "Delete Queue" button and action from UI
* Stop deploying queued version when deleting a lock

### DIFF
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -1071,33 +1071,3 @@ func readFile(fs billy.Filesystem, path string) ([]byte, error) {
 		return io.ReadAll(file)
 	}
 }
-
-// ProcessQueue checks if there is something in the queue
-// deploys if necessary
-// deletes the queue
-func (s *State) ProcessQueue(ctx context.Context, fs billy.Filesystem, environment string, application string) (string, error) {
-	queuedVersion, err := s.GetQueuedVersion(environment, application)
-	queueDeploymentMessage := ""
-	if err != nil {
-		// could not read queued version.
-		return "", err
-	} else {
-		if queuedVersion == nil {
-			// if there is no version queued, that's not an issue, just do nothing:
-			return "", nil
-		}
-
-		currentlyDeployedVersion, err := s.GetEnvironmentApplicationVersion(environment, application)
-		if err != nil {
-			return "", err
-		}
-
-		if currentlyDeployedVersion != nil && *queuedVersion == *currentlyDeployedVersion {
-			// delete queue, it's outdated! But if we can't, that's not really a problem, as it would be overwritten
-			// whenever the next deployment happens:
-			err = s.DeleteQueuedVersion(environment, application)
-			return fmt.Sprintf("deleted queued version %d because it was already deployed. app=%q env=%q", *queuedVersion, application, environment), err
-		}
-	}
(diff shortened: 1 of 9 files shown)


services/cd-service/pkg/repository/repository.go
```diff
@@ -1098,30 +1097,6 @@ func (s *State) ProcessQueue(ctx context.Context, fs billy.Filesystem, environme
 			// whenever the next deployment happens:
 			err = s.DeleteQueuedVersion(environment, application)
 			return fmt.Sprintf("deleted queued version %d because it was already deployed. app=%q env=%q", *queuedVersion, application, environment), err
```
tameremad, line 1099: Why not remove the entire `ProcessQueue` function or make it `return "", nil`?

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -1458,26 +1459,22 @@ func TestTransformer(t *testing.T) {
...
 					t.Fatal(err)
 				}
-				if i == nil {
-					t.Errorf("unexpected version: expected 1, actual nil")
-				} else {
-					if *i != 1 {
-						t.Errorf("unexpected version: expected 1, actual %d", *i)
-					}
+				if i != nil {
+					t.Errorf("unexpected version: unexpected nil")
```
tameremad, line 1470: Can you explain the error message, please? 

```suggestion
					t.Errorf("unexpected version %d: expected: nil", i)
```

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -1458,26 +1459,22 @@ func TestTransformer(t *testing.T) {
...
 				}
 				q, err := s.GetQueuedVersion("production", "test")
 				if err != nil {
 					t.Fatal(err)
 				}
-				if q != nil {
-					t.Errorf("unexpected version: expected nil, actual %d", *q)
+				if q == nil {
+					t.Errorf("unexpected version: expected %d, actual nil", *q)
```
tameremad, line 1477: I think the expected value here is 1. 
```suggestion
					t.Errorf("unexpected version: expected 1, actual nil")
```

services/frontend-service/src/legacy-ui/ReleaseDialog.tsx
```diff
@@ -366,17 +352,6 @@ const ReleaseEnvironment: VFC<{
...
-            deleteQueue: {
-                application: applicationName,
-                environment: environmentName,
-                currentlyDeployedVersion: currentlyDeployedVersion,
-            },
-        }),
-        [applicationName, environmentName, currentlyDeployedVersion]
-    );
-
     const queuedVersion = overview.environments[environmentName].applications[applicationName]?.queuedVersion || 0;
```
tameremad, line 355: Question: Should we remove the queuedVersion for here as well?

mouhsen-ibrahim: The story only included removing the button and not the message

services/cd-service/pkg/repository/repository.go
mouhsen-ibrahim: We still need it to delete the queue if necessary

tameremad (approved):
:+1:

PR #410, opened 2022-12-07
Add hidden releases indicator

mouhsen-ibrahim:


services/frontend-service/src/ui/components/ServiceLane/ServiceLane.tsx
```diff
@@ -23,6 +23,21 @@ import { Application } from '../../../api/api';
 export const ServiceLane: React.FC<{ application: Application }> = (props) => {
     const { application } = props;
     const releases = useDeployedReleases(application.name);
+    const releases_lane = releases.map((rel, index) => {
+        if (index > 0) {
+            const diff = releases[index - 1] - rel - 1;
```
medo-freiheit, line 28: Are you trying to subtract the whole release here? Should it be something like releases[index - 1].version - rel .version- 1

services/frontend-service/src/ui/components/ServiceLane/ServiceLane.tsx
```diff
@@ -23,6 +23,21 @@ import { Application } from '../../../api/api';
 export const ServiceLane: React.FC<{ application: Application }> = (props) => {
     const { application } = props;
     const releases = useDeployedReleases(application.name);
+    const releases_lane = releases.map((rel, index) => {
```
medo-freiheit, line 26:
```suggestion
    const releases_lane = !!releases && releases.map((rel, index) => {
```
just to make sure, Kuberpult will not crash if there is no releases

medo-freiheit (approved):

medo-freiheit (approved):

PR #412, opened 2022-12-08
Bugfix: Write undeploy versions correctly into manifest repo

sven-urbanski-freiheit-com:
go-billys WriteFile function has a funny behavior: It does not allow overwriting an existing file with empty content. We work around this by giving the file a single space. For argoCd this should not make a difference

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -372,7 +372,10 @@ func (c *CreateUndeployApplicationVersion) Transform(ctx context.Context, state 
 			return "", err
 		}
 		// note that the manifest is empty here!
-		if err := util.WriteFile(fs, fs.Join(envDir, "manifests.yaml"), []byte(""), 0666); err != nil {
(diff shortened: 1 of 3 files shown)


mouhsen-ibrahim (approved):

services/cd-service/pkg/repository/transformer.go
```diff
@@ -372,7 +372,10 @@ func (c *CreateUndeployApplicationVersion) Transform(ctx context.Context, state
 			return "", err
 		}
 		// note that the manifest is empty here!
-		if err := util.WriteFile(fs, fs.Join(envDir, "manifests.yaml"), []byte(""), 0666); err != nil {
+		// but actually it's not quite empty!
```
sven-urbanski-freiheit-com, line 375: test is missing!

jaypontes: Added the missing tests! 👍

mouhsen-ibrahim (approved):

PR #424, opened 2022-12-13
Implement release page

mouhsen-ibrahim:


services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -376,3 +376,24 @@ export const useReleasesForApp = (app: string) =>
             a.version === -1 ? -1 : b.version === -1 ? 1 : b.version - a.version
         )
     );
+
+// Get all releases for an app and group them by date
+export const useReleasesForAppGroupByDate = (app: string) =>
```
sven-urbanski-freiheit-com, line 381: please add a test for this function

mouhsen-ibrahim: This function doesn't work here, so I replaced it with a function in the component's code

sven-urbanski-freiheit-com (approved):

PR #427, opened 2022-12-15
Bugfix: Show lock message text & label properly

sven-urbanski-freiheit-com:


sven-urbanski-freiheit-com:
This is how it looks now:
2022-12-15_14-43_kuberpult_lock

Before, the message ("test lock") was invisible and the title "Lock Message" was aligned wrong

jaypontes (approved):

PR #462, opened 2023-01-23
Add Environment Groups to getOverview response

sven-urbanski-freiheit-com:
This adds a new field `EnvironmentGroups` to the server response.
The old field `Environments` still exists, but is deprecated. Logically, they contain the same information, but `EnvironmentGroups` is structured better for use in the UI.

This commits moves over functions like `calculateEnvironmentPriorities` to the server, that where implemented on client side before.

### DIFF
--- a/pkg/api/api.proto
+++ b/pkg/api/api.proto
@@ -172,6 +172,13 @@ message EnvironmentGroup {
   repeated Environment environments = 2;
 }
 
+enum Priority {
+  PROD = 0;
+  PRE_PROD = 1;
+  UPSTREAM = 2;
+  OTHER = 3;
+}
+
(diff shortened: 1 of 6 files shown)


pkg/api/api.proto
```diff
@@ -203,6 +216,8 @@ message Environment {
   Config config = 2;
   map<string, Lock> locks = 3;
   map<string, Application> applications = 4;
+  uint32 distanceToUpstream = 5;
```
medo-freiheit, line 219: If we are putting the `distanceToUpstream` in the `EnvironmentGroup`, do we need to duplicate it here ?

medo-freiheit (approved):

PR #466, opened 2023-01-24
SRX-ZP4P4F Create AppLock store funcs with shallow comparable results

jaypontes:
Some of these functions produce shallow comparable results (strings which represent the lock IDs) to address the Locks' page console errors.
While others are for translating a lock ID (string) to a DisplayLock so that, after the mapper is done using the shallow comparable results, we can translate those results (strings) into displayable information in the UI.


services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -300,6 +300,67 @@ export const searchCustomFilter = (queryContent: string | null, val: string | un
...
+    useOverview(({ environments }) =>
+        Object.values(environments)
+            .map((env) => Object.values(env.applications))
+            .flat()
+            .map((app) => Object.values(app.locks))
+            .flat()
+            .map((lock) => lock.lockId)
+    );
+
+export const useApplicationLock = (id: string) =>
```
sven-urbanski-freiheit-com, line 326:
```suggestion
export const useApplicationLock = (lockId: string) =>
```

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -300,6 +300,67 @@ export const searchCustomFilter = (queryContent: string | null, val: string | un
...
+        )!.name,
+        application: useOverview(({ environments }) =>
+            Object.values(environments)
+                .map((env) => Object.values(env.applications))
+                .flat()
+                .find((app) => Object.values(app.locks).find((lock) => lock.lockId === id))
+        )!.name,
+    } as DisplayLock);
+
+export const useLock = (id: string) =>
```
sven-urbanski-freiheit-com, line 361:
```suggestion
export const useLock = (lockId: string) =>
```

services/frontend-service/src/ui/Pages/Locks/LocksPage.test.tsx
```diff
@@ -40,120 +46,551 @@ describe('LocksPage', () => {
...
+            expectedLockIDs: ['ui-v2-321', 'ui-v2-1337'],
+        },
+    ];
+
+    describe.each(sampleAppFilteredData)(`Test Filtered Lock IDs`, (testcase) => {
+        it(testcase.name, () => {
+            // given
+            UpdateOverview.set({ environments: testcase.envs });
+            // when
+            const obtained = renderHook(() => useFilteredApplicationLockIDs(testcase.filter)).result.current;
```
sven-urbanski-freiheit-com, line 532: the goal of this PR is not clear to me.
Are all of these changes necessary to avoid the react error message?

services/frontend-service/src/ui/Pages/Locks/LocksPage.test.tsx
```diff
@@ -40,120 +46,551 @@ describe('LocksPage', () => {
...
+            // then
+            const sortedObtained = renderHook(() => sortAppLocksFromIDs(obtained, 'ascending')).result.current;
```
sven-urbanski-freiheit-com, line 534: Why do we implement a function `useFilteredApplicationLockIDs` that is only used in tests? Where is it use in the src?

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -350,6 +411,56 @@ export const sortLocks = (displayLocks: DisplayLock[], sorting: string) => {
     return displayLocks;
 };
 
+export const sortEnvLocksFromIDs = (lockIDs: string[], sorting: 'ascending' | 'descending') => {
```
leandro-freiheit, line 414:
```suggestion
export const sortEnvLocksFromIDs = (lockIDs: string[], sorting: 'oldestToNewest' | 'newestToOldest') => {
```

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -350,6 +411,56 @@ export const sortLocks = (displayLocks: DisplayLock[], sorting: string) => {
...
+            if (aValues[aValues.length - 1] === bValues[aValues.length - 1]) {
+                return 0;
+            }
+        }
+        return 0;
+    });
+    return displayLocks.map((lock) => lock.lockId);
+};
+
+export const sortAppLocksFromIDs = (lockIDs: string[], sorting: 'ascending' | 'descending') => {
```
leandro-freiheit, line 439:
```suggestion
export const sortAppLocksFromIDs = (lockIDs: string[], sorting: 'oldestToNewest' | 'newestToOldest') => {
```

services/frontend-service/src/ui/utils/store.tsx
leandro-freiheit: I think this naming makes it more explicit what the sorting is

services/frontend-service/src/ui/Pages/Locks/LocksPage.test.tsx
jaypontes: I'm still working on getting them to work in the src. I'm getting some hook specific errors

services/frontend-service/src/ui/Pages/Locks/LocksPage.test.tsx
jaypontes: The errors are due to the way we currently get the locks (as a `DisplayLock`) from the store. For the mapper to be able to produce shallow comparable results, the array that is mapped must be a primitive type or a very simple object (with just primitive types in the "1st level").
example: `{key: string; id: number}` works, `{key: string; date: Date}` doesn't work because `Date` is also an object

So my plan was to create functions in the store that would return just strings (lock IDs). Those strings would go into the map, the map would be able to produce shallow comparable results because of the simple strings. Then the strings would be used to get the full object from the store to be able to display it.

PR #473, opened 2023-01-26
SRX-8OM7M0 Install pnpm in builder

jaypontes:
Installed pnpm in builder image after go version upgrade.
Also installed npm for 2 reasons:
- Simpler command to install pnpm (could've used `wget` as stated in the pnpm docs)
- npm depends on node which is necessary to run the tests anyway

### DIFF
--- a/services/cd-service/Makefile
+++ b/services/cd-service/Makefile
@@ -22,7 +22,7 @@ MAKEFLAGS += --no-builtin-rules
 
 export CGO_ENABLED=1
 
-IMAGENAME?=$(IMAGE_REGISTRY)/kuberpult-cd-service:$(VERSION)
+IMAGENAME?=$(IMAGE_REGISTRY)/kuberpult/kuberpult-cd-service:$(VERSION)
 
 ifeq ($(WITH_DOCKER),)
 COMPILE_WITH_DOCKER := false
--- a/services/cd-service/pkg/service/overview.go
+++ b/services/cd-service/pkg/service/overview.go
@@ -21,13 +21,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"google.golang.org/protobuf/types/known/timestamppb"
 	"os"
 	"path/filepath"
 	"regexp"
 	"sort"
 	"sync"
 	"sync/atomic"
+
+	"google.golang.org/protobuf/types/known/timestamppb"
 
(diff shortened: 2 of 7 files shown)


leandro-freiheit (approved):

mouhsen-ibrahim (changes requested):
You need to use the new builder image and test building cd and frontend service using it

mouhsen-ibrahim (approved):
revert changes used for testing cd and frontend services then merge

services/cd-service/pkg/service/overview.go
```diff
@@ -477,6 +478,7 @@ func mapEnvironmentsToGroups(envs map[string]config.EnvironmentConfig) []*api.En
 		}
 	}
 	calculateEnvironmentPriorities(tmpEnvs)
+	sort.Sort(EnvironmentGroupsByDistance(result))
```
sven-urbanski-freiheit-com, line 481: if you sort here, remove line 469, it calls the same function.

I'm not sure what difference it would make, I think none. But I'm ok with doing the sorting here.

sven-urbanski-freiheit-com: What problem does it solve to do the sorting here? I'd like to investigate that a little bit. Sounds a bit like we have an unintended side effect here somewhere.

sven-urbanski-freiheit-com (approved):

services/cd-service/Makefile
```diff
@@ -22,7 +22,7 @@ MAKEFLAGS += --no-builtin-rules
 
 export CGO_ENABLED=1
 
-IMAGENAME?=$(IMAGE_REGISTRY)/kuberpult-cd-service:$(VERSION)
+IMAGENAME?=$(IMAGE_REGISTRY)/kuberpult/kuberpult-cd-service:$(VERSION)
```
sven-urbanski-freiheit-com, line 25: please change it also in the file `.github/workflows/execution-plan-snippet.yml`

services/cd-service/pkg/service/overview.go
sven-urbanski-freiheit-com: please undo this change about sorting.

jaypontes: Done!

sven-urbanski-freiheit-com (approved):

mouhsen-ibrahim (approved):

PR #480, opened 2023-02-03
Make v2 publicly available (beta)

sven-urbanski-freiheit-com:
Enables the url path `/v2/home` for everyone.

Note that this url is usable, but still missing some features. In general, we recommend to keep using the old UI for now, under url path `/`.

### DIFF
--- a/services/cd-service/pkg/cmd/server.go
+++ b/services/cd-service/pkg/cmd/server.go
@@ -54,6 +54,7 @@ type Config struct {
 	DogstatsdAddr     string `default:"127.0.0.1:8125" split_words:"true"`
 }
 
+// DELETE ME
 func (c *Config) readPgpKeyRing() (openpgp.KeyRing, error) {
 	if c.PgpKeyRing == "" {
 		return nil, nil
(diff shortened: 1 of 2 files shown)


services/frontend-service/pkg/cmd/server.go
```diff
@@ -228,12 +226,7 @@ func RunServer() {
 						return
 					}
 				}
-				if strings.HasPrefix(req.URL.Path, "/v2/home") {
-					http.ServeFile(resp, req, "build/index.html")
-				} else {
-					mux.ServeHTTP(resp, req)
-				}
-
+				mux.ServeHTTP(resp, req)
```
mouhsen-ibrahim, line 229: it doesn't work like this, the http server in go is configured to convert URLs to paths in order to return responses, there is nothing in the path `v2/home` on disk so it will return 404, the current config routes `/v2/home` to `build/index.html` which is generated by react and it handles routing

mouhsen-ibrahim (commented):
This will disable access to new UI not enable it

services/cd-service/pkg/cmd/server.go
```diff
@@ -54,6 +54,7 @@ type Config struct {
 	DogstatsdAddr     string `default:"127.0.0.1:8125" split_words:"true"`
 }
 
+// DELETE ME
```
mouhsen-ibrahim, line 57: please add a TODO here so we can easily spot this later

mouhsen-ibrahim (approved):

PR #484, opened 2023-02-07
SRX-NYXX8S Run kp in github actions

jaypontes:


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,26 @@
+# This runs the integration tests for kuberpult
+name: "Kuberpult Integration tests"
+'on':
+  pull_request:
+    branches:
```
mouhsen-ibrahim, line 5: this means run it only on a branch called `pr` remove it

jaypontes: Thank you!!! :)

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,29 @@
...
+    - name: Create bare repository
+      run: |
+        mkdir -p ./services/cd-service/repository_remote
+        git -C ./services/cd-service/repository_remote init --bare
+        git clone ./services/cd-service/repository_remote ./services/cd-service/repository_checkedout
+    - name: Run docker compose
+      run: |
+        IMAGE_REGISTRY=europe-west3-docker.pkg.dev/fdc-public-docker-registry/kuberpult \
+        IMAGE_TAG=$(make version) \
+        docker compose -f ./docker-compose.tpl.yml up -d
```
sven-urbanski-freiheit-com, line 27: where are the template values replaced by real values? Does this just happen by having the env vars set?

docker-compose.tpl.yml
```diff
@@ -0,0 +1,22 @@
+services:
+  backend:
```
sven-urbanski-freiheit-com, line 2: can we call this one `cd`?
I find it confusing that we come up with a different name here

.github/workflows/execution-plan-main.yml
```diff
@@ -13,6 +13,12 @@ jobs:
       trigger: main
     secrets: inherit
 
+  integration-tests:
```
sven-urbanski-freiheit-com, line 16: for testing purposes, add this to execution-plan-pr.yaml
I

sven-urbanski-freiheit-com, line 16: add this to execution-plan-pr.yaml, otherwise it won't run until you merge this PR

sven-urbanski-freiheit-com (approved):
approved with comments

.github/workflows/integration-tests.yml
jaypontes: Yes! 😄 
I tested it locally and docker compose can use env vars implicitly

.github/workflows/execution-plan-main.yml
jaypontes: I did this in the previous commit, I tested with an already pushed tag.
I moved it from the PR workflow for these reasons:
- Having an integration test every time you push to a PR sounds really nice but it takes quite a while to run this workflow. I think this should be something discussed as a team.
- But more importantly: the image is not built & pushed in the PR workflow, only when merging with main. So it's not possible to use `make version` yet... and building the image every time you push to a PR would make the 1st point even worse.

jaypontes: I addressed this in the comment above

sven-urbanski-freiheit-com: @jaypontes ok fair.
Could you link a successful run here?

I would still keep it in *for now*, and remove it again before you merge it. You need some way to show that it works before merging.

Since this PR is still a draft, I assume you want to make further changes still.

.github/workflows/integration-tests.yml
sven-urbanski-freiheit-com: ok, nice :D I did not know

.github/workflows/execution-plan-main.yml
jaypontes: Here it is: https://github.com/freiheit-com/kuberpult/actions/runs/4136581181/jobs/7150679117 😁

docker-compose.tpl.yml
jaypontes: Done!

mouhsen-ibrahim (approved):

PR #490, opened 2023-02-15
Add documentation for Readme & queues, etc

sven-urbanski-freiheit-com:


pkg/api/api.proto
```diff
@@ -206,6 +206,8 @@ message Environment {
     string name = 1;
     uint64 version = 2;
     map<string, Lock> locks = 3;
+    // "version" describe the currently deployed version. "queuedVersion" describes a version that was *tried* to be deployed, but a lock stopped it:
```
leandro-freiheit, line 209:
```suggestion
    // "version" describes the currently deployed version. "queuedVersion" describes a version that was to be deployed, but a lock stopped the deployment:
```

leandro-freiheit: I understand the focus on "try" but the phrasing felt a bit confusing

PR #492, opened 2023-02-16
Refactoring: Require return types in new ui

sven-urbanski-freiheit-com:
Added rule '@typescript-eslint/explicit-function-return-type' which requires all functions to have an explicit return type.
    
Also added a specific .eslintrc for the old ui, which disables this rule, because we do not want to spend much time with the old ui anymore.
    
Adapted all functions that didn't have a return type.
    
This is a refactoring, it does not change any behavior.

SRX-2DXE9T

### DIFF
--- a/services/frontend-service/src/ui/utils/store.tsx
+++ b/services/frontend-service/src/ui/utils/store.tsx
@@ -307,14 +307,14 @@ export const useEnvironmentLock = (lockId: string): DisplayLock =>
         )?.name,
     } as DisplayLock);
 
-export const searchCustomFilter = (queryContent: string | null, val: string | undefined): string | undefined | null => {
+export const searchCustomFilter = (queryContent: string | null, val: string | undefined): string => {
     if (!!val && !!queryContent) {
         if (val.includes(queryContent)) {
             return val;
         }
-        return null;
+        return '';
     } else {
-        return val;
+        return val || '';
     }
 };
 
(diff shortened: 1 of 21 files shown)


services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -306,7 +307,7 @@ export const useEnvironmentLock = (lockId: string) =>
         )?.name,
     } as DisplayLock);
 
-export const searchCustomFilter = (queryContent: string | null, val: string | undefined) => {
+export const searchCustomFilter = (queryContent: string | null, val: string | undefined): string | undefined | null => {
```
tameremad, line 310: Can we handle the `undefined | null` cases here and return perhaps `''`?

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -363,10 +364,10 @@ export const useApplicationLock = (lockId: string) =>
         )?.name,
     } as DisplayLock);
 
-export const useLock = (lockId: string) =>
+export const useLock = (lockId: string): DisplayLock | undefined =>
```
tameremad, line 367: Also the same comment as before? 

Additionally, to me, this function has no usages anywhere. do we need to keep it?

tameremad (approved):
Sounds good!

services/frontend-service/src/ui/utils/store.tsx
sven-urbanski-freiheit-com: done

services/frontend-service/src/ui/utils/store.tsx
sven-urbanski-freiheit-com: done, removed it.

PR #496, opened 2023-02-17
Update homepage design. 

tameremad:
Starting with the new homepage design:
* Adding a tooltip component from mdc.
* styling it with corners and borders and etc...
* change the environment colors
* display short environment tags on release cards
* fix bug with release cards shadows
* make release cards smaller and shift around where the details are displayed

SRX-OD9V6S


services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.tsx
```diff
@@ -43,27 +47,50 @@ export const ReleaseCard: React.FC<ReleaseCardProps> = (props) => {
...
+            content={
+                <>
+                    <h2 className="mdc-tooltip__title release__details">
+                        {!!sourceMessage && <b>{sourceMessage}</b>}
+                        {!!sourceCommitId && <Button className="release__hash" label={sourceCommitId} />}
+                        {!!sourceAuthor && <div>{'| ' + sourceAuthor + ' |'}</div>}
+                        {!!createdAt && (
+                            <div className="release__metadata mdc-typography--subtitle2">
+                                <div>
+                                    {`${createdAt.getDay()}-${createdAt.getMonth()}-${createdAt.getFullYear()}` +
```
sven-urbanski-freiheit-com, line 61: nit: could you extract this into a variable or function?
e.g.
const formattedDate = ...
const formattedTime = ...

services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.tsx
```diff
@@ -43,27 +47,50 @@ export const ReleaseCard: React.FC<ReleaseCardProps> = (props) => {
...
+                </div>
+                <div className={classNames('mdc-card release-card', className)}>
+                    <div
+                        className="mdc-card__primary-action release-card__description"
+                        ref={control}
+                        tabIndex={0}
+                        onClick={clickHandler}>
+                        <div className="release-card__header">
+                            <div className="release__title mdc-typography--headline6">
+                                {sourceMessage === '' ? 'Undeploy Version' : sourceMessage}
```
sven-urbanski-freiheit-com, line 86: There's a field "undeployVersion", please use that instead of checking the sourceMessage.

services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.tsx
```diff
@@ -43,27 +47,50 @@ export const ReleaseCard: React.FC<ReleaseCardProps> = (props) => {
...
+                        {!!sourceAuthor && <div>{'| ' + sourceAuthor + ' |'}</div>}
+                        {!!createdAt && (
+                            <div className="release__metadata mdc-typography--subtitle2">
+                                <div>
+                                    {`${createdAt.getDay()}-${createdAt.getMonth()}-${createdAt.getFullYear()}` +
+                                        ' @ ' +
+                                        `${createdAt.getHours()}:${createdAt.getMinutes()}` +
+                                        ' | '}
+                                    <i>
+                                        {daysToString(((Date.now().valueOf() - createdAt.valueOf()) / MsPerDay) >> 0)}
```
sven-urbanski-freiheit-com, line 66: Nit: Similar here:
can we put that into a variable or function?

sven-urbanski-freiheit-com: my goal here is to give it a name, then it's easier to understand.
I actually don't understand what `/ MsPerDay >> 0` does.
Can we please avoid the shift operator?

services/frontend-service/src/ui/components/tooltip/index.ts
```diff
@@ -0,0 +1,18 @@
...
+Kuberpult is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+MIT License for more details.
+
+You should have received a copy of the MIT License
+along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>.
+
+Copyright 2023 freiheit.com*/
+import { Tooltip } from './tooltip';
```
sven-urbanski-freiheit-com, line 16: Do we really need this index.ts? Why can't we just use `tooltip.tsx` (and scss)?

sven-urbanski-freiheit-com (approved):
approved with comments

services/frontend-service/src/ui/components/tooltip/index.ts
tameremad: index.ts has been removed :+1:

services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.tsx
tameremad: done!

services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.tsx
tameremad: Also done! 
I forgot about the floor function and couldn't think of it :grimacing:

tameremad:
There was a bug in ReleaseDialo. The time gets displayed in UTC timezone at all times. 
e.g. I create a lock at 12:15
when it shows up on the ui it just says 11:15. 
The actual stored date is correct but it's displayed incorrectly. 
https://github.com/freiheit-com/kuberpult/blob/beed1f7e0e5ab892d0ed2edbb1fe9d5fa547dd42/services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.tsx#L219-L225

After fixing this issue. the release dialog snapshot started failing. I updated the snapshot to match my local machine output. but when it runs on github, the timezone is different and it fails. 
I updated the test to mock the time display function as it's tested in Release Card

services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.tsx
```diff
@@ -15,11 +15,52 @@ along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>
...
+            return '1 day ago';
+        } else {
+            return `${daysSinceDate} days ago`;
+        }
+    }
+};
+
+export const getFormattedReleaseDate = (createdAt: Date): JSX.Element => {
+    // date format: dd-mm-yyyy, with no leading zeros, month is 0-indexed.
+    const formattedDate = `${createdAt.getDate()}-${createdAt.getMonth() + 1}-${createdAt.getFullYear()}`;
```
sven-urbanski-freiheit-com, line 51: I know this is like in the design, but I think we should change it.
I would much prefer an ISO-like date:
2023-12-31

sven-urbanski-freiheit-com (approved):

PR #503, opened 2023-02-23
Show more older releases on home

tameremad:
* As in the figma design, show a trailing number of releases after the currently deployed ones.
* add a margin for whenever environment tags are not displayed.
* add Diff Elements and styles
* add the small lock icons on environment chips

### DIFF
--- a/services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.scss
+++ b/services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.scss
@@ -64,14 +64,8 @@
   @extend .text-bold;
   align-self: center;
   z-index: 2;
-
+  min-height: 42px;
   > *:not(:last-child) {
     margin-right: 8px;
-  }
-
-  &--empty {
-    // when no environment chips are displayed, take the same space with an empty div
-    // used to make release cards on the same level
-    height: 42px;
   }
 }
--- a/services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.scss
+++ b/services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.scss
@@ -11,7 +11,7 @@
   max-height: 136px;
   min-width: 240px;
   max-width: 240px;
-  margin: -20px 16px 4px 16px;
+  margin: -20px 8px 4px 8px;
   color: var(--mdc-theme-on-surface);
   box-shadow: none;
 
--- a/services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.scss
+++ b/services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.scss
@@ -68,4 +68,10 @@
   > *:not(:last-child) {
     margin-right: 8px;
   }
+
+  &--empty {
+    // when no environment chips are displayed, take the same space with an empty div
+    // used to make release cards on the same level
+    height: 42px;
+  }
 }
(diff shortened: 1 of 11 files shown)


medo-freiheit (approved):

services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.scss
```diff
@@ -68,4 +68,10 @@
   > *:not(:last-child) {
     margin-right: 8px;
   }
+
+  &--empty {
+    // when no environment chips are displayed, take the same space with an empty div
+    // used to make release cards on the same level
+    height: 42px;
```
sven-urbanski-freiheit-com, line 75: Maybe `min-height` would make more sense for this use-case?

services/frontend-service/src/ui/components/ServiceLane/ServiceLane.tsx
```diff
@@ -22,6 +22,25 @@ import { Tooltip } from '@material-ui/core';
...
 import * as React from 'react';
 
+// number of releases on home. based on design
+const numberOfDisplayedReleasesOnHome = 6;
+
+const getAllDisplayedReleases = (deployedReleases: number[], allReleases: number[]): number[] => {
+    // number of remaining releases to get from history
+    const numOfTrailingReleases = numberOfDisplayedReleasesOnHome - deployedReleases.length;
+    // last deployed release e.g. Prod
+    const oldestDeployedRelease = deployedReleases[deployedReleases.length - 1];
```
sven-urbanski-freiheit-com, line 32: this will crash if `deployedReleases` is empty, which isn't likely, but can happen.

services/frontend-service/src/ui/components/ServiceLane/ServiceLane.tsx
```diff
@@ -22,6 +22,25 @@ import { Tooltip } from '@material-ui/core';
...
+
+    const allDisplayedReleases = deployedReleases;
+    // go over the remaining spots and fill from history
+    for (let i = 1; i <= numOfTrailingReleases; i++) {
+        const index = allReleases.indexOf(oldestDeployedRelease) + i;
+        if (index >= allReleases.length) break;
+        allDisplayedReleases.push(allReleases[index]);
+    }
+    return allDisplayedReleases;
+};
```
sven-urbanski-freiheit-com, line 42: I can't help but think this is more complicated than it needs to be.
Can't we just do
`(deployedRelease + allRelease).slice(0, numberOfDisplayedReleasesOnHome)`
?

sven-urbanski-freiheit-com (changes requested):
.

services/frontend-service/src/ui/components/ServiceLane/ServiceLane.tsx
tameremad: The logic has to stay the same because:
* `deployedReleases` is definitely contained in `allReleases`
* adding them up would lead to repetition.
* We need to throw away all versions newer than the oldest deployed one.
  e.g. we don't display ver. 6 before ver. 4 which is deployed to production. 
  the important/ deployed releases should go first and then we fill from history.
 
 -----
 
 I did improve it and use slice instead of a for loop. :+1:

services/frontend-service/src/ui/components/ServiceLane/ServiceLane.tsx
tameremad: good catch! thanks :)

sven-urbanski-freiheit-com (approved):

PR #505, opened 2023-02-24
Automatically open cart when actions are added

ahmed-nour-fdc:
When adding anything to the cart: Open the cart
show number of actions in the top: "planned actions (42)"

### DIFF
--- a/services/frontend-service/src/ui/components/SideBar/SideBar.test.tsx
+++ b/services/frontend-service/src/ui/components/SideBar/SideBar.test.tsx
@@ -25,7 +25,7 @@ import {
     deleteAllActions,
     UpdateOverview,
 } from '../../utils/store';
-import { ActionDetails, ActionTypes, getActionDetails } from './SideBar';
+import { ActionDetails, ActionTypes, getActionDetails, SideBar, clearActionsNumber } from './SideBar';
 
 describe('Show and Hide Sidebar', () => {
     interface dataT {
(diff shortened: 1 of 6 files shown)


services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -54,7 +54,25 @@ export type ActionDetails = {
     lockMessage?: string;
     version?: number;
 };
-
+let totalActions = 0;
```
tameremad, line 57: you don't have to keep count of the number of actions.

you can get that from the actions store. 

https://github.com/freiheit-com/kuberpult/blob/e419e420d24400bcc385711a2814ff692555e46d/services/frontend-service/src/ui/utils/store.tsx#L44

tameremad: and in the `actionsCart` you can use that hook.
https://github.com/freiheit-com/kuberpult/blob/e419e420d24400bcc385711a2814ff692555e46d/services/frontend-service/src/legacy-ui/ActionsCart/ActionsCart.tsx#L91-L93

```
// in Actions Cart
const numberOfActions = useNumberOfActions();
```

```
// in the store.ts
useNumberOfActions = useActions().length;
```

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -26,6 +26,7 @@ import {
 import { useApi } from './GrpcApi';
 import { useMemo } from 'react';
 import { Empty } from '../../google/protobuf/empty';
+import { updateActionsNumber } from '../components/SideBar/SideBar';
```
tameremad, line 29: therefore you also don't need the update callback here.

tameremad (changes requested):

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -54,7 +54,25 @@ export type ActionDetails = {
...
+export const clearActionsNumber = (): void => {
+    totalActions = 0;
+    const elem = document.getElementsByClassName('mdc-drawer-sidebar-header-title');
+    if (elem[0]) {
+        elem[0].innerHTML = 'Planned Actions';
+    }
+};
+export const updateActionsNumber = (added: boolean): void => {
+    added ? (totalActions += 1) : (totalActions -= 1);
+    const elem = document.getElementsByClassName('mdc-drawer-sidebar-header-title');
```
tameremad, line 67: Also, just a tip. It's not recommended to use this way to update some components or text in the app. 

they are already somewhere in the code, and you should do the updating there.

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
ahmed-nour-fdc: The actionsCart is in the legacy ui and is not present in the new ui. Perhaps you meant something else?

ahmed-nour-fdc: I updated the method of keeping track of the actions in cart as you mentioned

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
ahmed-nour-fdc: I changed it to be directly edited where it is defined.

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
tameremad: You're correct, I meant the `sidebar.tsx` file.

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -236,6 +244,7 @@ export const SideBar: React.FC<{ className: string; toggleSidebar: () => void }>
                 deleteAllActions();
             });
         handleClose();
+        updateActionsNumber(0);
```
tameremad, line 247: the `updateActionsNumber` is a little overcomplicated. what it actually does is update the sidebar _title_. 

the function name should indicate what the function does or its purpose. e.g. `updateSidebarTitle`

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
ahmed-nour-fdc: Done

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -51,11 +51,21 @@ export const useApplyActions = (): Promise<Empty> => useApi.batchService().Proce
...
+export const useSidebarShown = (): boolean => useSidebar(({ shown }) => shown);
+
+export const useNumberOfActions = (): number => useAction(({ actions }) => actions.length);
+
 export const updateActions = (actions: BatchAction[]): void => {
     deleteAllActions();
     actions.forEach((action) => addAction(action));
 };
 
+export const appendAction = (actions: BatchAction[]): void => {
```
tameremad, line 65: What does this function do?

tameremad (approved):
:+1:

services/frontend-service/src/ui/utils/store.tsx
ahmed-nour-fdc: It appends to the actions list, unlike updateActions which deletes all actions and adds another actions.
It is used in the unit testing of the sidebar.

tameremad: Is there any difference between `updateActions` and `appendActions` where your code is concerned? 
in other words, what are you trying to test? 

I think if you mock everything it would make your unit tests simpler.

services/frontend-service/src/ui/components/SideBar/SideBar.test.tsx
```diff
@@ -565,4 +566,149 @@ describe('Action details', () => {
...
+            };
+            return (
+                <MemoryRouter>
+                    <SideBar {...defaultProps} {...overrides} />
+                </MemoryRouter>
+            );
+        };
+        const getWrapper = (overrides?: {}) => render(getNode(overrides));
+        it('Create an action initially', () => {
+            updateActions([{ action: { $case: 'undeploy', undeploy: { application: 'test' } } }]);
```
tameremad, line 681: for example, no need to create all the complicated data for the deploy and such. 

use any only in tests will make testing much easier. 

the `useNumberOfActions` hook only checks the length of the array so it doesn't matter what you put inside.

```suggestion
            updateActions(['foo', 'bar', 'baz'] as any); 
```

services/frontend-service/src/ui/utils/store.tsx
ahmed-nour-fdc: I am testing the functionality of showing the number of actions in the cart. I first add 2 actions then make sure that the reading is 2, then add another action and check that it is 3, and so on. This isn't achievable with updateActions as it clears all actions first which would always reset the number of actions to before adding them.

services/frontend-service/src/ui/components/SideBar/SideBar.test.tsx
ahmed-nour-fdc: Alright. I will take this into consideration in the future.

services/frontend-service/src/ui/utils/store.tsx
tameremad: adding an action basically just changes the array. 
the function of automatic refresh and (update the number from 2 to 3) is already included in the store library. there's no need to test this. 

what you need to test is what your code does. 
e.g. if there's no actions title is 'Planned Actions'
if there's two then it's 'Planned Actions (2)'
etc. 

the hook simply uses the array.length which is just part of the language. also not needed to test. 

check [this](https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/WORKSPACE_DOCUMENT/-NIHiZUxqRbTHNflVb69/) out I hope it helps.

PR #509, opened 2023-02-28
Allow deleting locks from the locks page

ahmed-nour-fdc:
On the locks page (/v2/locks) it is possible to remove application locks and environment locks.

### DIFF
--- a/services/frontend-service/src/ui/components/LockDisplay/LockDisplay.test.tsx
+++ b/services/frontend-service/src/ui/components/LockDisplay/LockDisplay.test.tsx
@@ -13,7 +13,12 @@ You should have received a copy of the MIT License
 along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>.
 
 Copyright 2023 freiheit.com*/
-import { calcLockAge, daysToString, isOutdated } from '../LockDisplay/LockDisplay';
+import { act, render } from '@testing-library/react';
+import { BatchAction } from '../../../api/api';
+import { useApi } from '../../utils/GrpcApi';
+import { updateActions } from '../../utils/store';
+import { calcLockAge, daysToString, isOutdated, LockDisplay } from '../LockDisplay/LockDisplay';
+import { SideBar } from '../SideBar/SideBar';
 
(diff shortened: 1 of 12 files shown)


services/frontend-service/src/ui/components/LockDisplay/LockDisplay.test.tsx
```diff
@@ -117,3 +122,76 @@ describe('Test Auxiliary Functions for Lock Display', () => {
...
+            updateActions(testcase.actions);
+            api.batchService().ProcessBatch(testcase.actions);
+            // when
+            // const { container } = getWrapper({});
+            render(<LockDisplay lock={lock} />);
+            const result = document.querySelector('.service-action--delete')! as HTMLElement;
+            act(() => {
+                result.click();
+            });
+            render(<SideBar />);
```
tameremad, line 190: please remove the SideBar from the Lock display test. 
unit tests are small and only test specific functions. if the sidebar changes, this test should not get affected. 

use mocks to mock any inputs and outputs of this component. 
(i.e. mock the addiction function and expect it to be called with the delete lock action)

services/frontend-service/src/ui/components/LockDisplay/LockDisplay.test.tsx
```diff
@@ -117,3 +122,76 @@ describe('Test Auxiliary Functions for Lock Display', () => {
...
+    describe.each(data)('', (testcase) => {
+        it('Cart initially empty', () => {
+            render(<LockDisplay lock={lock} />);
+            expect(document.getElementsByClassName('mdc-drawer-sidebar-list').length).toBe(0);
+        });
+        it(testcase.name, () => {
+            const api = useApi;
+            // given
+            updateActions(testcase.actions);
+            api.batchService().ProcessBatch(testcase.actions);
```
tameremad, line 182: please don't use the api here. you can use the overview store and fill it with locks.
But that's not even required. the LockDisplay does not read anything directly from the store.

tameremad (changes requested):
only some comments on the unit tests

tameremad (approved):
:+1:

PR #510, opened 2023-03-01
Whitelist create environment endpoint 

medo-freiheit:


pkg/auth/azure_test.go
```diff
@@ -247,35 +247,97 @@ func TestHttpMiddleware(t *testing.T) {
...
+			ExpectedError: "Failed to parse the JWT.\nError: token contains an invalid number of segments",
+			Authenticated: false,
+		},
+		{
+			Name:          "api call create environment - no env",
+			Path:          "/environments/",
+			Method:        http.MethodPost,
+			ExpectedError: "Failed to parse the JWT.\nError: token contains an invalid number of segments",
+			Authenticated: false,
 		},
```
jaypontes, line 341: Should there be a test with POST that is successful? Or is this not possible?

jaypontes (approved):

leandro-freiheit:
I am not sure how the xpath function works, so I am not sure if I can properly review this.

pkg/auth/azure.go
```diff
@@ -161,17 +161,26 @@ func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfu
...
+	// Skip azure authentication with ID for `/` (POST: createEnv), `/release`, `/releasetrain` and `/locks`  endpoints. The requests will be validated with pgp signature
 	// usage in requests from outside the cluster (e.g. by GitHub Actions and the publish.sh script).
-	releaseTrainRx := regexp.MustCompile("^/environments/[^/]+/releasetrain$")
-	if releaseTrainRx.MatchString(req.URL.Path) {
-		return nil
-	}
-	environmentLocksRx := regexp.MustCompile("^/environments/[^/]+/locks/[^/]+$")
-	if environmentLocksRx.MatchString(req.URL.Path) {
-		return nil
+	group, tail := xpath.Shift(req.URL.Path)
```
leandro-freiheit, line 167: As I understand, this is shifting something in the request, but I don't understand what. Can you explain a bit @medo-freiheit?

jaypontes: From what I understand the shift simply splits the leftmost part of the URL
So if you have `/environments/prod/blabla`, the shift will `environments` part in the `group` variable and the `/prod/blabla` into the `tail` variable

leandro-freiheit: Alright, I am satisfied with the answer! Looks good to me

PR #544, opened 2023-03-09
Fix rendering of group labels on ReleaseCards

sven-urbanski-freiheit-com:
Before they were display vertically, now they are displayed horizontally.

Old:
2023-03-09_18-02_old

New:
2023-03-09_18-02_new


tameremad (approved):
That's interesting! :thinking: it doesn't usually happen. 
Could it be because of having many environments? which somehow think they need more space than they actually need?

PR #546, opened 2023-03-09
Create more readable test data

sven-urbanski-freiheit-com:
Randomizes the authors and commit messages when creating a new release.


sven-urbanski-freiheit-com:
Note that it's actually quite random. So it, can happen that the same commit message is selected twice in a row.

PR #550, opened 2023-03-10
Warn on startup if an upstream does not exist

sven-urbanski-freiheit-com:
ideally we should also warn in CreateEnvironment, but that requires a more involved approach, so this commit is only checking it on startup


jaypontes (commented):
I don't really know much about this part of the code so I don't feel like the right person to review this :(

tameremad (approved):

PR #552, opened 2023-03-16
Retry connection on errors with exponential backoff

tameremad:
- Update the browserslist
- add rxjs retryWhen to retry subscription to the streamOverview Observable.
- use a randomized exponential backoff strategy.

SRX-0ELJF9


services/frontend-service/src/ui/App/index.tsx
```diff
@@ -23,6 +23,26 @@ import { PanicOverview, showSnackbarWarn, UpdateOverview, useReleaseDialog, useR
 import { useApi } from '../utils/GrpcApi';
 import { AzureAuthProvider, UpdateFrontendConfig, useAzureAuthSub } from '../utils/AzureAuthProvider';
 import { Snackbar } from '../components/snackbar/snackbar';
+import { mergeMap, retryWhen } from 'rxjs/operators';
+import { Observable, throwError, timer } from 'rxjs';
+
+// retry strategy: retries the observable subscription with randomized exponential backoff
+// source: https://www.learnrxjs.io/learn-rxjs/operators/error_handling/retrywhen#examples
+function retryStrategy(maxRetryAttempts = 3) {
```
sven-urbanski-freiheit-com, line 31: 3 seems low to me, I would suggest 6 or more

services/frontend-service/src/ui/App/index.tsx
```diff
@@ -23,6 +23,26 @@ import { PanicOverview, showSnackbarWarn, UpdateOverview, useReleaseDialog, useR
 import { useApi } from '../utils/GrpcApi';
 import { AzureAuthProvider, UpdateFrontendConfig, useAzureAuthSub } from '../utils/AzureAuthProvider';
 import { Snackbar } from '../components/snackbar/snackbar';
+import { mergeMap, retryWhen } from 'rxjs/operators';
+import { Observable, throwError, timer } from 'rxjs';
+
+// retry strategy: retries the observable subscription with randomized exponential backoff
+// source: https://www.learnrxjs.io/learn-rxjs/operators/error_handling/retrywhen#examples
```
sven-urbanski-freiheit-com, line 30: I would prefer to use an explicit value, instead of a default value

services/frontend-service/src/ui/App/index.tsx
```diff
@@ -23,6 +23,26 @@ import { PanicOverview, showSnackbarWarn, UpdateOverview, useReleaseDialog, useR
...
+function retryStrategy(maxRetryAttempts = 3) {
+    return (attempts: Observable<any>): Observable<any> =>
+        attempts.pipe(
+            mergeMap((error, i) => {
+                // retry attempt number
+                const retryAttempt = i;
```
hannesg, line 36: You are never incrementing this here. Is that correct?

services/frontend-service/src/ui/App/index.tsx
hannesg: I also think so. With just three retries and the backoff formula you used, you would stop retrying after 17 seconds. I think you can safely go to much higher retry counts.

services/frontend-service/src/ui/App/index.tsx
hannesg, line 1: Where are the tests?

services/frontend-service/src/ui/App/index.tsx
sven-urbanski-freiheit-com: yes, I think we can go much higher. Since it's exponential, it shouldn't be a problem

tameremad: Okay, I increased the attempts to 8 which takes about 2 mins to give up.

services/frontend-service/src/ui/App/index.tsx
tameremad: this doesn't need incrementing. (I had to figure out counting from 0 / 1) I just decided eventually to just remove the i

services/frontend-service/src/ui/App/index.tsx
tameremad: Added tests :+1:

services/frontend-service/src/ui/App/index.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+    },
+}));
+
+const getNode = (): JSX.Element => <App />;
+const getWrapper = () => render(getNode());
+
+describe('App uses the API', () => {
+    beforeAll(() => {
+        jest.useFakeTimers('modern');
+        jest.setSystemTime(new Date('2/1/22'));
```
sven-urbanski-freiheit-com, line 53: Is there a way to initialize the date with a more readible value? like `2022-01-02`?

sven-urbanski-freiheit-com (approved):

PR #553, opened 2023-03-16
refactoring: Use environmentGroup in servicelane

sven-urbanski-freiheit-com:
field "environments" is deprecated.

### DIFF
--- a/services/frontend-service/src/ui/components/ServiceLane/ServiceLane.test.tsx
+++ b/services/frontend-service/src/ui/components/ServiceLane/ServiceLane.test.tsx
@@ -17,7 +17,7 @@ import { render } from '@testing-library/react';
 import { ServiceLane } from './ServiceLane';
 import { UpdateOverview } from '../../utils/store';
 import { Spy } from 'spy4js';
-import { Application, Release } from '../../../api/api';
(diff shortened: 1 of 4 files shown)


services/frontend-service/src/ui/components/ServiceLane/ServiceLane.test.tsx
```diff
@@ -113,24 +120,38 @@ const data = [
                 createdAt: new Date(2002),
             },
         ] as Array<Release>,
```
jaypontes, line 122: Is it worth to keep `as Array<Release>` in?
iirc, having it declared in line `106` should be enough.
Plus I think using `as` adds complexity and is sometimes annoying to debug. Maybe we should think about this more deeply because I think using the explicit `type` (like in line `103 to 108` a bit better.

jaypontes (approved):

services/frontend-service/src/ui/components/ServiceLane/ServiceLane.test.tsx
sven-urbanski-freiheit-com: Yes, you're right, I fixed.
The nice things about "as ..." in tests, is that we don't need to specify so much data. But I think type safety is more important than that.

PR #556, opened 2023-03-17
Fix version parameter of release endpoint

sven-urbanski-freiheit-com:
This fixes an issue when the "release" endpoint gets a "version" parameter that is less than the newest version number.
In that case, kuberpult didn't generate any manifest files (just the basic files like "author" were written).


hannesg:
Looks good! Can you add a unit test.

hannesg (approved):

sven-urbanski-freiheit-com:
@hannesg I added 2 tests, and ensured one of them failed before the src change

PR #579, opened 2023-03-27
disable apply button, fix button ripples bug

tameremad:
* make the deploy button disabled when a version is already deployed.
* use the `button--unelevated` class so that contained/filled buttons would have ripples too.
* also use this class for the sidebar and the servicelane buttons
* fix a bug with the add lock button with white space wrapping
* ~~fix bug with store returning empty release for undeployed versions~~
before:
image
after:
image
fixes the ripple.
[Screencast from 29.03.2023 16:49:34.webm](https://user-images.githubusercontent.com/88655074/228578153-45b7e7ba-0c57-44f3-9c26-2bcbe7117178.webm)

SRX-HLVCYS


services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -487,7 +487,9 @@ export const useCurrentlyDeployedAtGroup = (application: string, version: number
 // Get release information for a version
 export const useReleaseInfo = (app: string, version: number): Release =>
     useOverview(({ applications }) => {
-        const releaseInfo = applications[app]?.releases.filter((release) => release.version === version)[0];
+        const releaseInfo = applications[app]?.releases.filter((r) =>
```
sven-urbanski-freiheit-com, line 490: if you just want the first result, use `find` instead of `filter`.

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -487,7 +487,9 @@ export const useCurrentlyDeployedAtGroup = (application: string, version: number
...
+            version === -1 ? r.undeployVersion : r.version === version
```
sven-urbanski-freiheit-com, line 491: do we really need to handle "-1" here?
I'd like to get red of all "-1".

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -487,7 +487,9 @@ export const useCurrentlyDeployedAtGroup = (application: string, version: number
...
+        )[0];
         if (!releaseInfo) {
             return {} as Release;
```
sven-urbanski-freiheit-com, line 494: please don't use `as`.
If there's nothing found we should return undefined.

tameremad: :+1:

sven-urbanski-freiheit-com: that is why I added functions like `useReleaseOptional`

tameremad: I will leave that for now. and will comeback to it later when there is more time.

services/frontend-service/src/ui/utils/store.tsx
tameremad: done :+1:

services/frontend-service/src/ui/utils/store.tsx
tameremad: I will leave this as it is too.

sven-urbanski-freiheit-com:
this is hard to review, please add before/after screenshots

sven-urbanski-freiheit-com (approved):
please clean up the commit message before merging

PR #580, opened 2023-03-27
new UI: add prepareToUndeploy and Undeploy functions

sven-urbanski-freiheit-com:


services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.tsx
```diff
@@ -122,6 +130,14 @@ export const EnvironmentListItem: React.FC<{
             </div>
         );
     const otherRelease = useReleaseOptional(app, env);
+    const application = env.applications[app];
+    const commitString = application
+        ? release.undeployVersion
```
jaypontes, line 135: Nitpick: This is kind of hard to read

jaypontes (approved):

services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.tsx
sven-urbanski-freiheit-com: rewrote this as a function, which makes it much clearer.
Also handled the case where `otherRelease` is the undeploy version (in that case there was no commit message showing up)

PR #589, opened 2023-03-29
Fix issue with deleting locks

sven-urbanski-freiheit-com:
There was a loop over react hooks, which is not allowed and sometimes caused crashes when deleting locks.

This loop was removed. We now have one hook useAllLocks that returns everything neceessary, instead of first getting all IDs and then later getting the lock by id.

### DIFF
--- a/services/frontend-service/src/ui/utils/store.tsx
+++ b/services/frontend-service/src/ui/utils/store.tsx
@@ -28,13 +28,13 @@ import { useMemo } from 'react';
 import { Empty } from '../../google/protobuf/empty';
 
 export interface DisplayLock {
-    date: Date;
+    date?: Date;
     environment: string;
     application?: string;
     message: string;
     lockId: string;
-    authorName: string;
-    authorEmail: string;
+    authorName?: string;
+    authorEmail?: string;
 }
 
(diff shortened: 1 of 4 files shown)


services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -344,16 +335,50 @@ export const searchCustomFilter = (queryContent: string | null, val: string | un
...
-export const useApplicationLockIDs = (): string[] =>
-    useOverview(({ environments }) =>
-        Object.values(environments)
-            .map((env) => Object.values(env.applications))
-            .flat()
-            .map((app) => Object.values(app.locks))
-            .flat()
-            .map((lock) => lock.lockId)
-    );
+export type AllLocks = {
```
tameremad, line 338: I remember having to change this once before because of the flat mapper error with the store. If it doesn't show anything on the console I would assume it's fine. :thinking:

tameremad (approved):

services/frontend-service/src/ui/utils/store.tsx
sven-urbanski-freiheit-com: I created and deleted a couple locks, and I don't see anything in the console.

PR #594, opened 2023-03-30
New UI is now available under /

sven-urbanski-freiheit-com:
The old UI has been removed.
    
Any unknown paths, e.g. the old /v2 will redirect to /home.


jaypontes:
Even with the `/v2` redirecting to `/home`, shouldn't we remove the `/v2` from [#server.go L230](https://github.com/freiheit-com/kuberpult/blob/su-remove-old-ui/services/frontend-service/pkg/cmd/server.go#L230) and [#server.go L234](https://github.com/freiheit-com/kuberpult/blob/su-remove-old-ui/services/frontend-service/pkg/cmd/server.go#L234)?

jaypontes (approved):

PR #597, opened 2023-03-31
Update alpine image + libgit

hannesg:


services/frontend-service/pnpm-lock.yaml
sven-urbanski-freiheit-com, line 1: the whole file is deleted? Is this correct?

sven-urbanski-freiheit-com (approved):

PR #604, opened 2023-04-04
Bugfix: Allow full page reload in UI

sven-urbanski-freiheit-com:
Before this change, a full page reload in the browser on the pages "/locks" and "/environments" lead to a "not found" or similar message

### DIFF
--- a/services/frontend-service/pkg/cmd/server.go
+++ b/services/frontend-service/pkg/cmd/server.go
@@ -231,7 +231,7 @@ func RunServer() {
 						return
 					}
 				}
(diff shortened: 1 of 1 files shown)


tameremad (approved):
Can you test that the release train still works with this change? 

e.g.
`curl -X PUT localhost:3000/environments/staging/releasetrain`

sven-urbanski-freiheit-com:
> Can you test that the release train still works with this change?
> 
> e.g. `curl -X PUT localhost:3000/environments/staging/releasetrain`

Good catch. Fixed it!

PR #619, opened 2023-04-10
Allow Release Dialog to open via query parameters

tameremad:
The two query parameters `dialog-app` and `dialog-version` open the dialog automatically. the release card would set the parameters to show a new dialog. closing the dialog will simply clear the parameters. 

* delete the `useReleaseInfo` hook from the store.
* `useOpenReleaseDialog` to add the query params for some app/version
* `useCloseReleaseDialog` to clear those params
* `useReleaseDialogParams` to get the app/version from the URL and verify that they represent an actual release. => Then open the dialog.
* remove the release dialog store

### DIFF
--- a/services/frontend-service/src/ui/App/index.tsx
+++ b/services/frontend-service/src/ui/App/index.tsx
@@ -19,7 +19,7 @@ import { ReleaseDialog } from '../components/ReleaseDialog/ReleaseDialog';
 import { PageRoutes } from './PageRoutes';
 import '../../assets/app-v2.scss';
 import * as React from 'react';
-import { PanicOverview, showSnackbarWarn, UpdateOverview, useReleaseDialog, useReleaseInfo } from '../utils/store';
+import { PanicOverview, showSnackbarWarn, UpdateOverview, useValidReleaseDialogParams } from '../utils/store';
 import { useApi } from '../utils/GrpcApi';
 import { AzureAuthProvider, UpdateFrontendConfig, useAzureAuthSub } from '../utils/AzureAuthProvider';
 import { Snackbar } from '../components/snackbar/snackbar';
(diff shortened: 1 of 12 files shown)


services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -175,9 +174,34 @@ export const addAction = (action: BatchAction): void => {
...
+export const useCloseReleaseDialog = (): (() => void) => {
+    const [params, setParams] = useSearchParams();
+    return useCallback(() => {
+        params.delete('dialog-app');
+        params.delete('dialog-version');
+        setParams(params);
+    }, [params, setParams]);
+};
+
+export const useValidReleaseDialogParams = (): [string, number] => {
```
sven-urbanski-freiheit-com, line 195: please return a struct instead of an array, to make it more readable.

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -175,9 +174,34 @@ export const addAction = (action: BatchAction): void => {
...
+    const [params] = useSearchParams();
+    const app = params.get('dialog-app') ?? '';
+    const version = +(params.get('dialog-version') ?? '');
+    const valid = useOverview(({ applications }) =>
+        applications[app] ? !!applications[app].releases.find((r) => r.version === version) : false
+    );
+    return valid ? [app, version] : ['', 0];
```
sven-urbanski-freiheit-com, line 202: please use null/undefined to signify that there is no result.

sven-urbanski-freiheit-com:
Please adapt the commit message (header and body) to explain what you achieved first (then afterwards you can go into details about the changed functions that you mention)
e.g. as a first line
`Allow ReleaseDialog to open via query parameters`

And then something like this: (not 100% correct)
`The query parameters (name1, name2) are used to open he release dialog automatically on the release page. On other pages the parameters will be ignored....`

services/frontend-service/src/ui/utils/store.tsx
sven-urbanski-freiheit-com, line 195:
```suggestion
export const useReleaseDialogParams = (): [string, number] => {
```

tameremad: done :+1:

services/frontend-service/src/ui/App/index.tsx
```diff
@@ -88,13 +88,12 @@ export const App: React.FC = () => {
         }
     );
 
-    const { app, version } = useReleaseDialog(({ app, version }) => ({ app, version }));
-    const releaseInfo = useReleaseInfo(app, version);
+    const { app, version } = useReleaseDialogParams();
```
sven-urbanski-freiheit-com, line 91: please rename this function in the commit message as well.

tameremad: oh yes. right :+1: done

PR #624, opened 2023-04-11
Fix: Azure Auth for batch service in new UI

sven-urbanski-freiheit-com:
Before azure authentication in the new UI only worked for the overviewservice, meaning one could see all the services, but not do anything with them.

Also removed some services that are not needed in the UI.

### DIFF
--- a/services/frontend-service/src/ui/components/SideBar/SideBar.tsx
+++ b/services/frontend-service/src/ui/components/SideBar/SideBar.tsx
@@ -30,6 +30,7 @@ import { ChangeEvent, useCallback, useMemo, useState } from 'react';
 import { useApi } from '../../utils/GrpcApi';
 import { TextField, Dialog, DialogTitle, DialogActions } from '@material-ui/core';
 import classNames from 'classnames';
+import { useAzureAuthSub } from '../../utils/AzureAuthProvider';
 
 export enum ActionTypes {
     Deploy,
@@ -207,6 +208,7 @@ export const SideBar: React.FC<{ className?: string; toggleSidebar: () => void }
     const [lockMessage, setLockMessage] = useState('');
     const api = useApi;
     const [open, setOpen] = useState(false);
+    const { authHeader, authReady } = useAzureAuthSub((auth) => auth);
 
     const handleClose = useCallback(() => setOpen(false), []);
     const handleOpen = useCallback(() => setOpen(true), []);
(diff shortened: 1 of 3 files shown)


services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -235,17 +237,19 @@ export const SideBar: React.FC<{ className?: string; toggleSidebar: () => void }
...
-                showSnackbarSuccess('Actions were applied successfully');
-            })
-            .catch(() => {
-                showSnackbarError('Actions were not applied. Please try again');
-            });
-        handleClose();
-    }, [actions, api, handleClose, lockCreationList, lockMessage]);
+        if (authReady) {
+            api.batchService()
+                .ProcessBatch({ actions }, authHeader)
```
qr4, line 242: how does this work if we dont use azure auth? is authHeader empty?

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -235,17 +237,19 @@ export const SideBar: React.FC<{ className?: string; toggleSidebar: () => void }
...
+                .then((result) => {
+                    deleteAllActions();
+                    showSnackbarSuccess('Actions were applied successfully');
+                })
+                .catch(() => {
```
qr4, line 247: 🔧 can we log the error on console.error? otherwise it is quite hard to figure out what went wrong?

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
sven-urbanski-freiheit-com: yes: https://github.com/freiheit-com/kuberpult/blob/main/services/frontend-service/src/ui/utils/AzureAuthProvider.tsx#L113

tameremad:
Yes, We did have the authHeader in the old UI and it's missing here. 
https://github.com/freiheit-com/kuberpult/blob/93b5b2a177bd4c63cec38fd063291d052ce381a9/services/frontend-service/src/legacy-ui/ActionsCart/CheckoutDialog.tsx#L38-L55

I think this is the fix :+1:

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
sven-urbanski-freiheit-com: good point, done

tameremad (approved):
LGTM :+1:

PR #647, opened 2023-04-17
WIP Add Swagger

sven-urbanski-freiheit-com:


hannesg:
I would really like to generate go code from a spec ( like we do with grpc ) and not generate the spec from code.

PR #660, opened 2023-04-25
Add rollout service base

hannesg:
Add rollout service base.

This adds a new service to the helm chart, the rollout service. It watches the argocd instance, translates commit hashes of the manifests repository to application versions and allows clients to fetch all applications.

The service is not yet used anywhere but already operational.


services/rollout-service/Makefile
```diff
@@ -0,0 +1,88 @@
...
+
+
+IMAGENAME?=$(IMAGE_REGISTRY)/kuberpult-rollout-service:${VERSION}
+
+GARBAGE := *.sentinel
+
+proto:
+	$(MAKE) -C../../pkg/api all
+
+bin/main: $(GO_FILES) | proto
```
sven-urbanski-freiheit-com, line 36: please make `bin/main` also dependent on the go files in /pkg, like here: 
https://github.com/freiheit-com/kuberpult/blob/main/services/cd-service/Makefile#L83

services/rollout-service/pkg/cmd/server.go
```diff
@@ -0,0 +1,167 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+	"google.golang.org/protobuf/types/known/emptypb"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+
+	grpctrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc"
+)
+
+type Config struct {
+	CdServer      string `default:"kuberpult-cd-service:8443"`
```
sven-urbanski-freiheit-com, line 43: I would prefer if we parse variable without envconfig and use `os.Lookup` instead, see here: https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/WORKSPACE_DOCUMENT/-MnQK-Om1MZM1qBy6TV9/

(but this could be done in another PR)

services/rollout-service/pkg/cmd/server.go
```diff
@@ -0,0 +1,167 @@
...
+		Background: []setup.BackgroundTaskConfig{
+			{
+				Name: "consume events",
+				Run: func(ctx context.Context) error {
+					return service.ConsumeEvents(ctx, appClient, versions.New(overview), broadcast, func() { ready = true })
+				},
+			}},
+		GRPC: &setup.GRPCConfig{
+			Port: "8443",
+			/*			Opts: []grpc.ServerOption{
```
sven-urbanski-freiheit-com, line 152: remove the comments please

services/rollout-service/pkg/service/broadcast_test.go
```diff
@@ -0,0 +1,36 @@
...
+		Name string
+	}{
+		{
+			Name: "simple case",
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.Name, func(t *testing.T) {
+
```
sven-urbanski-freiheit-com, line 33: looks like a missing test

services/rollout-service/pkg/service/service.go
```diff
@@ -0,0 +1,104 @@
...
+				Environment:      environment,
+				SyncStatusCode:   ev.Application.Status.Sync.Status,
+				HealthStatusCode: ev.Application.Status.Health.Status,
+				Version:          version,
+			})
+		}
+	}
+}
+
+// We currently exploit that the project is always the environment and the application is env + app joined by "-"
```
sven-urbanski-freiheit-com, line 89: fine for now, but in the future we should use something like labels/annotations

sven-urbanski-freiheit-com (approved):

services/rollout-service/pkg/cmd/server.go
hannesg: We are doing this in many places already and the recommendation is just against default values, not against using envparse. I would post-pone this.

services/rollout-service/pkg/service/service.go
hannesg: Yes, we already have them. I updated the PR to also use them.

services/rollout-service/pkg/cmd/server.go
hannesg: Done, I added the actual tracer.

services/rollout-service/pkg/service/broadcast_test.go
hannesg: I removed it for now.

PR #662, opened 2023-04-28
Add button to lock an entire environment group at once

sven-urbanski-freiheit-com:
This adds a button to add a lock for an entire environment group.
This button will add one lock for each environment in the group to the shopping cart.

And added tests for EnvironmentGroupCard.

Also this changes the css so that the "planned actions" do not overlap with the "delete lock" buttons.

UI after this change:
Screenshot from 2023-04-28 13-47-30

SRX-DDND6N


services/frontend-service/src/ui/components/EnvironmentCard/EnvironmentCard.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { EnvironmentGroupCard } from './EnvironmentCard';
+
+const getNode = (group: EnvironmentGroup): JSX.Element | any => (
+    <MemoryRouter>
+        <EnvironmentGroupCard environmentGroup={group} />
+    </MemoryRouter>
+);
+const getWrapper = (group: EnvironmentGroup) => render(getNode(group));
+
+describe('Test env locks', () => {
```
leandro-freiheit, line 29: I think we should have a test for more than 1 group

services/frontend-service/src/ui/Pages/Environments/EnvironmentsPage.tsx
```diff
@@ -13,17 +13,30 @@ You should have received a copy of the MIT License
...
-import { useEnvironmentNames } from '../../utils/store';
-import { EnvironmentCard } from '../../components/EnvironmentCard/EnvironmentCard';
+import { useEnvironmentGroups, useEnvironmentNames } from '../../utils/store';
+import { EnvironmentCard, EnvironmentGroupCard } from '../../components/EnvironmentCard/EnvironmentCard';
 
 export const EnvironmentsPage: React.FC = () => {
+    const envsGroups = useEnvironmentGroups();
     const envs = useEnvironmentNames();
-
+    const useGroups = envsGroups.length < envs.length;
```
leandro-freiheit, line 22:
```suggestion
    const useGroups = envsGroups.length != envs.length;
```
The current expression may be a bit misleading since we can't have actually have more environment groups than environments

Archomedian:
I think it'd be nice if we added a bit of space between the env group's box and the last env in that env group.
Also we should add the same amount of border radius to the env group's box as the env's box

services/frontend-service/src/ui/components/EnvironmentCard/EnvironmentCard.test.tsx
sven-urbanski-freiheit-com: done (in EnvironmentsPage.test.tsx)

leandro-freiheit (approved):

sven-urbanski-freiheit-com:
> I think it'd be nice if we added a bit of space between the env group's box and the last env in that env group. Also we should add the same amount of border radius to the env group's box as the env's box

Good points, done.

PR #663, opened 2023-04-28
Added git commit parameter and field to the overview service

hannesg:
In the rollout service, I get the currently deployed commit id from argo cd. I need some way to figure out the application version of the services in each environment. The overview endpoint already contains this information and we also have a streaming endpoint for it which would make it a perfect fit to use for that usecase. The only thing that is missing is, is the git commit id and an option to retrieve older overviews.

This commit implements exactly this.

### DIFF
--- a/pkg/api/api.proto
+++ b/pkg/api/api.proto
@@ -144,12 +144,15 @@ service OverviewService {
 }
 
 message GetOverviewRequest {
+  // Retrieve the overview at a certain state of the repository. If it's empty, the latest commit will be used.
+  string git_revision = 1;
 }
 
 message GetOverviewResponse {
   map<string, Environment> environments = 1 [deprecated=true]; // use environmentGroups instead!
   map<string, Application> applications = 2;
   repeated EnvironmentGroup environmentGroups = 3;
+  string git_revision = 4;
 }
 
 message GetDeployedOverviewRequest {
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -635,16 +635,16 @@ func (r *repository) State() *State {
 }
 
 func (r *repository) StateAt(oid *git.Oid) (*State, error) {
-    commit, err := r.repository.LookupCommit(oid)
-		if err != nil {
-			return nil, err
-		}
-      return &State{
-	      Filesystem:             fs.NewTreeBuildFS(r.repository, commit.TreeId()),
-	      Commit:                 commit,
-	      BootstrapMode:          r.config.BootstrapMode,
-	      EnvironmentConfigsPath: r.config.EnvironmentConfigsPath,
-      }, nil
+	commit, err := r.repository.LookupCommit(oid)
+	if err != nil {
+		return nil, err
+	}
+	return &State{
+		Filesystem:             fs.NewTreeBuildFS(r.repository, commit.TreeId()),
+		Commit:                 commit,
+		BootstrapMode:          r.config.BootstrapMode,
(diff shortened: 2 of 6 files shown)


pkg/api/api.proto
```diff
@@ -144,12 +144,15 @@ service OverviewService {
 }
 
 message GetOverviewRequest {
+  // Retrieve the overview at a certain state of the repository. If it's empty, the latest commit will be used.
+  string git_revision = 1;
```
sven-urbanski-freiheit-com, line 148:
```suggestion
  string gitRevision = 1;
```

pkg/api/api.proto
```diff
@@ -144,12 +144,15 @@ service OverviewService {
...
 }
 
 message GetOverviewResponse {
   map<string, Environment> environments = 1 [deprecated=true]; // use environmentGroups instead!
   map<string, Application> applications = 2;
   repeated EnvironmentGroup environmentGroups = 3;
+  string git_revision = 4;
```
sven-urbanski-freiheit-com, line 155:
```suggestion
  string gitRevision = 4;
```

services/cd-service/pkg/repository/repository.go
```diff
@@ -633,6 +634,19 @@ func (r *repository) State() *State {
...
+		return nil, err
+	}
+	return &State{
+		Filesystem:             fs.NewTreeBuildFS(r.repository, commit.TreeId()),
+		Commit:                 commit,
+		BootstrapMode:          r.config.BootstrapMode,
+		EnvironmentConfigsPath: r.config.EnvironmentConfigsPath,
+	}, nil
+}
+
```
sven-urbanski-freiheit-com, line 649: this is very similar to `buildState()`.
Can we make it so that `buildState` gets the git.Oid as parameter and then decides how to build it? (if nil, same as before, if not nil, do `repository.LookupCommit`.

I think that would be a little nicer structure.
You can also make another function
```
func buildState() {
return buildState(nil);
}
```
so you don't have to change all callers.

services/cd-service/pkg/service/overview_test.go
```diff
@@ -420,6 +429,171 @@ func TestOverviewService(t *testing.T) {
 	}
 }
 
+func TestOverviewService2(t *testing.T) {
```
sven-urbanski-freiheit-com, line 432: Maybe
```suggestion
func TestOverviewServiceFromCommit(t *testing.T) {
```
?

pkg/api/api.proto
hannesg: buf and the official protobuf style guide recommend lower snake case for fields: https://protobuf.dev/programming-guides/style/#message-field-names
we mostly follow this and I would rather go with the style guide and fix the remaining occurences where we don't

services/cd-service/pkg/repository/repository.go
hannesg: I thought about this myself but discarded it because buildState has a very specific handlig for when the repository is empty. I did that no. Can you have another look?

pkg/api/api.proto
sven-urbanski-freiheit-com: Ok, then please change the api.proto accordingly.

sven-urbanski-freiheit-com (approved):

pkg/api/api.proto
sven-urbanski-freiheit-com: Or plan a story if it's taking a lot of time. 
Ideally we should also find a linter for this and enable it in the CI

hannesg: I already have a story for the whole buf lint suite. That would cover this as well.

PR #669, opened 2023-05-03
Use same lock ID for locks in one batch

sven-urbanski-freiheit-com:
This will be used later to identify locks by ID.

SRX-VMZUKT


hannesg (commented):
Interesting idea. I would have thought that we would just inject the lock id right before we send the batch, just like we are doing it with the lock comment. I think that would be way simpler because we don't have to juggle react states  around.

sven-urbanski-freiheit-com:
> Interesting idea. I would have thought that we would just inject the lock id right before we send the batch, just like we are doing it with the lock comment. I think that would be way simpler because we don't have to juggle react states around.

I'll try that.

sven-urbanski-freiheit-com:
> > Interesting idea. I would have thought that we would just inject the lock id right before we send the batch, just like we are doing it with the lock comment. I think that would be way simpler because we don't have to juggle react states around.
> 
> I'll try that.

@hannesg  I implemented your suggestion, it is indeed much simpler.

hannesg (approved):

PR #670, opened 2023-05-04
Add button to delete all similar locks

sven-urbanski-freiheit-com:
If a user removes a lock (adds it to the planned actions), kuberpult now checks if there are "similar" locks. We assume a lock is similar, if it has the same ID. Locks can have the same ID if they are different in other fields like "app" and "env".

UI:
Screenshot from 2023-05-04 17-30-54

### DIFF
--- a/services/frontend-service/src/ui/components/SideBar/SideBar.scss
+++ b/services/frontend-service/src/ui/components/SideBar/SideBar.scss
@@ -57,6 +57,35 @@
     @extend .text-regular;
 }
 
+.mdc-drawer-sidebar-list-item-delete-all {
+    @extend .text-regular;
+    display: flex;
+    width: 100%;
+    justify-content: space-between;
+    border-top: solid;
+    padding-top: 10px;
+    margin-top: 10px;
+    .no-wrap {
+        white-space: nowrap;
+    }
+    .mdc-button__label {
+        font-size: 8px;
+        font-weight: 700;
+    }
+    .mdc-button__ripple {
+        position: relative;
+    }
+    button.mdc-button {
+        color: red;
+        border-color: red;
+        border: solid;
+        border-radius: 5px;
+        padding-left: 10px;
+        padding-right: 10px;
+        height: min-content;
+    }
+}
+
 .mdc-drawer-sidebar-list-item-delete-icon {
     flex: 2;
     position: absolute;
--- a/services/frontend-service/src/ui/components/SideBar/SideBar.scss
+++ b/services/frontend-service/src/ui/components/SideBar/SideBar.scss
@@ -69,8 +69,8 @@
         white-space: nowrap;
     }
     .mdc-button__label {
-        font-size: 8px;
-        font-weight: 700;
+        font-size: $small-warn-button-font-size;
+        font-weight: $small-warn-button-font-weight;
     }
     .mdc-button__ripple {
         position: relative;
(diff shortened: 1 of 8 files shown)


ahmed-nour-fdc (approved):

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -26,12 +26,15 @@ import {
...
     randomLockId,
+    addAction,
+    useLocksSimilarTo,
 } from '../../utils/store';
 import { ChangeEvent, useCallback, useMemo, useState } from 'react';
 import { useApi } from '../../utils/GrpcApi';
 import { TextField, Dialog, DialogTitle, DialogActions } from '@material-ui/core';
 import classNames from 'classnames';
 import { useAzureAuthSub } from '../../utils/AzureAuthProvider';
+import * as React from 'react';
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: done

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -176,11 +178,69 @@ export const SideBarListItem: React.FC<{ children: BatchAction }> = ({ children:
     const actionDetails = getActionDetails(action, appLocks, environmentLocks);
 
     const handleDelete = useCallback(() => deleteAction(action), [action]);
+    const similarLocks = useLocksSimilarTo(action);
+    const handleAddAll = useCallback(() => {
+        similarLocks.appLocks.forEach((displayLock: DisplayLock) => {
+            if (!displayLock.environment) {
+                throw new Error('action of type ' + ActionTypes.DeleteEnvironmentLock + ' must have environment set');
```
tameremad, line 185: Is the Action type here correct?
If the environment is not optional do we need to check this here too?

services/frontend-service/src/ui/components/SideBar/SideBar.scss
```diff
@@ -57,6 +57,35 @@
...
+    .mdc-button__ripple {
+        position: relative;
+    }
+    button.mdc-button {
+        color: red;
+        border-color: red;
+        border: solid;
+        border-radius: 5px;
+        padding-left: 10px;
+        padding-right: 10px;
```
tameremad, line 84: nitpick: I'd like to refactor padding / border radius to variables.scss. 
and also the font size just above, maybe we need a new font size definition for tiny buttons?

services/frontend-service/src/ui/components/SideBar/SideBar.scss
```diff
@@ -57,6 +57,35 @@
...
 }
 
+.mdc-drawer-sidebar-list-item-delete-all {
+    @extend .text-regular;
+    display: flex;
+    width: 100%;
+    justify-content: space-between;
+    border-top: solid;
+    padding-top: 10px;
+    margin-top: 10px;
```
tameremad, line 67: Why do we use both margin and padding? (if there is no difference, I'd rather use padding only (or margins only) and not both at the same time)

sven-urbanski-freiheit-com: the difference is that we display the border here, and want the same spacing on the inside and outside.

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
sven-urbanski-freiheit-com: the errors are outdated, I will update them

services/frontend-service/src/ui/components/SideBar/SideBar.scss
tameremad: Okay, I have missed that part. Thank you for explaining :+1:

services/frontend-service/src/ui/components/SideBar/SideBar.scss
sven-urbanski-freiheit-com: done

PR #679, opened 2023-05-05
Fix environment colors on Environments Page

tameremad:
image

### DIFF
--- a/services/frontend-service/src/ui/components/EnvironmentCard/EnvironmentCard.tsx
+++ b/services/frontend-service/src/ui/components/EnvironmentCard/EnvironmentCard.tsx
@@ -13,16 +13,18 @@ You should have received a copy of the MIT License
 along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>.
 
 Copyright 2023 freiheit.com*/
-import { addAction, useFilteredEnvironmentLockIDs } from '../../utils/store';
+import { addAction, useEnvironmentPriorityClassName, useFilteredEnvironmentLockIDs } from '../../utils/store';
 import { Button } from '../button';
 import { Locks } from '../../../images';
 import * as React from 'react';
 import { EnvironmentLockDisplay } from '../EnvironmentLockDisplay/EnvironmentLockDisplay';
 import { EnvironmentGroup } from '../../../api/api';
+import classNames from 'classnames';
 
 export const EnvironmentCard: React.FC<{ environment: string }> = (props) => {
     const { environment } = props;
     const locks = useFilteredEnvironmentLockIDs(environment);
+    const priorityClassName = useEnvironmentPriorityClassName(environment);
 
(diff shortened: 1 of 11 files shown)


tameremad:
https://github.com/freiheit-com/kuberpult/issues/442

services/frontend-service/src/ui/components/EnvironmentCard/EnvironmentCard.tsx
```diff
@@ -62,6 +64,8 @@ export const EnvironmentCard: React.FC<{ environment: string }> = (props) => {
 
 export const EnvironmentGroupCard: React.FC<{ environmentGroup: EnvironmentGroup }> = (props) => {
     const { environmentGroup } = props;
+    // all envs in the same group have the same priority
+    const priorityClassName = useEnvironmentPriorityClassName(environmentGroup.environments[0]?.name);
```
leandro-freiheit, line 68: This does not seem completely right. If the first element in `environmentGroup.environments` has a lower priority but the next element has a higher priority, then the priority of the envGroup will be lower than I expected. Shouldn't we look into every environment in the environment group and have the `priorityClassName` be equal to the highest priority environment in the environment group?

leandro-freiheit: Or do we already sort the environments by priority?

services/frontend-service/src/ui/components/EnvironmentCard/EnvironmentCard.tsx
```diff
@@ -13,16 +13,18 @@ You should have received a copy of the MIT License
...
 import { Locks } from '../../../images';
 import * as React from 'react';
 import { EnvironmentLockDisplay } from '../EnvironmentLockDisplay/EnvironmentLockDisplay';
 import { EnvironmentGroup } from '../../../api/api';
+import classNames from 'classnames';
 
 export const EnvironmentCard: React.FC<{ environment: string }> = (props) => {
     const { environment } = props;
     const locks = useFilteredEnvironmentLockIDs(environment);
+    const priorityClassName = useEnvironmentPriorityClassName(environment);
```
sven-urbanski-freiheit-com, line 27: `useEnvironmentPriorityClassName` should not be a hook.
Suggestion: Make a `useEnvironment` Function
```
const useEnvironment = (string)=>Environment ...
```
Then write another function to get the priorityClassname, e.g.
```
const getPriorityClassName = (env:Environment):string => ...
```

services/frontend-service/src/ui/components/EnvironmentCard/EnvironmentCard.tsx
tameremad: https://github.com/freiheit-com/kuberpult/pull/683

leandro-freiheit (approved):

PR #681, opened 2023-05-05
Bugfix for deleting similar locks

sven-urbanski-freiheit-com:
Also render button on env locks when there is another app lock with the same id


services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -434,8 +434,8 @@ export const useLocksSimilarTo = (cartItemAction: BatchAction | undefined): AllL
         if (isInCart(envLock)) {
             return;
         }
-        // if the id is the same, but we are on a different environment:
-        if (envLock.lockId === data.lockId && envLock.environment !== data.env) {
+        // if the id is the same, but we are on a different environment, or it's an app lock:
+        if ((envLock.lockId === data.lockId && envLock.environment !== data.env) || data.app !== undefined) {
```
tameremad, line 438:
```suggestion
        if (envLock.lockId === data.lockId && (envLock.environment !== data.env || data.app !== undefined)) {
```

tameremad (approved):

PR #683, opened 2023-05-08
Validate environment groups, environments distance to upstream

tameremad:


services/cd-service/pkg/mapper/environments_config_test.go
```diff
@@ -0,0 +1,380 @@
...
+						Environment: nameWhoKnows,
+					},
+					ArgoCd: nil,
+				},
+			},
+			ExpectedResult: []*api.EnvironmentGroup{
+				{
+					EnvironmentGroupName: nameDevDe,
+					Environments: []*api.Environment{
+						makeEnv(nameDevDe, nameDevDe, makeUpstreamLatest(), 0, api.Priority_UPSTREAM),
```
leandro-freiheit, line 188: This will fail because the other environmentGroupName has a non existent upstream, correct?

leandro-freiheit (approved):

services/cd-service/pkg/mapper/environments_config_test.go
tameremad: No this will not fail. Technically, if the real environment setup in the manifest repo was like that, then yes! it will log a warning/fail. 

But this function does not check or validate that, it just assigns a random distance i.e. 667  We don't want this to crash / panic here.

PR #717, opened 2023-06-08
Enable tracing for release endpoint

sven-urbanski-freiheit-com:


hannesg (approved):
Looks okay, though I would really like to have tests for tracing.

Issue #636, opened 2023-04-14
Azure users with non-ascii usernames cannot make changes


sven-urbanski-freiheit-com:
The problem is likely in this line:
https://github.com/freiheit-com/kuberpult/blob/main/pkg/auth/azure.go#L186
Either the claims is not a string or something similar is wrong.

sven-urbanski-freiheit-com:
reported by @qr4

sven-urbanski-freiheit-com:
Fixed with next release 0.4.72

PR #721, opened 2023-06-12
Encode username and mail with base64

sven-urbanski-freiheit-com:
We are sending those as grpc headers, which fails when there are non-ascii chars.

Also removed the "defaultUser" variable, because it can easily be overridden by accident.

This should fix https://github.com/freiheit-com/kuberpult/issues/636

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -62,30 +62,29 @@ func Extract(ctx context.Context) *User {
 		originalEmailArr := md.Get("author-email")
 		if len(originalEmailArr) == 0 {
 			return MakeDefaultUser()
-		} else {
-			originalEmail := originalEmailArr[0]
-			userMail, err := decode64(originalEmail)
-			if err != nil {
-				logger.FromContext(ctx).Warn(fmt.Sprintf("Extract: non-base64 in author-email %s", originalEmail))
-				return MakeDefaultUser()
-			}
-			originalNameArr := md.Get("author-username")
-			if len(originalNameArr) == 0 {
-				logger.FromContext(ctx).Warn(fmt.Sprintf("Extract: username undefined but mail defined %s", userMail))
-				return MakeDefaultUser()
-			}
-			originalName := originalNameArr[0]
-			userName, err := decode64(originalNameArr[0])
-			if err != nil {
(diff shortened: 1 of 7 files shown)


ahmed-nour-fdc (approved):

pkg/auth/auth.go
```diff
@@ -39,12 +66,32 @@ func Extract(ctx context.Context) *User {
...
+			}
+			originalNameArr := md.Get("author-username")
+			if len(originalNameArr) == 0 {
+				logger.FromContext(ctx).Warn(fmt.Sprintf("Extract: username undefined but mail defined %s", userMail))
+				return MakeDefaultUser()
+			}
+			originalName := originalNameArr[0]
+			userName, err := decode64(originalNameArr[0])
+			if err != nil {
+				logger.FromContext(ctx).Warn(fmt.Sprintf("Extract: bbbbb non-base64 in author-username %s", userName))
```
hannesg, line 80:
```
				logger.FromContext(ctx).Warn(fmt.Sprintf("Extract: non-base64 in author-username %s", userName))
```

pkg/auth/auth.go
```diff
@@ -55,12 +95,15 @@ func Extract(ctx context.Context) *User {
...
 	if u == nil || u.Email == "" {
-		u = DefaultUser
+		u = MakeDefaultUser()
 	}
 	// if no username was specified, use email as username
 	if u.Name == "" {
 		u.Name = u.Email
 	}
+
+	u.Name = encode64(u.Name)
```
hannesg, line 105: That looks wrong. Why do you modify the user here?

pkg/auth/auth.go
```diff
@@ -55,12 +95,15 @@ func Extract(ctx context.Context) *User {
...
+	u.Email = encode64(u.Email)
 	ctx = metadata.AppendToOutgoingContext(ctx, "author-email", u.Email, "author-username", u.Name)
```
hannesg, line 107: The grpc standard says you need to suffix binary headers with "-bin". Probably grpc-go is doing something else then: https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#storing-binary-data-in-metadata

sven-urbanski-freiheit-com: I guess that should work too. But now that I have a working solution, I'm a bit hesitant to try out another way.

PR #722, opened 2023-06-12
Allow easier app removal

sven-urbanski-freiheit-com:
This always renders the finalizer and sets AllowEmpty to true.

This is a preparation for SRX-W559H7.

### DIFF
--- a/services/cd-service/pkg/argocd/render.go
+++ b/services/cd-service/pkg/argocd/render.go
@@ -32,9 +32,8 @@ type ApiVersion string
 const V1Alpha1 ApiVersion = "v1alpha1"
 
 type AppData struct {
-	AppName           string
-	TeamName          string
-	IsUndeployVersion bool
+	AppName  string
+	TeamName string
 }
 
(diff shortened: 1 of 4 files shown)


services/cd-service/pkg/argocd/render.go
```diff
@@ -173,10 +171,8 @@ func RenderApp(gitUrl string, gitBranch string, applicationAnnotations map[strin
...
 }
 
-func calculateFinalizers(isUndeployVersion bool) []string {
+func calculateFinalizers() []string {
 	finalizers := []string{}
-	if isUndeployVersion {
-		finalizers = append(finalizers, "resources-finalizer.argocd.argoproj.io")
-	}
+	finalizers = append(finalizers, "resources-finalizer.argocd.argoproj.io")
 	return finalizers
```
hannesg, line 177: You could even make this a constant.

sven-urbanski-freiheit-com: done

PR #731, opened 2023-06-16
Render warnings for unusual deployment situations

sven-urbanski-freiheit-com:
We now render 2 warnings:

1) If "prod" is ahead of "staging", that is unusual. Probably someone deployed a bugfix on prod, but forgot to lock staging.
Screenshot from 2023-06-16 17-04-08

2) If one env has no version deployed but the downstream of it has one. This means the next release train would get rid of it.
Screenshot from 2023-06-16 17-03-13

SRX-S3C8AL

### DIFF
--- a/services/cd-service/pkg/mapper/environments_config_test.go
+++ b/services/cd-service/pkg/mapper/environments_config_test.go
@@ -32,7 +32,7 @@ func makeUpstreamLatest() *api.EnvironmentConfig_Upstream {
 	}
 }
 
-func MakeUpstreamEnvironment(env string) *api.EnvironmentConfig_Upstream {
+func makeUpstreamEnvironment(env string) *api.EnvironmentConfig_Upstream {
 	return &api.EnvironmentConfig_Upstream{
 		Environment: &env,
 	}
@@ -122,7 +122,7 @@ func TestMapEnvironmentsToGroup(t *testing.T) {
 				{
 					EnvironmentGroupName: nameStagingDe,
 					Environments: []*api.Environment{
-						makeEnv(nameStagingDe, nameStagingDe, MakeUpstreamEnvironment(nameDevDe), 1, api.Priority_PROD),
+						makeEnv(nameStagingDe, nameStagingDe, makeUpstreamEnvironment(nameDevDe), 1, api.Priority_PROD),
 					},
 					DistanceToUpstream: 1,
 				},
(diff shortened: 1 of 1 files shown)


hannesg (changes requested):
I think logging this on frontend requests is not a good idea. This will spam our logs with warnings when a user views the page and will be silent when nobody is looking. I would either do this as a frontend-only feature and if you want to write a log then log this when the problem is created instead of when it's viewed.

sven-urbanski-freiheit-com:
> I think logging this on frontend requests is not a good idea. This will spam our logs with warnings when a user views the page and will be silent when nobody is looking. I would either do this as a frontend-only feature and if you want to write a log then log this when the problem is created instead of when it's viewed.

@hannesg which line are you referring to?
I guess this one? https://github.com/freiheit-com/kuberpult/pull/731/files#diff-127dcd8e4f149221e75212d5fb9fed3b0184531ffe4f41952d75d8e787659658R244

I could just remove that - reworking this to calculate the warnings on frontend-only would be a huge change though (And I'm not sure if that's the right direction to go in).

> write a log then log this when the problem is created instead of when it's viewed.
It's already logged on startup.

services/frontend-service/src/ui/components/ServiceLane/Warnings.scss
```diff
@@ -0,0 +1,7 @@
+.service-lane__warning {
```
hannesg, line 1: There is no license header in this file.

services/frontend-service/src/images/ShowBar.svg
hannesg, line 1: I don't think this whitespace change is intentional. Is it?

hannesg (approved):
Let's give it a try!

services/frontend-service/src/images/ShowBar.svg
sven-urbanski-freiheit-com: Actually it is, that is the proper way to format it according to our linter.

services/cd-service/pkg/mapper/environments_config_test.go
```diff
@@ -32,7 +32,7 @@ func makeUpstreamLatest() *api.EnvironmentConfig_Upstream {
 	}
 }
 
-func makeUpstreamEnvironment(env string) *api.EnvironmentConfig_Upstream {
+func MakeUpstreamEnvironment(env string) *api.EnvironmentConfig_Upstream {
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: done

services/frontend-service/src/ui/components/ServiceLane/Warnings.scss
sven-urbanski-freiheit-com: true. We generally don't have that in scss files.
I will create a separate PR to add it.

PR #732, opened 2023-06-16
Also lock when deploying manually

sven-urbanski-freiheit-com:
Manual deployment happen usually when there is an important bugfix to be deployed. In that scenario, usually a lock is a good idea. With this change, the "deploy" button is replaced by "deploy & lock".
Screenshot from 2023-06-16 17-48-11

SRX-S3C8AL

### DIFF
--- a/services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.tsx
+++ b/services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.tsx
@@ -108,7 +108,7 @@ export const EnvironmentListItem: React.FC<EnvironmentListItemProps> = ({
             });
         }
         createAppLock();
-    }, [app, env.name, release.version, createAppLock]);
+    }, [app, env.name, release.version]);
     const queueInfo =
         queuedVersion === 0 ? null : (
             <div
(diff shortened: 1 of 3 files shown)


services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.tsx
```diff
@@ -94,20 +107,8 @@ export const EnvironmentListItem: React.FC<EnvironmentListItemProps> = ({
...
-                createEnvironmentApplicationLock: {
-                    environment: env.name,
-                    application: app,
-                    lockId: '',
-                    message: '',
-                },
-            },
-        });
-    }, [app, env.name]);
+        createAppLock();
```
hannesg, line 110: I'm a bit surprised that you needed to change so few tests. Do we really not check the actions in cart after clicking a button?

hannesg (approved):

services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.tsx
sven-urbanski-freiheit-com: You are right. I added some `expect`s to 2 tests

Issue #712, opened 2023-06-02
Show error when pushing is not possible


sven-urbanski-freiheit-com:
e.g. on bitbucket (or elsewhere) there is an option to protect a branch.
if kuberpult's ssh key has access to read AND write, then it may still fail because of that option.
This currently yields to no logs in the cd service.
There must be at least an error log on cd service. Ideally also in the UI.

sven-urbanski-freiheit-com:
Note that this is really sublte, because the cd service still keeps the commit lying around, so the UI shows the new state. Just the remote repo does not get the updated state.

sven-urbanski-freiheit-com:
It should be easy to reproduce, by giving the ssh key READ only access to the repo.

hannesg:
I debugged this last year already for a day. The thing is, it's ultra-tricky to get a good error log here because from a git perspective ~this is not an error~ because everything works as expected, just no branch got updated. Here is somebody complaining about this: https://github.com/libgit2/libgit2/issues/2628

hannesg:
Ah, found a commit linked in the answer. You have to add a https://pkg.go.dev/github.com/libgit2/git2go/v34?utm_source=godoc#PushUpdateReferenceCallback to the PushOptions and this is called with the ref and a non-empty string if the update was refused. Here is a working go version: https://github.com/rust-lang/crates.io/commit/7f731d4b5a48031ebb6b11503787e0f77a6f1998

sven-urbanski-freiheit-com:
https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/STORY/-NXLNShvmUKMWgw2v9CV/

PR #747, opened 2023-06-22
Do not ignore git push errors

sven-urbanski-freiheit-com:
Do not ignore git push errors
    
Also return readible error in case we cannot push to the manifest repo.
    
2 error cases are handled in this PR:
1) the ssh key does only have read only access. This returns a normal error (at least on github)
2) there is a branch protection rule stopping access. This is not a normal error, but signaled via the "PushUpdateReferenceCallback"

In both cases we return the error to the caller instead of "internal error".

In order to test these 2 error cases, I had to introduce 2 Callbacks for `ProcessQueueOnce`.

This is the solution for https://github.com/freiheit-com/kuberpult/issues/712

### DIFF
--- a/services/cd-service/Makefile
+++ b/services/cd-service/Makefile
@@ -77,7 +77,9 @@ run: docker
 	docker run --rm -e KUBERPULT_GIT_URL="/repository_remote" -e KUBERPULT_GIT_BRANCH=master -p 8080:8080 -p 8443:8443 -v $(shell pwd)/repository_remote:/repository_remote $(IMAGENAME)
 else
 run: build
-	KUBERPULT_GIT_URL="./repository_remote" KUBERPULT_GIT_BRANCH=master ./bin/main
+	#KUBERPULT_GIT_URL="./repository_remote" KUBERPULT_GIT_BRANCH=master ./bin/main
+#	KUBERPULT_GIT_URL="git@github.com:sven-urbanski-freiheit-com/manifest-test.git" KUBERPULT_GIT_BRANCH="main" KUBERPULT_GIT_SSH_KEY="../../private.key" KUBERPULT_GIT_SSH_KNOWN_HOSTS=../../known_hosts_github  ./bin/main
+	KUBERPULT_GIT_URL="git@github.com:freiheit-com/fdc-standard-setup-dev-env-manifest.git" KUBERPULT_GIT_BRANCH="main" KUBERPULT_GIT_SSH_KEY="../../private.key" KUBERPULT_GIT_SSH_KNOWN_HOSTS=../../known_hosts_github  ./bin/main
 endif
 
 bin/main: bin/ $(ALL_GO_FILES) | proto
--- a/services/cd-service/Makefile
+++ b/services/cd-service/Makefile
@@ -77,9 +77,7 @@ run: docker
 	docker run --rm -e KUBERPULT_GIT_URL="/repository_remote" -e KUBERPULT_GIT_BRANCH=master -p 8080:8080 -p 8443:8443 -v $(shell pwd)/repository_remote:/repository_remote $(IMAGENAME)
 else
 run: build
-	#KUBERPULT_GIT_URL="./repository_remote" KUBERPULT_GIT_BRANCH=master ./bin/main
-#	KUBERPULT_GIT_URL="git@github.com:sven-urbanski-freiheit-com/manifest-test.git" KUBERPULT_GIT_BRANCH="main" KUBERPULT_GIT_SSH_KEY="../../private.key" KUBERPULT_GIT_SSH_KNOWN_HOSTS=../../known_hosts_github  ./bin/main
-	KUBERPULT_GIT_URL="git@github.com:freiheit-com/fdc-standard-setup-dev-env-manifest.git" KUBERPULT_GIT_BRANCH="main" KUBERPULT_GIT_SSH_KEY="../../private.key" KUBERPULT_GIT_SSH_KNOWN_HOSTS=../../known_hosts_github  ./bin/main
+	KUBERPULT_GIT_URL="./repository_remote" KUBERPULT_GIT_BRANCH=master ./bin/main
 endif
 
 bin/main: bin/ $(ALL_GO_FILES) | proto
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -279,7 +279,7 @@ func (r *repository) ProcessQueue(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case e := <-r.queue.elements:
-			r.ProcessQueueOnce(ctx, e)
+			r.ProcessQueueOnce(ctx, e, DefaultPushUpdate, DefaultPushActionCallback)
 		}
 	}
 }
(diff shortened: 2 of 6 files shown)


services/cd-service/Makefile
```diff
@@ -77,7 +77,9 @@ run: docker
 	docker run --rm -e KUBERPULT_GIT_URL="/repository_remote" -e KUBERPULT_GIT_BRANCH=master -p 8080:8080 -p 8443:8443 -v $(shell pwd)/repository_remote:/repository_remote $(IMAGENAME)
 else
 run: build
-	KUBERPULT_GIT_URL="./repository_remote" KUBERPULT_GIT_BRANCH=master ./bin/main
+	#KUBERPULT_GIT_URL="./repository_remote" KUBERPULT_GIT_BRANCH=master ./bin/main
```
leandro-freiheit, line 80: Is these comment supposed to be here?

services/cd-service/pkg/repository/repository.go
```diff
@@ -368,7 +393,9 @@ func (r *repository) ProcessQueueOnce(e element) {
 	}
 
 	// Try pushing once
-	err = r.Push(e.ctx, pushAction)
+	logger.Error(fmt.Sprintf("SU DEBUG before push"))
```
leandro-freiheit, line 396: I think this should be removed

services/cd-service/pkg/repository/repository.go
```diff
@@ -382,15 +409,20 @@ func (r *repository) ProcessQueueOnce(e element) {
...
+			err = httperrors.PublicError(ctx, errors.New(fmt.Sprintf("could not push to manifest repository '%s' on branch '%s' - this indicates that the ssh key does not have write access", r.config.URL, r.config.Branch)))
+		}
+	} else {
+		if !pushSuccess {
+			err = fmt.Errorf("failed to push - this indicates that branch protection is enabled in '%s' on branch '%s'", r.config.URL, r.config.Branch)
 		}
 	}
-	err = nil
 	r.notify.Notify()
+	logger.Error(fmt.Sprintf("SU DEBUG after notify"))
```
leandro-freiheit, line 425: This should also be removed

services/cd-service/Makefile
```diff
@@ -77,7 +77,9 @@ run: docker
...
+#	KUBERPULT_GIT_URL="git@github.com:sven-urbanski-freiheit-com/manifest-test.git" KUBERPULT_GIT_BRANCH="main" KUBERPULT_GIT_SSH_KEY="../../private.key" KUBERPULT_GIT_SSH_KNOWN_HOSTS=../../known_hosts_github  ./bin/main
+	KUBERPULT_GIT_URL="git@github.com:freiheit-com/fdc-standard-setup-dev-env-manifest.git" KUBERPULT_GIT_BRANCH="main" KUBERPULT_GIT_SSH_KEY="../../private.key" KUBERPULT_GIT_SSH_KNOWN_HOSTS=../../known_hosts_github  ./bin/main
```
hannesg: Is this change intended?

services/cd-service/pkg/repository/repository.go
sven-urbanski-freiheit-com: done

services/cd-service/pkg/repository/repository.go
```diff
@@ -328,7 +329,32 @@ func (r *repository) drainQueue() []element {
...
 
-func (r *repository) ProcessQueueOnce(e element) {
+// DefaultPushUpdate is public for testing reasons only.
+// It returns always nil
+// success is set to true if the push was successful
+func DefaultPushUpdate(branch string, success *bool) git.PushUpdateReferenceCallback {
+	return func(refName string, status string) error {
+		var expectedRefName = fmt.Sprintf("refs/heads/%s", branch)
+		// if we were successful the status is empty and the ref contains our branch:
+		*success = refName == expectedRefName && status == ""
```
hannesg, line 339: Uh, this looks hard.

services/cd-service/Makefile
sven-urbanski-freiheit-com: i removed them

services/cd-service/pkg/repository/repository.go
sven-urbanski-freiheit-com: i removed it

services/cd-service/pkg/service/batch.go
```diff
@@ -193,7 +192,7 @@ func (d *BatchServer) ProcessBatch(
 	}
 	err := d.Repository.Apply(ctx, transformers...)
 	if err != nil {
-		return nil, httperrors.InternalError(ctx, fmt.Errorf("could not apply transformer: %w", err))
+		return nil, err
```
hannesg, line 195: I'm a bit sad that this didn't trigger any tests.

leandro-freiheit (approved):

services/cd-service/pkg/repository/repository.go
```diff
@@ -382,14 +407,18 @@ func (r *repository) ProcessQueueOnce(e element) {
...
 				return
 			}
-			if pushErr := r.Push(e.ctx, pushAction); pushErr != nil {
+			if pushErr := r.Push(e.ctx, pushAction(pushOptions, r)); pushErr != nil {
 				err = &InternalError{inner: pushErr}
 			}
 		} else {
-			err = &InternalError{inner: err}
+			logger.Error(fmt.Sprintf("error while pushing: %s", err))
+			err = httperrors.PublicError(ctx, errors.New(fmt.Sprintf("could not push to manifest repository '%s' on branch '%s' - this indicates that the ssh key does not have write access", r.config.URL, r.config.Branch)))
```
hannesg, line 415: Most callers are not http but grpc. How will they display this error?

hannesg:
I think you could construct a very realistic test-case by adding running the cd-service against a bare repo with an `update` hook.

services/cd-service/pkg/repository/repository.go
sven-urbanski-freiheit-com: I don't know what this means, but I'm also not super happy with this code.
Do you think it won't work in some cases?
I tried it on GitHub and it works fine.

services/cd-service/pkg/repository/repository.go
sven-urbanski-freiheit-com: the package name is misleading. It is actually returning the error fine for grpc, I tested it with `evans`.

sven-urbanski-freiheit-com:
> I think you could construct a very realistic test-case by adding running the cd-service against a bare repo with an `update` hook.

I tested this manually with a real github repo.

services/cd-service/pkg/repository/repository.go
sven-urbanski-freiheit-com: After lots of experimentation, I think I'll have to leave it as it is.
We tried running `git init` in a test with a server-side hook in the repo, to reproduce this case where the pushing is not successful. This would avoid this rather ugly callback with the `success` pointer. However, git hooks, incl server-hooks are not supported in this setup, and seem to generally not supported by gitlib2 https://github.com/libgit2/libgit2/blob/main/docs/projects.md?plain=1#L74

PR #760, opened 2023-06-27
Adding username/email as datadog tags to span

sven-urbanski-freiheit-com:
Also added a new span to see the argoCd rendering in datadog.

### DIFF
--- a/services/frontend-service/pkg/cmd/server.go
+++ b/services/frontend-service/pkg/cmd/server.go
@@ -351,13 +351,23 @@ func getRequestAuthorFromAzure(r *http.Request) *auth.User {
 
 func (p *Auth) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	logger.Wrap(r.Context(), func(ctx context.Context) error {
+		var u *auth.User = nil
+		var source = ""
 		if c.AzureEnableAuth {
-			u := getRequestAuthorFromAzure(r)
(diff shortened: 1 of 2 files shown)


services/frontend-service/pkg/cmd/server.go
```diff
@@ -351,13 +351,23 @@ func getRequestAuthorFromAzure(r *http.Request) *auth.User {
...
-			p.HttpServer.ServeHTTP(w, r.WithContext(auth.ToContext(ctx, u)))
+			u = getRequestAuthorFromAzure(r)
+			source = "azure"
 		} else {
-			u := getRequestAuthorFromGoogleIAP(ctx, r)
-			p.HttpServer.ServeHTTP(w, r.WithContext(auth.ToContext(ctx, u)))
+			u = getRequestAuthorFromGoogleIAP(ctx, r)
+			source = "iap"
 		}
+		span, ctx := tracer.StartSpanFromContext(ctx, "ServeHTTP")
```
hannesg, line 363: I think this is very subtle that we start the tracing span in the Auth type. I have two ideas in mind:

1. Rename the Auth type to something more generic.
2. Keep starting the span outside of the Auth type and use [SpanFromContext](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1@v1.52.0/ddtrace/tracer#SpanFromContext) to attach the tags.

hannesg (approved):
Let's do this now

PR #765, opened 2023-06-28
Make git author configurable in helm chart & refactor context usage

sven-urbanski-freiheit-com:
Make git author configurable in helm chart & refactor context usage
    
This is technically not a breaking change, but it is recommended to now set the author and email when invoking the /release endpoint. Note that both headers must be base-64 encoded. See create-release.sh for a valid example call with headers. If you do not provide these headers, the git author will be set to "empty".
    
New helm chart variables can be set, but they come with the same default values as before:
    
* git.author.name
* git.author.email
    
This also refactored how to access the context.
    
The cd-service now always* returns an error when the context should be there but is not, instead of always falling back to default values.
    
Exception is the release endpoint, as this would be a breaking change. (And of course /health does not require any headers)
    
The frontend-service still uses the headers if supplied, but falls back to the default values that are configured in the helm chart.
    
This effects all endpoints, http and grpc.
    
The context functions (formerly "Extract", etc) have been refactored, so they do exactly one thing, e.g. reading from one context.
There are 3 options: grpc context, go context and http headers. All have a function for reading and writing respectively.
    
Story SRX-6SMIU3

### DIFF
--- a/charts/kuberpult/templates/frontend-service.yaml
+++ b/charts/kuberpult/templates/frontend-service.yaml
@@ -83,6 +83,10 @@ spec:
             cpu: "{{ .Values.frontend.resources.requests.cpu }}"
             memory: "{{ .Values.frontend.resources.requests.memory }}"
         env:
+        - name: KUBERPULT_GIT_AUTHOR_NAME
+          value: {{ .Values.git.author.name | quote }}
+        - name: KUBERPULT_GIT_AUTHOR_EMAIL
+          value: {{ .Values.git.author.email | quote }}
         - name: KUBERPULT_CDSERVER
           value: kuberpult-cd-service:8443
         - name: KUBERPULT_HTTP_CD_SERVER
--- a/infrastructure/scripts/create-testdata/create-release.sh
+++ b/infrastructure/scripts/create-testdata/create-release.sh
@@ -44,14 +44,14 @@ echo ${msgs[$index]}" "$commit_id > "${commit_message_file}"
 
 ls ${commit_message_file}
 
-release_version=()
+release_version=''
 case "${RELEASE_VERSION:-}" in
 	'') echo "No RELEASE_VERSION set. Using non-idempotent versioning scheme";;
 	*[!0-9]*) echo "Please set the env variable RELEASE_VERSION to a number"; exit 1;;
-	*) release_version=(--form-string "version=${RELEASE_VERSION:-}");;
+	*) release_version='--form-string '"version=${RELEASE_VERSION:-}";;
 esac
 
-echo "release version:" "${release_version[@]}"
+echo "release version:" "${release_version}"
 
 configuration=()
 configuration+=("--form" "team=${applicationOwnerTeam}")
@@ -77,11 +77,13 @@ done
 echo commit id: ${commit_id}
 
 
-curl http://localhost:8081/release \
+curl http://localhost:8080/release \
+  -H "author-email:dW5kZWZpbmVkLXVzZXJAZXhhbXBsZS5jb20=" \
+  -H "author-name:dW5kZWZpbmVkLXVzZXI=" \
   --form-string "application=$name" \
   --form-string "source_commit_id=${commit_id}" \
   --form-string "source_author=${author}" \
-  "${release_version[@]}" \
+  ${release_version} \
   --form "source_message=<${commit_message_file}" \
   "${configuration[@]}" \
   "${manifests[@]}"
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -19,84 +19,133 @@ package auth
 import (
 	"context"
 	"encoding/base64"
+	"errors"
 	"fmt"
 
 	"github.com/freiheit-com/kuberpult/pkg/logger"
+	"github.com/freiheit-com/kuberpult/services/cd-service/pkg/httperrors"
 	"google.golang.org/grpc/metadata"
+	"net/http"
 )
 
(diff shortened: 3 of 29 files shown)


services/frontend-service/pkg/handler/release.go
```diff
@@ -28,6 +30,17 @@ func (s Server) HandleRelease(w http.ResponseWriter, req *http.Request, tail str
...
 		return
 	}
+	ctx := req.Context()
+	// we just always use the default user here:
+	user := &auth.User{
+		Email: s.Config.GitAuthorEmail,
+		Name:  s.Config.GitAuthorName,
+	}
+	auth.WriteUserToHttpHeader(req, *user, ctx)
+	logger.FromContext(ctx).Warn(fmt.Sprintf("HandleRelease: wrote name %s. mail: %s", user.Name, user.Email))
```
hannesg, line 40: This warning shouldn't be here, I guess

pkg/auth/auth.go
```diff
@@ -51,51 +57,102 @@ func decode64(s string) (string, error) {
 	return string(b), err
 }
 
-// Extract takes the User from middleware.
-// It always returns a User
-func Extract(ctx context.Context) *User {
-	// check if User is in go Context
+// ReadUserFromContextOrDie returns a user from the ctx or an error if none was found, or it is invalid
+func ReadUserFromContextOrDie(ctx context.Context) (*User, error) {
```
hannesg, line 61: `OrDie` usually implies that the function panics but this one does not.

pkg/auth/auth.go
```diff
@@ -51,51 +57,102 @@ func decode64(s string) (string, error) {
...
+		}, nil
+	}
+	return nil, httperrors.AuthError(ctx, errors.New("ExtractUserHttp: did not find data in headers"))
+}
+
+// WriteUserToHttpHeader should only be used in the frontend-service
+// WriteUserToHttpHeader writes the user into http headers
+// it is used for requests like /release which are delegated from frontend-service to cd-service
+func WriteUserToHttpHeader(r *http.Request, user User, ctxLOG context.Context) {
+	r.Header.Set(HttpHeaderUserName, encode64(user.Name))
```
hannesg, line 147: Is there any deep reason why you don't use the same header here as for grpc?

hannesg (commented):
Please remove the logs :)

pkg/auth/auth.go
sven-urbanski-freiheit-com: no. I'll change it to use the same

services/frontend-service/pkg/handler/release.go
sven-urbanski-freiheit-com: done

pkg/auth/auth.go
sven-urbanski-freiheit-com: done

pkg/auth/auth.go
sven-urbanski-freiheit-com: i renamed the functions

services/cd-service/pkg/httperrors/httperrors.go
```diff
@@ -35,3 +35,9 @@ func PublicError(ctx context.Context, err error) error {
 	logger.Error("grpc.public", zap.Error(err))
 	return status.Error(codes.InvalidArgument, "error: "+err.Error())
 }
+
+func AuthError(ctx context.Context, err error) error {
+	logger := logger.FromContext(ctx)
+	logger.Error("grpc.auth", zap.Error(err))
```
hannesg, line 41: This is hard against our engineering standards: handle OR log an error but not both.

hannesg (approved):

infrastructure/scripts/create-testdata/create-release.sh
```diff
@@ -77,7 +77,9 @@ done
 echo commit id: ${commit_id}
 
 
-curl http://localhost:8081/release \
+curl http://localhost:8080/release \
+  -H "user-email-http:dW5kZWZpbmVkLXVzZXJAZXhhbXBsZS5jb20=" \
+  -H "user-name-http:dW5kZWZpbmVkLXVzZXI=" \
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: this is consistently "user-email" now

charts/kuberpult/templates/frontend-service.yaml
```diff
@@ -83,6 +83,10 @@ spec:
             cpu: "{{ .Values.frontend.resources.requests.cpu }}"
             memory: "{{ .Values.frontend.resources.requests.memory }}"
         env:
+        - name: KUBERPULT_GIT_AUTHOR_NAME
+          value: "{{ .Values.git.author.name }}"
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: done

services/cd-service/pkg/httperrors/httperrors.go
sven-urbanski-freiheit-com: done

charts/kuberpult/templates/frontend-service.yaml
```diff
@@ -83,6 +83,10 @@ spec:
...
+        - name: KUBERPULT_GIT_AUTHOR_EMAIL
+          value: "{{ .Values.git.author.email }}"
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: done

hannesg (approved):

PR #767, opened 2023-06-29
Adding Deployment Author file to cd-service

gsandok:
Creating file to hold Deployment Author and time of deployment within manifest repo
To be displayed in frontend in later PR


pkg/api/api.proto
```diff
@@ -253,6 +253,13 @@ message UpstreamNotDeployed {
 message Environment {
 
   message Application {
+    message Environment {
+      message DeploymentMetaData {
+        string deployAuthor = 1;
```
hannesg, line 258: Shouldn't these fields be in the Environment_Application?

services/cd-service/pkg/repository/transformer.go
```diff
@@ -869,10 +869,36 @@ func (c *DeployApplicationVersion) Transform(ctx context.Context, state *State)
 	if err := util.WriteFile(fs, manifestFilename, manifestContent, 0666); err != nil {
 		return "", err
 	}
+
+	deployInfo := auth.Extract(ctx).Name + "; " + time.Now().UTC().String() + "\n"
+	deploymentFilename := fs.Join(releaseDir, "environments", c.Environment, "deployment_metadata.yaml")
```
hannesg, line 874: Please write the data as dedicated files (one per field) like it's done for locks here: https://github.com/freiheit-com/kuberpult/blob/300a1f54f5818b80a3bb338a5428612658fbcb0a/services/cd-service/pkg/repository/transformer.go#L607C11-L607C11

pkg/api/api.proto
```diff
@@ -273,6 +280,7 @@ message Environment {
     // google.protobuf.Timestamp deploy_date = 5; // This is never used
     bool undeployVersion = 6;
     ArgoCD argoCD = 7;
+    map<string, Environment> environments = 8;
```
hannesg, line 283: Why do you have multiple environments in one again?

pkg/api/api.proto
```diff
@@ -253,6 +253,13 @@ message UpstreamNotDeployed {
 message Environment {
 
   message Application {
+    message Environment {
+      message DeploymentMetaData {
+        string deployAuthor = 1;
+        string deployTime = 2;
```
sven-urbanski-freiheit-com, line 259:
```suggestion
        int64 deployAtSecondsUtc = 2;
```
It should be clear that it's a timestamp, it's in UTC, and it should be an int of some form (maybe int32 is enough).

hannesg (commented):
Hi. What exactly do you want to display in the frontend? Maybe it's easier to write a deployment log or something like this to the application version and then make a dedicated endpoint for that?

services/cd-service/pkg/repository/repository.go
```diff
@@ -894,6 +899,25 @@ func (s *State) GetEnvironmentApplicationLocks(environment, application string)
 	}
 }
 
+func (s *State) GetDeploymentMetaData(environment, application string) (deployAuthor, deployTime string, err error) {
+	base := s.Filesystem.Join("environments", environment, "applications", application, "version", "environments", environment)
+	logs, err := readFile(s.Filesystem, s.Filesystem.Join(base, "deployment_metadata.yaml"))
```
sven-urbanski-freiheit-com, line 904: it's not yaml though, is it?
please remove the yaml suffix. Maybe replace with ".csv"

pkg/api/api.proto
hannesg: int32 is never enough

services/cd-service/pkg/repository/repository.go
```diff
@@ -894,6 +899,25 @@ func (s *State) GetEnvironmentApplicationLocks(environment, application string)
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return "defaultUser", "", nil
```
sven-urbanski-freiheit-com, line 907: ahhh, I'm just refactoring the defaultUser.
for now, please use the constant in auth.go.
But depending on who merges first, we need to adapt this.

services/cd-service/pkg/repository/repository.go
```diff
@@ -894,6 +899,25 @@ func (s *State) GetEnvironmentApplicationLocks(environment, application string)
...
+		} else {
+			return "", "", err
+		}
+	}
+	readLines := strings.Split(string(logs), "\n")
+	// avoid that file ends in newline
+	results := strings.Split(readLines[len(readLines)-2], "; ")
+	if len(results) != 2 {
+		return "defaultUser", "", nil
```
sven-urbanski-freiheit-com, line 916: here too

services/cd-service/pkg/repository/repository.go
```diff
@@ -894,6 +899,25 @@ func (s *State) GetEnvironmentApplicationLocks(environment, application string)
...
+	if err != nil {
+		if os.IsNotExist(err) {
+			return "defaultUser", "", nil
+		} else {
+			return "", "", err
+		}
+	}
+	readLines := strings.Split(string(logs), "\n")
+	// avoid that file ends in newline
+	results := strings.Split(readLines[len(readLines)-2], "; ")
```
sven-urbanski-freiheit-com, line 914: think this will fail if logs is an empty string

services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: We will store multiple deployments. I think for this case it's more consistent to have one file with multiple entries.
Otherwise we can get into odd edgecases where there's 1 author and 2 timestamps.

(story: https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-NQPLdkfXUNc4bE9KOiO/FILESYSTEM/FOLDER/ROOT/STORY/-NWiltvEUw2EQt4SdlvI?iteration=CURRENT_ITERATIONS&user=-NHUO4EmpZZtkcSSqr1i)

Alternative: We create a directory with an array (like "releases") and have single files in `deploymentMetadata/1/author` and `deploymentMetadata/1/deploymentTimestamp`

What do you think @hannesg ?

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -393,11 +393,12 @@ func TestCreateUndeployApplicationVersionErrors(t *testing.T) {
...
-		Transformers     []Transformer
-		expectedError    string
-		expectedPath     string
-		expectedFileData []byte
+		Name                           string
+		Transformers                   []Transformer
+		expectedError                  string
+		expectedPath                   string
+		expectedFileData               []byte
+		expectedDeploymentMetadataPath string
```
sven-urbanski-freiheit-com, line 401: please also add a filed for `expectedDeploymentMetadataContent string` and test that the content is as expected.

services/cd-service/pkg/service/overview.go
```diff
@@ -172,6 +173,22 @@ func (o *OverviewServiceServer) getOverview(
 							}
 						}
 					}
+					if appEnvs, err := s.GetAppsEnvs(envName, appName); err != nil {
+						return nil, err
+					} else {
```
sven-urbanski-freiheit-com, line 178: since you have a return in the `then` part, 
please remove the `else` to avoid unnecessary nesting

pkg/api/api.proto
sven-urbanski-freiheit-com: ahh yes, I think that was my suggestion. But now I see the redundancy. I also think it should go in the `Environment_Application`.

pkg/api/api.proto
```diff
@@ -253,6 +253,13 @@ message UpstreamNotDeployed {
 message Environment {
 
   message Application {
+    message Environment {
+      message DeploymentMetaData {
```
sven-urbanski-freiheit-com, line 257: to be consistent, we should write both the author name and author email.

services/cd-service/pkg/repository/repository.go
```diff
@@ -894,6 +899,25 @@ func (s *State) GetEnvironmentApplicationLocks(environment, application string)
 	}
 }
 
+func (s *State) GetDeploymentMetaData(environment, application string) (deployAuthor, deployTime string, err error) {
```
sven-urbanski-freiheit-com, line 902: should be:
`/environments/development/applications
/echo/deployment_author`
`/environments/development/applications
/echo/deployment_time_utc`

pkg/api/api.proto
```diff
@@ -261,6 +261,10 @@ message Environment {
       }
       repeated SyncWindow syncWindows = 1;
     }
+    message DeploymentMetaData {
+      string deployAuthor = 1;
+      string deployTime = 2;
```
sven-urbanski-freiheit-com, line 266: it shouldn't be a string here, but an uint64

sven-urbanski-freiheit-com, line 266:
```suggestion
      // we use a string here, because the UI cannot handle int64 as a type.
      // the string contains the unix timestamps in seconds (utc)
      string deployTime = 2;
```

sven-urbanski-freiheit-com (approved):

sven-urbanski-freiheit-com:
Before merging, please adapt the commit message, e.g.
> Store person who triggered deployment
>
> Stores users name and current date in the manifest repository.
> To be displayed in frontend in later PR.

services/cd-service/pkg/repository/repository.go
```diff
@@ -894,6 +894,34 @@ func (s *State) GetEnvironmentApplicationLocks(environment, application string)
 	}
 }
 
+func (s *State) GetDeploymentMetaData(environment, application string) (string, time.Time, error) {
+	base := s.Filesystem.Join("environments", environment, "applications", application)
+	author, err := readFile(s.Filesystem, s.Filesystem.Join(base, "deployed_by"))
```
hannesg, line 899: No email here?

hannesg (commented):
Didn't you miss saving the email address? But the code looks nice!

services/cd-service/pkg/repository/transformer.go
```diff
@@ -869,6 +869,15 @@ func (c *DeployApplicationVersion) Transform(ctx context.Context, state *State)
 	if err := util.WriteFile(fs, manifestFilename, manifestContent, 0666); err != nil {
 		return "", err
 	}
+
+	if err := util.WriteFile(fs, fs.Join(applicationDir, "deployed_by"), []byte(auth.Extract(ctx).Name), 0666); err != nil {
```
sven-urbanski-freiheit-com, line 873: if you merge this after my change, you need to access the name like this:
```
	user, err := auth.ReadUserFromContext(ctx)
	if err != nil {
		return err
	}
       /// ... auth.Name
```

sven-urbanski-freiheit-com: by "my change" i mean this: https://github.com/freiheit-com/kuberpult/pull/765

services/cd-service/pkg/repository/repository.go
gsandok: Was not mentioned in the meeting we had on Thursday but can add now

hannesg (approved):

PR #779, opened 2023-06-30
Refactoring: Move interceptors to frontend

sven-urbanski-freiheit-com:
They are only used in the frontend, and it doesn't make much sense to share them, given their purpose


jaypontes:
Sorry I'm not really in the loop regarding these changes so I don't feel appropriate reviewing them. Sorry! :(

jaypontes (approved):

PR #782, opened 2023-07-03
Adds DEX methods to Auth Package

bernardo-bastos:
Creates DEX methods on the auth package. Here is a brief description of the most important methods:
- NewDexAppClient: Creates a Dex client with the initial configuration. The values to configure it will be later passed via the Helm chart installation.
- NewDexReverseProxy: Creates a reverse proxy for Dex server to avoid exposing it via a virtual service. 
- handleDexLogin: Handles Dex Login.
- handleCallback: Handles Dex callback once after the Log in.


pkg/auth/dex.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// Redirects to the Dex login page with the pre configured connector.
+func (a *DexAppClient) handleDexLogin(w http.ResponseWriter, r *http.Request) {
+	oauthConfig, err := a.oauth2Config(a.Scopes)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
```
leandro-freiheit, line 150: Will this be addressed in later on?

pkg/auth/dex.go
```diff
@@ -0,0 +1,247 @@
...
+		ClientID:     a.ClientID,
+		ClientSecret: a.ClientSecret,
+		Endpoint:     p.Endpoint(),
+		Scopes:       scopes,
+		RedirectURL:  a.RedirectURI,
+	}, nil
+}
+
+// TODO(BB): Verifies if the token has the cookie and validates it.
+func (a *DexAppClient) verifyToken(r *http.Request) error {
```
leandro-freiheit, line 245: Any reason why this function is not finished?

pkg/auth/dex_test.go
```diff
@@ -0,0 +1,135 @@
...
+			wantStatusCode: http.StatusOK,
+			wantErr:        false,
+		},
+		{
+			Name: "Dex reverse proxy is working as expected on error",
+			mockDexServer: httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+				rw.WriteHeader(http.StatusInternalServerError)
+			})),
+			wantStatusCode: http.StatusInternalServerError,
+			wantErr:        false,
```
leandro-freiheit, line 82: Just a comment, I understand the wantErr differs from the error status code but it can be misleading. I would be expecting that when the status code is 500 that I want an error in fact. I may be nitpicking though.

pkg/auth/dex.go
```diff
@@ -0,0 +1,247 @@
...
+	http.HandleFunc(loginPATH, a.handleDexLogin)
+	// Call back to the current app once the login is finished
+	http.HandleFunc(callbackPATH, a.handleCallback)
+}
+
+// NewDexReverseProxy returns a reverse proxy to the Dex server.
+func NewDexReverseProxy(serverAddr string) func(writer http.ResponseWriter, request *http.Request) {
+	target, err := url.Parse(serverAddr)
+	if err != nil {
+		logger.FromContext(context.Background()).Warn(fmt.Sprintf("Could not parse server URL with error: %s", err))
```
sven-urbanski-freiheit-com, line 111: we continue here despite the error?

pkg/auth/dex.go
```diff
@@ -0,0 +1,247 @@
...
+			if err != nil {
+				return err
+			}
+			err = resp.Body.Close()
+			if err != nil {
+				return err
+			}
+			resp.Body = io.NopCloser(bytes.NewReader(make([]byte, 0)))
+			return nil
+		}
```
sven-urbanski-freiheit-com, line 127: what does this block achieve?
If there is an internal error, we read the body and discard it?

pkg/auth/dex.go
```diff
@@ -0,0 +1,247 @@
...
+func (s DexRewriteURLRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	r.URL.Host = s.DexURL.Host
+	r.URL.Scheme = s.DexURL.Scheme
+	r.Host = s.DexURL.Host
+	return s.T.RoundTrip(r)
+}
+
+// Registers dex handlers for login
+func (a *DexAppClient) registerDexHandlers() {
+	// Handles the login callback to redirect to dex page.
```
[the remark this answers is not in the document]
bernardo-bastos: It was not correct. Fixed the comment.

pkg/auth/dex_test.go
bernardo-bastos: Yes, you are right. The `wantErr` was not doing anything in there. I removed the variable from the test.

pkg/auth/dex.go
bernardo-bastos: Added a logger with `Error()` and `return nil` statement.

pkg/auth/dex.go
bernardo-bastos: Forgot to log the message. Added an `Error()` logger message.

pkg/auth/dex.go
bernardo-bastos: I did not want to make the PR too extensive, but I guess one method wont make a big difference 😅 I added it.

pkg/auth/dex_test.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func TestDexRoundTripper(t *testing.T) {
+	testCases := []struct {
+		Name           string
+		mockDexServer  *httptest.Server
+		wantStatusCode int
+	}{
+		{
+			Name: "Round tripper works as expected",
```
sven-urbanski-freiheit-com, line 107: I don't like that they're calling it tripper, but I guess we should be consistent, so keep as it is.

pkg/auth/dex_test.go
```diff
@@ -0,0 +1,132 @@
...
+	testCases := []struct {
+		Name           string
+		mockDexServer  *httptest.Server
+		wantStatusCode int
+	}{
+		{
+			Name: "Dex reverse proxy is working as expected on success",
+			mockDexServer: httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+				rw.WriteHeader(http.StatusOK)
+			})),
```
sven-urbanski-freiheit-com, line 71: nitpick:
I would prefer if this was extracted into a function, e.g.
`makeNewMockServer(http.StatusOK)`

bernardo-bastos: Yes, good idea. Added that method as a helper function.

sven-urbanski-freiheit-com (approved):

PR #786, opened 2023-07-05
Hide Buttons behind "..." menu in service lane

sven-urbanski-freiheit-com:
The buttons "show history", "Prepare Undeploy Release" and "Delete Forever" are now hidden in a menu, since they are rather uncommon.

This also allows us to add more buttons easily.

Story SRX-36LKZS

Screenshot from 2023-07-05 20-00-48

### DIFF
--- a/services/frontend-service/src/ui/components/ServiceLane/DotsMenu.tsx
+++ b/services/frontend-service/src/ui/components/ServiceLane/DotsMenu.tsx
@@ -1,0 +1,72 @@
+/*This file is part of kuberpult.
+
+Kuberpult is free software: you can redistribute it and/or modify
+it under the terms of the Expat(MIT) License as published by
+the Free Software Foundation.
+
+Kuberpult is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+MIT License for more details.
+
+You should have received a copy of the MIT License
+along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>.
+
+Copyright 2023 freiheit.com*/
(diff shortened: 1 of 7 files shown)


services/frontend-service/src/ui/components/ServiceLane/ServiceLane.tsx
```diff
@@ -22,13 +22,14 @@ import {
     useVersionsForApp,
 } from '../../utils/store';
 import { ReleaseCard } from '../ReleaseCard/ReleaseCard';
-import { Button } from '../button';
+// import { Button } from '../button';
```
leandro-freiheit, line 25: I think we should remove this

services/frontend-service/src/ui/components/ServiceLane/DotsMenu.tsx
```diff
@@ -0,0 +1,72 @@
...
+                <Button className="mdc-button--unelevated" label={'⋮'} onClick={myOpen} />
+            </div>
+        );
+    }
+
+    return (
+        <div className={'dots-menu'} onMouseLeave={myClose}>
+            <ul className={'list'}>
+                {props.buttons.map((button, index) => (
+                    <li className={'item'} key={'button-menu-' + String(index)}>
```
leandro-freiheit, line 59: Can we apply a rounding to the edges of the menu? It is not consistent with the header that is on the background.

sven-urbanski-freiheit-com: done

services/frontend-service/src/ui/components/ServiceLane/ServiceLane.tsx
sven-urbanski-freiheit-com: done

leandro-freiheit (approved):

PR #787, opened 2023-07-05
Bugfix: Display locks only for the correct  application

sebastianmueller:
In the previous implementation locks would also be displayed on apps that names are contained in a locked app's name.

### DIFF
--- a/services/frontend-service/src/ui/utils/store.tsx
+++ b/services/frontend-service/src/ui/utils/store.tsx
@@ -300,7 +300,7 @@ export const useFilteredApplicationLocks = (appNameParam: string | null): Displa
                     );
                 });
         });
-    const filteredLocks = finalLocks.filter((val) => searchCustomFilter(appNameParam, val.application));
+    const filteredLocks = finalLocks.filter((val) => appNameParam === val.application);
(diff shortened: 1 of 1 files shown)


services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -335,7 +335,7 @@ export const useEnvironmentLock = (lockId: string): DisplayLock => {
 
 export const searchCustomFilter = (queryContent: string | null, val: string | undefined): string => {
     if (!!val && !!queryContent) {
-        if (val.includes(queryContent)) {
+        if (val === queryContent) {
```
sven-urbanski-freiheit-com, line 338: please adapt the test in Home.test.tsx, it should fail now.

sven-urbanski-freiheit-com (approved):

sebastianmueller:
> we should keep the fuzzyness on the lockspage. Maybe split up in 2 functions, name one "fuzzy" the other "exact", and use the fuzzy one on the lockspage.

Changed `searchCustomFilter` back. Not sure though if a special function is needed for this case.

sven-urbanski-freiheit-com (approved):

PR #789, opened 2023-07-06
Remove outdated LockService

sven-urbanski-freiheit-com:
It was only accessible over grpc+web, but the client did not use it anymore. We switched to do everything regarding Locks with the BatchService


services/frontend-service/pkg/handler/handle_test.go
```diff
@@ -833,29 +864,29 @@ func (m *mockDeployClient) ReleaseTrain(_ context.Context, in *api.ReleaseTrainR
...
 }
 
-type mockLockClient struct {
-	createEnvironmentLockRequest            *api.CreateEnvironmentLockRequest
-	deleteEnvironmentLockRequest            *api.DeleteEnvironmentLockRequest
-	createEnvironmentApplicationLockRequest *api.CreateEnvironmentApplicationLockRequest
-	deleteEnvironmentApplicationLockRequest *api.DeleteEnvironmentApplicationLockRequest
+type mockBatchClient struct {
+	batchRequest *api.BatchRequest
+	//deleteEnvironmentLockRequest            *api.DeleteEnvironmentLockRequest
```
hannesg, line 869: Why just commented out? Git has history! Just remove it.

hannesg (approved):

services/frontend-service/pkg/handler/handle_test.go
sven-urbanski-freiheit-com: done

PR #796, opened 2023-07-07
Write source logs to stdout in transformer test

sven-urbanski-freiheit-com:
This adds a function to make it easier to get the logs of the *source* in the *test*.

Without this, only logs of the *test* will be printed.


pkg/logger/logger.go
```diff
@@ -43,6 +42,16 @@ func WithLogger(ctx context.Context, logger *zap.Logger) context.Context {
 	return ctxzap.ToContext(ctx, logger)
 }
 
+/*
+WrapPlain can be used to get logs in tests.
+*/
+func WrapPlain(ctx context.Context, inner func(ctx context.Context)) {
```
hannesg, line 48: This looks too simple to be useful. I think we should rather panic when there is no logger in context (sadly the zap context doesn't do that - have to roll our own) and should  provide an instrumentation testsetup that also contains traces.

PR #798, opened 2023-07-07
Disable autorebase when behind base branch option for renovate

ahmed-nour-fdc:
* DIsable autorebase for renovate.
* Change schedule to be daily.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -35,9 +35,7 @@
     "gomodUpdateImportPaths",
     "gomodTidy"
   ],
-  "rebaseWhen": "behind-base-branch",
   "schedule": [
-    "before 12pm on Tuesday",
-    "before 12pm on Friday"
+    "before 10am"
   ]
 }


hannesg:
Could you explain why we should not use autorebase?

sven-urbanski-freiheit-com:
> Could you explain why we should not use autorebase?

It's just very spammy and mostly not necessary.
If we merge 1 change, N builds would be retriggered.

ahmed-nour-fdc:
> Could you explain why we should not use autorebase?

Renovate should only run on a schedule to check wether there is a package that needs to be updated, but currently with `autorebase` it runs whenever a commit is merged which causes lots of emails to be sent for no real reason.
I think once per day should be sufficient.

hannesg (approved):

hannesg:
Makes sense. LGTM!

renovate.json
```diff
@@ -35,9 +35,7 @@
     "gomodUpdateImportPaths",
     "gomodTidy"
   ],
-  "rebaseWhen": "behind-base-branch",
```
sven-urbanski-freiheit-com, line 38: maybe we need to set this to "conflicted" excplicitely. Not sure about the default value

PR #805, opened 2023-07-12
Add a dummy user to the rollout service

hannesg:
The cd service requires a user in the grpc metadata. It can be literally anything but it must be there.

### DIFF
--- a/services/rollout-service/pkg/versions/versions.go
+++ b/services/rollout-service/pkg/versions/versions.go
@@ -21,8 +21,16 @@ import (
 	"fmt"
 
 	"github.com/freiheit-com/kuberpult/pkg/api"
+	"github.com/freiheit-com/kuberpult/pkg/auth"
 	"k8s.io/utils/lru"
 )
+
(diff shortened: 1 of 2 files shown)


leandro-freiheit (approved):

services/rollout-service/pkg/versions/versions.go
```diff
@@ -21,9 +21,17 @@ import (
 	"fmt"
 
 	"github.com/freiheit-com/kuberpult/pkg/api"
+	"github.com/freiheit-com/kuberpult/pkg/auth"
 	"k8s.io/utils/lru"
 )
 
+// This is a the user that the rollout service uses to query the versions.
+// It is not written to the repository.
+var RolloutServiceUser auth.User = auth.User{
```
sven-urbanski-freiheit-com, line 30: I think for now this is fine, because the GetOverview service doesn't make any changes in git.
Once we call an endpoint that actually makes a git commit (e.g. batchService), we need to make this name/email configurable like it is in the frontendService (or you call the frontendservice instead).

PR #807, opened 2023-07-14
Move functionality from DeploymentService and EnvironmentService to BatchService

hannesg:
In order to reduce the number of endpoints, I moved all the remaining writing grpc functions to the BatchEndpont.


services/cd-service/pkg/service/batch.go
```diff
@@ -166,21 +167,67 @@ func (d *BatchServer) processAction(
...
+		}, nil, nil
 	case *api.BatchAction_DeleteEnvFromApp:
 		act := action.DeleteEnvFromApp
 		return &repository.DeleteEnvFromApp{
 			Environment: act.Environment,
 			Application: act.Application,
-		}, nil
+		}, nil, nil
+	case *api.BatchAction_ReleaseTrain:
+		in := action.ReleaseTrain
```
sven-urbanski-freiheit-com, line 178: i think these checks should happen in the release train transformer.

pkg/ptr/ptr.go
```diff
@@ -26,3 +26,7 @@ func ToString(s *string) string {
 	}
 	return *s
 }
+
+func Bool(b bool) *bool {
```
sven-urbanski-freiheit-com, line 30: ohhh, that is much nicer than what I do usually in this case :+1:

services/frontend-service/pkg/handler/handle_test.go
```diff
@@ -274,24 +314,32 @@ func TestServer_Handle(t *testing.T) {
...
+							CreateEnvironment: &api.CreateEnvironmentRequest{
+								Environment: "stg",
+								Config: &api.EnvironmentConfig{
+									Upstream: &api.EnvironmentConfig_Upstream{
+										Latest: &latestFlag,
+									},
+									Argocd: &api.EnvironmentConfig_ArgoCD{
+										SyncWindows: nil,
+										Destination: &api.EnvironmentConfig_ArgoCD_Destination{
+											Server:    "https://intwawiplatform-dev-aks-a17d3d29.privatelink.westeurope.azmk8s.io:443",
```
sven-urbanski-freiheit-com, line 330: let's use example.com or something else here

hannesg: Uhja

services/cd-service/pkg/service/batch.go
hannesg: I just moved them around from here: https://github.com/freiheit-com/kuberpult/blob/ebfbe67f7382150b5c6070cd88aecfd7f4c5cbc5/services/cd-service/pkg/service/deploy.go#L88
I generally think it's fine to do such preflight check upfront.

services/frontend-service/package.json
```diff
@@ -1,95 +1,95 @@
 {
-  "name": "raw-react-app",
```
sven-urbanski-freiheit-com, line 2: can you do this change without reformatting?
I feel like this will revert when someone else installs/removes a package.

charts/kuberpult/run-kind.sh
```diff
@@ -294,8 +294,8 @@ kubectl get pods
 
 if "$LOCAL_EXECUTION"
 then
-  echo "sleeping for 1h to allow debugging"
-  sleep 1h
+  echo "sleeping to allow debugging. press any key to exit"
```
sven-urbanski-freiheit-com, line 297:
```suggestion
  echo "sleeping to allow debugging. press enter to exit"
```
pretty sure `read` only reacts on enter or ctrl-c

services/cd-service/pkg/service/batch.go
sven-urbanski-freiheit-com: fine for now.
I will look if we can find a general approach on where to do the validations.

sven-urbanski-freiheit-com (approved):

PR #814, opened 2023-07-21
Replaced /release http endpoint in cd-service with grpc

sven-urbanski-freiheit-com:
Replaced /release http endpoint in cd-service with grpc
    
Before, we had a release http endpoint in the cd-service and the frontend-service had it too, just redirecting to the cd-service.
Now we have a release http endpoint in the frontend-service, and it uses a new grpc endpoint to call the cd-service.
    
This is technically a breaking change, because you may have used the cd-service /release endpoint directly.
However, if you are using the helm chart, then this will not be breaking, because it exposes the right endpoint correctly.
    
Only effective change is that the auther parameters for the release endpoint are not optional.
If not provided, the ones in the helm chart will be used. However, this may change in the future, so we recommend to continue to provide the author parameters.
    
Also added some integration test, because this endpoint is now only testable with an integration test, as 2 microservices are involved.


pkg/api/api.proto
```diff
@@ -27,6 +27,7 @@ message BatchAction {
     DeleteEnvironmentFromAppRequest delete_env_from_app = 8;
     CreateEnvironmentRequest create_environment = 9;
     ReleaseTrainRequest release_train = 10;
+    CreateNewReleaseRequest create_new_release = 11;
```
hannesg, line 30: Consitency: `CreateReleaseRequest` just like `CreateEnvironmentRequest`

pkg/api/api.proto
```diff
@@ -64,6 +66,31 @@ message DeleteEnvironmentApplicationLockRequest {
...
+  optional uint64 version = 5;
+  string sourceCommitId = 6;
+  string sourceAuthor = 7;
+  string sourceMessage = 8;
+  optional string sourceRepoUrl=9;
+}
+
+message CreateNewReleaseResponse {
+  repeated string deployedToEnvironments = 1;
+  string application = 2;
```
hannesg, line 85: Is there a good reason we echo back the application name? I never seen anyone using this.

pkg/api/api.proto
```diff
@@ -64,6 +66,31 @@ message DeleteEnvironmentApplicationLockRequest {
...
 }
 
+
+message CreateNewReleaseRequest {
+  string environment = 1;
+  string application = 2;
+  optional string team = 3;
+  // env->yaml
+  map<string, string> manifests = 4;
+  optional uint64 version = 5;
```
hannesg, line 76: I would avoid the three optionals here. We are already using 0 for "no version supplied" further downstream and empty strings aren't valid values for team and soure repo anyway.

services/frontend-service/pkg/handler/release.go
```diff
@@ -17,35 +17,205 @@ Copyright 2023 freiheit.com*/
...
+
+	//func (p *Auth) ServeHTTPRelease(tail string, w http.ResponseWriter, r *http.Request) {
+	logger.FromContext(ctx).Warn(fmt.Sprintf("ServeHTTPRelease start: %s", r.URL.String()))
+	user, err := auth.ReadUserFromHttpHeader(ctx, r)
+	if err != nil {
+		user = &auth.User{
+			Email: "empty",
+			Name:  "empty",
+		}
+		logger.FromContext(ctx).Warn(fmt.Sprintf("release endpoint was called without user headers - set the http headers '%s' and '%s' to the desired git commit author and make sure they are base-64 encoded - %v", auth.HeaderUserName, auth.HeaderUserEmail, err))
```
hannesg, line 86: We should make this a hard-error soon.

pkg/api/api.proto
sven-urbanski-freiheit-com: I'm fine with that as long as we don't expose the grpc endpoint.

If we do expose it, I would like to have clear indications of what is optional and what is not.

pkg/api/api.proto
sven-urbanski-freiheit-com: For the app name, not really.
I would however, like to return *something*, e.g. where this release was deployed to immediately. This requires a bit of a refactoring, because the transformers don't return anything though.

pkg/api/api.proto
sven-urbanski-freiheit-com: done

services/frontend-service/pkg/handler/release.go
sven-urbanski-freiheit-com: yes, but probably not in this PR. We need to figure out how to handle technical users first.
For now, we default back to the defaultUser in the frontend-service.

pkg/api/api.proto
sven-urbanski-freiheit-com: For now, this is handled in server.go:ServeHTTP:~400

services/frontend-service/pkg/handler/release.go
sven-urbanski-freiheit-com: For now, this is handled in server.go:ServeHTTP:~400

pkg/api/api.proto
sven-urbanski-freiheit-com: I removed it for now.

pkg/api/api.proto
sven-urbanski-freiheit-com: Solved it by not exposing it as grpc in the frontend service

.github/workflows/integration-tests.yml
```diff
@@ -130,22 +131,23 @@ jobs:
...
+#    - name: Print container logs
+#      if: steps.integrationTests.outcome == 'failure'
+#      run: |
+#        echo "CD logs"
+#        docker logs kuberpult-cd-service
+#        echo "Frontend logs"
+#        docker logs kuberpult-frontend-service
+#
+#        # We need to exit with error code because "continue-on-errors" bypasses the error and the build is considered successful
+#        exit 1
```
hannesg, line 153: Good idea. I think we can remove this now entirely!

charts/kuberpult/templates/ingress.yaml
```diff
@@ -60,16 +60,10 @@ spec:
   rules:
   - host: {{ required ".ingress.domainName is required" .Values.ingress.domainName | quote }}
     http:
+  {{- if .Values.ingress.exposeReleaseEndpoint }}
+  {{ fail "Values.ingress.exposeReleaseEndpoint cannot be used anymore. Use the frontend service to call the /release endpoint"}}
```
hannesg, line 64: Good idea!

.github/workflows/integration-tests.yml
sven-urbanski-freiheit-com: done!

tests/integration-tests/integration-tests.test
hannesg, line 1: Should this file be here?

tests/integration-tests/CreateRelease_test.go
hannesg, line 1: The file name is very uncommon. Why not just `release_test.go`?

services/frontend-service/pkg/handler/release.go
```diff
@@ -17,35 +17,204 @@ Copyright 2023 freiheit.com*/
...
 		return
 	}
-
-	// Release api requires file uploads in it, which is hard to implement in grpc.
-	// Release api in cd-service only exists in REST right now, so this endpoint directly calls the REST endpoint of cd-service instead of a grpc one.
-	cdServiceProxy := httputil.NewSingleHostReverseProxy(url)
-	cdServiceProxy.ServeHTTP(w, req)
+	jsonBlob, err := json.Marshal(response.Results[0].GetCreateReleaseResponse())
+	if err != nil {
+		return
```
hannesg, line 217: We should handle this somehow better. Just returning an empty body is not right thing.

services/frontend-service/pkg/handler/release.go
```diff
@@ -17,35 +17,204 @@ Copyright 2023 freiheit.com*/
...
+	}
+	w.Write(jsonBlob)
```
hannesg, line 219: The http status code is subtly changed from the current behaviour. Kuberpult responds with 201 if the new release was created and 200 if the release already exists. Not sure if that is *actually* an issue because people should treat all 2xx status as success. Maybe at least write down the response status explicitly and write this in the changelog?

services/frontend-service/pkg/cmd/server.go
```diff
@@ -377,21 +381,24 @@ func (p *Auth) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 			source = "iap"
 		}
-		if u != nil {
-			span.SetTag("current-user-name", u.Name)
-			span.SetTag("current-user-email", u.Email)
+		if user == nil {
+			user, _ = auth.ReadUserFromHttpHeader(ctx, r)
+		}
+		if user != nil {
+			span.SetTag("current-user-name", user.Name)
```
hannesg, line 397: This doens't look related to your story. Maybe put this in an extra PR?

hannesg (approved):
You have some `Warn` logs in your code. Please remove them before merging.

services/frontend-service/pkg/cmd/server.go
sven-urbanski-freiheit-com: It is subtle, but it is related:

Before we only read the user from google/azure. Now we read it from the http header too - otherwise it would be always empty for the /release endpoint.

It is now an issue, because I want to pass along the author from an http endpoint (release) in the frontend-service.
Since we asked users to provide the header (here https://github.com/freiheit-com/kuberpult/blob/main/README.md#releasing-a-new-version), I would like to keep that behavior as it was. To achieve that we need to read the http header here.

services/frontend-service/pkg/handler/release.go
sven-urbanski-freiheit-com: good point, I will return an internal server error and log an error.

services/frontend-service/pkg/handler/release.go
sven-urbanski-freiheit-com: ohhhhhhhhhh, this is a really good catch. I broke it even more than that. Will fix it properly.

services/frontend-service/pkg/handler/release.go
sven-urbanski-freiheit-com: done

services/frontend-service/pkg/handler/release.go
sven-urbanski-freiheit-com: done, also added a test for that case

PR #815, opened 2023-07-24
Block Users without Permission from Creating Environment Locks

gsandok:
Takes the extracted role information from the GRPC context. 
When Dex is enabled then checks if associated role has permissions for createEnvironmentLock. 
Blocks user if their role does not have permissions.

### DIFF
--- a/pkg/auth/auth.go
+++ b/pkg/auth/auth.go
@@ -82,10 +82,31 @@ func WriteUserRoleToGrpcContext(ctx context.Context, userRole string) context.Co
 	return metadata.AppendToOutgoingContext(ctx, HeaderUserRole, Encode64(userRole))
 }
 
+type GrpcContextReader interface {
+	ReadUserFromGrpcContext(ctx context.Context, dexEnabled bool) (*User, error)
+}
+
+type DexGrpcContextReader struct {
+}
+
+type DummyGrpcContextReader struct {
+}
+
+func (x *DummyGrpcContextReader) ReadUserFromGrpcContext(ctx context.Context, dexEnabled bool) (*User, error) {
+	user := &User{
+		Email: "dummyMail@example.com",
+		Name:  "userName",
+		DexAuthContext: &DexAuthContext{
+			Role: "Developer",
+		},
+	}
+	return user, nil
+}
+
 // ReadUserFromGrpcContext should only be used in the cd-service.
 // ReadUserFromGrpcContext takes the User from middleware (context).
 // It returns a User or an error if the user is not found.
-func ReadUserFromGrpcContext(ctx context.Context, dexEnabled bool) (*User, error) {
+func (x *DexGrpcContextReader) ReadUserFromGrpcContext(ctx context.Context, dexEnabled bool) (*User, error) {
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
 		return nil, httperrors.AuthError(ctx, errors.New("could not retrieve metadata context with git author in grpc context"))
--- a/pkg/auth/rbac.go
+++ b/pkg/auth/rbac.go
@@ -27,8 +27,8 @@ import (
 type RBACConfig struct {
 	// Indicates if Dex is enabled.
 	DexEnabled bool
-	// The RBAC policy. In later stories mapping for policies will be integrated here
-	// Policy map[string]*Permission
+	// The RBAC policy. Key is for example "p,Developer,EnvironmentLock,Create,production,allow"
+	Policy map[string]*Permission
 }
 
 // Inits the RBAC Config struct
--- a/services/cd-service/pkg/cmd/server.go
+++ b/services/cd-service/pkg/cmd/server.go
@@ -61,6 +61,7 @@ type Config struct {
 	EnableMetrics     bool   `default:"false" split_words:"true"`
 	DogstatsdAddr     string `default:"127.0.0.1:8125" split_words:"true"`
 	EnableSqlite      bool   `default:"true" split_words:"true"`
+	DexMock           bool   `default:"false" split_words:"true"`
 }
 
 func (c *Config) readRbacPolicy() (policy map[string]*auth.Permission, err error) {
@@ -104,8 +105,18 @@ func RunServer() {
 			logger.FromContext(ctx).Fatal("config.parse.error", zap.Error(err))
 		}
 
+		var reader auth.GrpcContextReader
+		if c.DexMock {
+			if !c.DexEnabled {
+				logger.FromContext(ctx).Fatal("dexEnabled must be true if dexMock is true")
+			}
+			reader = &auth.DummyGrpcContextReader{}
+		} else {
+			reader = &auth.DexGrpcContextReader{}
+		}
+		dexRbacPolicy := map[string]*auth.Permission{}
 		if c.DexEnabled {
-			dexRbacPolicy, err := c.readRbacPolicy()
+			dexRbacPolicy, err = c.readRbacPolicy()
 			if err != nil {
 				logger.FromContext(ctx).Fatal("dex.read.error", zap.Error(err))
 			}
@@ -122,7 +133,7 @@ func RunServer() {
 			req interface{},
 			info *grpc.UnaryServerInfo,
 			handler grpc.UnaryHandler) (interface{}, error) {
-			return interceptors.UnaryUserContextInterceptor(ctx, req, info, handler, c.DexEnabled)
+			return interceptors.UnaryUserContextInterceptor(ctx, req, info, handler, c.DexEnabled, reader)
 		}
 
 		grpcStreamInterceptors := []grpc.StreamServerInterceptor{
@@ -212,6 +223,7 @@ func RunServer() {
 						Repository: repo,
 						RBACConfig: auth.RBACConfig{
 							DexEnabled: c.DexEnabled,
+							Policy:     dexRbacPolicy,
 						},
 					})
 
--- a/services/cd-service/pkg/service/batch.go
+++ b/services/cd-service/pkg/service/batch.go
@@ -91,12 +91,44 @@ func ValidateApplication(
 	return nil
 }
 
+func (d *BatchServer) checkUserPermissions(user *auth.User, env, application, action string) error {
+	if !d.RBACConfig.DexEnabled {
+		return nil
+	}
+
+	environmentString := env
+	if env != "*" {
+		groups, err := d.Repository.State().GetEnvironmentConfigs()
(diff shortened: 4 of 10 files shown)


services/cd-service/pkg/service/batch.go
```diff
@@ -93,11 +93,17 @@ func ValidateApplication(
 }
 
 func (d *BatchServer) processAction(
-	batchAction *api.BatchAction,
+	batchAction *api.BatchAction, user *auth.User,
 ) (repository.Transformer, *api.BatchResult, error) {
 	switch action := batchAction.Action.(type) {
 	case *api.BatchAction_CreateEnvironmentLock:
 		act := action.CreateEnvironmentLock
+		if d.RBACConfig.DexEnabled {
```
hannesg, line 101: I would avoid inlining this whole check and maybe just have an iterface that you call to authorize this call.

services/cd-service/pkg/service/batch.go
```diff
@@ -93,11 +93,17 @@ func ValidateApplication(
...
+			_, permissionsExist := d.RBACConfig.Policy[fmt.Sprintf("p,%s,EnvironmentLock,Create,%s,allow", user.DexAuthContext.Role, act.Environment)]
+			if !permissionsExist {
+				return nil, nil, fmt.Errorf("user does not have permissions to create an environment lock")
```
hannesg, line 104: We definitely a dedicated error type for permission errors.

hannesg (commented):
Please add some tests.

services/cd-service/pkg/service/batch_test.go
```diff
@@ -145,73 +203,73 @@ func TestBatchServiceWorks(t *testing.T) {
...
-				if exists {
-					t.Error("lock was not deleted")
-				}
-			}
-			// check that the applock was created/deleted
-			{
-				appLocks, err := svc.Repository.State().GetEnvironmentApplicationLocks("production", "test")
-				if err != nil {
-					t.Fatal(err)
+			if err == nil {
```
hannesg, line 221: You can avoid indenting here by just returning from that function in the line before if there is an error.

pkg/auth/rbac.go
```diff
@@ -27,8 +27,8 @@ import (
 type RBACConfig struct {
 	// Indicates if Dex is enabled.
 	DexEnabled bool
-	// The RBAC policy. In later stories mapping for policies will be integrated here
-	// Policy map[string]*Permission
+	// The RBAC policy.
```
sven-urbanski-freiheit-com, line 30:
```suggestion
	// The RBAC policy. Key is for example "p,Developer,EnvironmentLock,Create,production,allow"
```

services/cd-service/pkg/service/batch_test.go
```diff
@@ -26,10 +26,12 @@ import (
 
 	"github.com/freiheit-com/kuberpult/services/cd-service/pkg/testutil"
 	"github.com/google/go-cmp/cmp"
+	"google.golang.org/grpc/codes"
```
sven-urbanski-freiheit-com, line 29: overall this PR is fine, but it's really hard to test.
I discussed with @gsandok an approach to solve this.
Basically we will provide a way mock the authentication handler (dex), and overwrite with a role.
This way we can test this locally without having to setup dex.
(And my hidden agenda here is that this may be useful for technical users too).

pkg/auth/auth.go
```diff
@@ -82,10 +82,31 @@ func WriteUserRoleToGrpcContext(ctx context.Context, userRole string) context.Co
...
+
+type DexGrpcContextReader struct {
+}
+
+type DummyGrpcContextReader struct {
+}
+
+func (x *DummyGrpcContextReader) ReadUserFromGrpcContext(ctx context.Context, dexEnabled bool) (*User, error) {
+	user := &User{
+		Email: "userMail",
```
sven-urbanski-freiheit-com, line 97:
```suggestion
		Email: "dummyMail@example.com",
```

services/cd-service/pkg/cmd/server.go
```diff
@@ -125,8 +126,17 @@ func RunServer() {
 			logger.FromContext(ctx).Fatal("azure.auth.error: pgpKeyRing is required to authenticate manifests when \"KUBERPULT_AZURE_ENABLE_AUTH\" is true")
 		}
 
+		var reader auth.GrpcContextReader
+		if c.DexMock {
+			c.DexEnabled = true
```
sven-urbanski-freiheit-com, line 131: rather than changing a configuration silently, please return an error if they configs do not match:
```suggestion
		if c.DexMock {
		        if !c.DexEnabled {
		            return nil, errors.New("dexEnabled must be true if dexMock is true")
		        }
```

services/cd-service/pkg/cmd/server.go
```diff
@@ -125,8 +126,17 @@ func RunServer() {
...
+			reader = &auth.DummyGrpcContextReader{}
+		} else {
+			reader = &auth.DexGrpcContextReader{}
+		}
+		dexRbacPolicy := map[string]*auth.Permission{}
 		if c.DexEnabled {
-			dexRbacPolicy, err := c.readRbacPolicy()
+			logger.FromContext(ctx).Info("IT WORKED :)")
```
sven-urbanski-freiheit-com, line 138: remove logging please

services/cd-service/pkg/service/batch.go
```diff
@@ -92,12 +93,37 @@ func ValidateApplication(
 	return nil
 }
 
+func (d *BatchServer) checkUserPermissions(user *auth.User, act *api.CreateEnvironmentLockRequest, application, action string) error {
```
sven-urbanski-freiheit-com, line 96:
```suggestion
func (d *BatchServer) checkUserPermissions(user *auth.User, env string, application, action string) error {
```
You're only using the environment.

services/cd-service/pkg/service/batch.go
```diff
@@ -92,12 +93,37 @@ func ValidateApplication(
...
+		return err
+	}
+	group := act.Environment
+	temp := groups[act.Environment].EnvironmentGroup
+	if temp != nil {
+		group = *temp
+	}
+	_, permissionsExist := d.RBACConfig.Policy[fmt.Sprintf("p,%s,%s,%s,%s,allow", user.DexAuthContext.Role, application, action, fmt.Sprintf("%s:%s", act.Environment, group))]
+	if !permissionsExist {
+		return status.Errorf(codes.PermissionDenied, "user does not have permissions to create an environment lock")
```
sven-urbanski-freiheit-com, line 111: please return more details here, like which line you were looking for in the rbacconfig.Policy.

services/cd-service/pkg/service/batch_test.go
```diff
@@ -145,29 +205,30 @@ func TestBatchServiceWorks(t *testing.T) {
...
+			tc.svc.Repository = repo
+			resp, err := tc.svc.ProcessBatch(
+				tc.context,
 				&api.BatchRequest{
 					Actions: tc.Batch,
 				},
 			)
+			if err != nil && err.Error() == tc.expectedError {
+				return
+			}
```
sven-urbanski-freiheit-com, line 221: Please invert this and do not just "return"
```suggestion
			if err != nil && err.Error() != tc.expectedError {
				t.Fatal(err.Error())
			}
```

hannesg (commented):
I need to rethink this testing approach a bit. Maybe it's easier to not pass this reader around but instead have an undocumented flag on the auth package that enables mocking? At least, we definitely don't need to pass this reader around twice. The interceptor should put the verified user in the context and then it shouldn't matter how it got there.

services/cd-service/pkg/service/batch.go
sven-urbanski-freiheit-com, line 96: Ideally, this function should be part of the `GrpcContextReader`.
But we can do that in another PR.

services/cd-service/pkg/service/batch.go
```diff
@@ -92,12 +93,37 @@ func ValidateApplication(
...
+	groups, err := d.Repository.State().GetEnvironmentConfigs()
+	if err != nil {
+		return err
+	}
+	group := act.Environment
+	temp := groups[act.Environment].EnvironmentGroup
+	if temp != nil {
+		group = *temp
+	}
+	_, permissionsExist := d.RBACConfig.Policy[fmt.Sprintf("p,%s,%s,%s,%s,allow", user.DexAuthContext.Role, application, action, fmt.Sprintf("%s:%s", act.Environment, group))]
```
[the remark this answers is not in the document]
gsandok: for which aspect of this are you asking about the format?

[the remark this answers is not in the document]
gsandok: making adjustment know to handle case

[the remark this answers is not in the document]
hannesg: Bernardo

services/cd-service/pkg/service/batch.go
```diff
@@ -229,7 +255,7 @@ func (d *BatchServer) ProcessBatch(
 	ctx context.Context,
 	in *api.BatchRequest,
 ) (*api.BatchResponse, error) {
-	user, err := auth.ReadUserFromGrpcContext(ctx, d.RBACConfig.DexEnabled)
+	user, err := d.Reader.ReadUserFromGrpcContext(ctx, d.RBACConfig.DexEnabled)
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: Yes I think you're right.
The interceptor already does put it in the context. So we can here just read from context (not grpc-context) @gsandok

services/cd-service/pkg/service/batch.go
sven-urbanski-freiheit-com: we'll do this later

services/cd-service/pkg/service/batch.go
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: @hannesg What does that mean?
It works for now, so we'll keep it as it is.

services/cd-service/pkg/service/batch.go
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: done

sven-urbanski-freiheit-com:
> At least, we definitely don't need to pass this reader around twice.

We don't do that anymore.
For the other part of your comment: We'll address this in another PR.

hannesg (approved):

sven-urbanski-freiheit-com (approved):

PR #822, opened 2023-07-27
Allow argocd to render our manifests in parallel

hannesg:
From tracing, I saw that the rendering requests to the repo server take quiet long when you have > 2k apps. There is an option in the repo server to make this faster by avoiding the global lock. This lock is in place because some manifests renderer ( e.g. kustomize with parameter overrides ) create local files and therefore cannot run in parallel. By putting a special file in the apps root path, we can signal argocd to avoid this locking all together.

See here: https://argo-cd.readthedocs.io/en/stable/operator-manual/high_availability/#enable-concurrent-processing
And here: https://github.com/argoproj/argo-cd/blob/f1607fee7c3bfad8eb56a53130cf9ac7cc22dd34/reposerver/repository/repository.go#L1023

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -879,6 +879,11 @@ func (c *DeployApplicationVersion) Transform(ctx context.Context, state *State) 
 	if err := util.WriteFile(fs, manifestFilename, manifestContent, 0666); err != nil {
 		return "", err
 	}
+	// .argocd-allow-concurrency is a special file. It signals that rendering this manifests has no side-effects and avoids holding
+	// a global lock.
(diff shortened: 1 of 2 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -879,6 +879,11 @@ func (c *DeployApplicationVersion) Transform(ctx context.Context, state *State)
 	if err := util.WriteFile(fs, manifestFilename, manifestContent, 0666); err != nil {
 		return "", err
 	}
+	// .argocd-allow-concurrency is a special file. It signals that rendering this manifests has no side-effects and avoids holding
+	// a global lock.
```
sven-urbanski-freiheit-com, line 883: Do we want to do this always, or hide behind a feature toggle?

sven-urbanski-freiheit-com (approved):

PR #831, opened 2023-07-31
DeployApplicationVersion Role based Access Control 

bernardo-bastos:
Implements RBAC on DeployApplicationVersion. Only activated if DEX is enabled.

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -844,6 +844,7 @@ func (c *QueueApplicationVersion) Transform(ctx context.Context, state *State) (
 }
 
 type DeployApplicationVersion struct {
+	Authentication
 	Environment   string
 	Application   string
 	Version       uint64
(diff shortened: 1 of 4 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -844,13 +844,18 @@ func (c *QueueApplicationVersion) Transform(ctx context.Context, state *State) (
...
 type DeployApplicationVersion struct {
+	Authentication
 	Environment   string
 	Application   string
 	Version       uint64
 	LockBehaviour api.LockBehavior
 }
 
 func (c *DeployApplicationVersion) Transform(ctx context.Context, state *State) (string, error) {
+	err := state.checkUserPermissions(ctx, c.Environment, "*", auth.PermissionDeployRelease, c.RBACConfig)
```
sven-urbanski-freiheit-com, line 855:
```suggestion
	err := state.checkUserPermissions(ctx, c.Environment, c.Application, auth.PermissionDeployRelease, c.RBACConfig)
```

sven-urbanski-freiheit-com: I think we need the app here. Not everyone is allowed to deploy all apps.

sven-urbanski-freiheit-com (changes requested):
.

services/cd-service/pkg/repository/transformer.go
bernardo-bastos: Yes correct, I added the app check here.

sven-urbanski-freiheit-com (approved):

PR #849, opened 2023-08-03
CreateApplicationVersion Role Based Access Control

bernardo-bastos:
Adds Role Based Access Control to CreateApplicationVersion.

Rev: DSN-J1WO8X

### DIFF
--- a/services/cd-service/pkg/repository/transformer_test.go
+++ b/services/cd-service/pkg/repository/transformer_test.go
@@ -794,6 +794,41 @@ func TestRbacTransformerTest(t *testing.T) {
 		ExpectedError string
 	}{
 		{
+			Name: "able to create application version with permissions policy",
+			Transformers: []Transformer{
+				&CreateEnvironment{
+					Environment: "acceptance",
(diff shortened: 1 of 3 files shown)


services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -793,6 +793,41 @@ func TestRbacTransformerTest(t *testing.T) {
 		Transformers  []Transformer
 		ExpectedError string
 	}{
+		{
+			Name: "able to create application version with permissions policy",
+			Transformers: []Transformer{
+				&CreateEnvironment{
+					Environment: "acceptance",
+					Config:      config.EnvironmentConfig{Upstream: &config.EnvironmentConfigUpstream{Environment: envAcceptance, Latest: false}},
```
sven-urbanski-freiheit-com, line 801: you can't have the same env as upstream

bernardo-bastos: Fixed the test.

bernardo-bastos: Following our discussion I also added the Authentication to `DeployApplicationVersion` that was missing. Added a test for covering this case as well.

PR #854, opened 2023-08-04
Do not render undeployed argoApps

sven-urbanski-freiheit-com:
This avoids the error in argoCd `rpc error: code = Unknown desc = environments/<env>/applications/<app>/manifests: app path does not exist`

This is usually a problem when creating a new app, because generally all manifest are created in the source repo at once, including for production environments.
However, once submitted to kuberpult via /release endpoint, kuberpult rendered the ArgoCd manifests for **all** environments, even those that are not deployed yet (like production).

With this change, kuberpult will only render the apps in an environment, if it is deployed there.

Reference: SRX-638BGR

### DIFF
--- a/services/cd-service/pkg/repository/transformer_test.go
+++ b/services/cd-service/pkg/repository/transformer_test.go
@@ -395,6 +395,225 @@ func TestCreateUndeployApplicationVersionErrors(t *testing.T) {
 	}
 }
 
+func TestDeployOnSelectedEnvs(t *testing.T) {
+	type Expected struct {
+		Path     string
+		fileData *string
+	}
+	tcs := []struct {
+		Name         string
+		Transformers []Transformer
+		Expected     []Expected
+	}{
+		{
+			Name: "generates multiple manifests",
+			Transformers: []Transformer{
+				&CreateEnvironment{
+					Environment: envAcceptance,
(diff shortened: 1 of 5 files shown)


services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -770,12 +989,12 @@ func TestReleaseTrainErrors(t *testing.T) {
 				}
 				actualMsg := commitMsg[len(commitMsg)-1]
 				if actualMsg != tc.expectedCommitMsg {
-					t.Fatalf("expected a different message.\nExpected: %q\nGot %q", tc.expectedCommitMsg, actualMsg)
+					t.Fatalf("Expected a different message.\nExpected: %q\nGot %q", tc.expectedCommitMsg, actualMsg)
```
leandro-freiheit, line 992: this should be lowercase

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -395,6 +395,225 @@ func TestCreateUndeployApplicationVersionErrors(t *testing.T) {
...
+  - '*'
+`),
+				},
+			},
+		},
+	}
+	for _, tc := range tcs {
+		tc := tc
+		t.Run(tc.Name, func(t *testing.T) {
+			//t.Parallel()
```
leandro-freiheit, line 588: This can be removed

leandro-freiheit (approved):

PR #871, opened 2023-08-14
WIP Call ArgoCd app refresh

sven-urbanski-freiheit-com:
This feature is off by default.
It allows kuberpult to call argoCd directly to inform it that an app has changed. This is a performance improvement.
See values.yaml.tpl for further explanations.

### DIFF
--- a/services/cd-service/Dockerfile
+++ b/services/cd-service/Dockerfile
@@ -1,6 +1,7 @@
 FROM alpine:3.18
 LABEL org.opencontainers.image.source https://github.com/freiheit-com/kuberpult
 RUN apk --update add ca-certificates tzdata libgit2 git sqlite-libs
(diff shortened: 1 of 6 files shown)


services/cd-service/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM alpine:3.18
 LABEL org.opencontainers.image.source https://github.com/freiheit-com/kuberpult
 RUN apk --update add ca-certificates tzdata libgit2 git sqlite-libs
+RUN wget https://github.com/argoproj/argo-cd/releases/download/v2.1.2/argocd-linux-amd64 -O /usr/local/bin/argocd && chmod +x /usr/local/bin/argocd
```
ahmed-nour-fdc, line 4: Why not use the latest version 2.8.0?

ahmed-nour-fdc (approved):

services/cd-service/Dockerfile
sven-urbanski-freiheit-com: good point, will do

hannesg:
Why not integrate this with the rollout service? That service already has all the config and it would be much more orthogonal.

PR #872, opened 2023-08-17
Tell ArgoCd to sync apps

sven-urbanski-freiheit-com:
With this feature, kuberpult sends webhooks that imitate github webhooks to argoCd, which trigger argoCds refreshs.

This is intended to reduce load on ArgoCd. Normally, argoCd renders all manifests again whenever there is a new commit in the manifest repository. However, with Webhooks, it only checks the actually changed files that are part of the Webhook body.

Note that this does not work for all cases, like deletion of environments from apps.

So you can make the "timeout.reconciliation" in argoCd bigger (e.g. 30min instead of the default 3min) - however you should not fully disable it. See https://argo-cd.readthedocs.io/en/stable/operator-manual/high_availability/#argocd-application-controller

To enable this kuberpult feature, set `argocd.server` in the helm chart to a hostname that reaches argoCds "argocd-server", usually this should be `http://argocd-server.${ARGO_NAMESPACE}.svc.cluster.local`.

If the certificate for the https connection is not accepted, set `argocd.insecure: true`.
This is usually safe, under the assumption that kuberpult and argoCd run on the same cluster anyway.

This feature is intended to be used when Webhooks are not an option for you (for example due to firewall policies).

Caveat: Kuberpult only retries failed connections a very limited number of times. That means, if argoCd does not react temporarily, webhooks may get lost, and therefore argoCd does not sync until the reconciliation timeout.

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -538,7 +538,14 @@ func (u *DeleteEnvFromApp) Transform(ctx context.Context, state *State) (string,
 		return "", wrapFileError(err, envAppDir, thisSprintf("Cannot delete app.'")), nil
 	}
 
-	changes := &TransformerResult{} // TODO implement
+	changes := &TransformerResult{
+		ChangedApps: []AppEnv{
+			{
+				App: u.Application,
+				Env: u.Environment,
+			},
+		},
+	}
 	return fmt.Sprintf("Environment '%v' was removed from application '%v' successfully.", u.Environment, u.Application), nil, changes
 }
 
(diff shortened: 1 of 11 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -771,29 +794,30 @@ type CreateEnvironmentApplicationLock struct {
...
-			return "", err
+			return "", nil, err
 		} else {
 			if err := createLock(ctx, chroot, c.LockId, c.Message); err != nil {
-				return "", err
+				return "", nil, err
 			} else {
 				GaugeEnvAppLockMetric(fs, c.Environment, c.Application)
-				return fmt.Sprintf("Created lock %q on environment %q for application %q", c.LockId, c.Environment, c.Application), nil
+				changes := &TransformerResult{} // TODO implement
```
ahmed-nour-fdc, line 819: Do we need this TODO here?

ahmed-nour-fdc (approved):

services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: done

PR #877, opened 2023-08-21
Increase nginx timeout to 300s

hannesg:
The publish endpoint takes a minute to reply if pushing to git takes a minute. This is not an error but nginx times out the reply after thirty seconds so that the caller has the impression that something went wrong. The timeout should be much higher. We will start with 5 minutes.

### DIFF
--- a/charts/kuberpult/values.yaml.tpl
+++ b/charts/kuberpult/values.yaml.tpl
@@ -64,7 +64,8 @@ rollout:
 
 ingress:
   create: true
-  annotations: {}
+  annotations:
+    nginx.ingress.kubernetes.io/proxy-read-timeout: 300
   domainName: null
   iap:
     enabled: false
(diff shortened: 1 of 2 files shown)


leandro-freiheit (approved):

charts/kuberpult/values.yaml.tpl
```diff
@@ -64,7 +64,8 @@ rollout:
 
 ingress:
   create: true
-  annotations: {}
+  annotations:
+    nginx.ingress.kubernetes.io/proxy-read-timeout: 300
```
sven-urbanski-freiheit-com, line 68: should this be a helm chart parameter?

hannesg: Good question. I think this the least bad way to do this. We could write it directly into the ingress template, but then it would conflict if somebody sets a value for this annotation directly. We could also merge the customer supplied annotations and our default annotations in the template and then render that new value. This would however be less obvious because  we would have more logic in the templates. I think this way has least frictions. It's completely obivious where the value comes from and how to change it.

PR #879, opened 2023-08-21
Fix commit Ids for webhooks to ArgoCd

sven-urbanski-freiheit-com:


services/cd-service/pkg/repository/repository.go
```diff
@@ -750,19 +743,31 @@ func (r *repository) ApplyTransformers(ctx context.Context, transformers ...Tran
 	var rev *git.Oid
 	if state.Commit != nil {
 		rev = state.Commit.Id()
+	} else {
+		// first commit in the repo!
```
leandro-freiheit, line 747: The comment is relevant but I don't like having an empty else statement

leandro-freiheit (approved):

services/cd-service/pkg/repository/repository.go
hannesg: +1

services/cd-service/pkg/repository/repository.go
```diff
@@ -676,6 +660,12 @@ type RootApp struct {
 type TransformerResult struct {
 	ChangedApps     []AppEnv
 	DeletedRootApps []RootApp
+	Commits         *Commits
+}
+
+type Commits struct {
+	Previous string
```
hannesg, line 667: I would find it more obvious if we would stick with *git.Oid here. Is there a specific reasons why you use string? If yes, maybe document it.

hannesg (approved):
How about some tests?

services/cd-service/pkg/repository/repository.go

services/cd-service/pkg/repository/repository.go

PR #890, opened 2023-08-23
Automatic release notes generation with labels

sven-urbanski-freiheit-com:


infrastructure/scripts/generate-changelog.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set eu -pipefail
+
+time docker run -it --rm -v "$(pwd)":/usr/local/src/your-app githubchangeloggenerator/github-changelog-generator \
```
hannesg, line 4: Do we need `time`? 

Also please use a fixed image version instead of latest.

hannesg (approved):

infrastructure/scripts/generate-changelog.sh
sven-urbanski-freiheit-com: done

PR #895, opened 2023-08-28
Fix argocd server processing and set plaintext option correctly

hannesg:
The argocd client doesn't allow passing the protocol anymore but demands that the plaintext field is set. This MR adds handling for urls in the kuberpult rollout service and also enforces the correct layout in the helm chart (including port!).


services/rollout-service/pkg/cmd/server.go
```diff
@@ -109,7 +110,13 @@ func runServer(ctx context.Context, config Config) error {
...
+	opts.ConfigPath = ""
+	u, err := url.ParseRequestURI(config.ArgocdServer)
+	if err != nil {
+		return fmt.Errorf("invalid argocd server url: %w", err)
+	}
+	opts.ServerAddr = u.Host
+	opts.PlainText = u.Scheme == "http"
 	opts.UserAgent = "kuberpult"
 	opts.Insecure = config.ArgocdInsecure
 	opts.AuthToken = config.ArgocdToken
```
sven-urbanski-freiheit-com, line 122: maybe extrac this as a function to make it better testable

hannesg: Done!

sven-urbanski-freiheit-com (approved):

PR #902, opened 2023-08-31
Allow locking and unlocking env groups over rest

sven-urbanski-freiheit-com:
This makes it's possible to lock/unlock all environments in one group with one rest call.
The endpoint is `/environment-groups/<envGroup>/locks/<lockId>`, see `infrastructure/scripts/create-testdata/create-env-group-lock.sh` and `infrastructure/scripts/create-testdata/delete-env-group-lock.sh` for details.
Note that this will create multiple *environment locks*.
That means if you create an env lock with id "x" you can then delete it for one environment by calling `/environments/<env>/locks/x" - even though it was created with the `environment-groups` endpoint.

So there are still only 2 kinds of locks: environment locks and app locks.

### DIFF
--- a/pkg/auth/azure.go
+++ b/pkg/auth/azure.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/freiheit-com/kuberpult/pkg/logger"
 	xpath "github.com/freiheit-com/kuberpult/pkg/path"
 	"log"
 	"net/http"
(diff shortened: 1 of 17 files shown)


pkg/auth/azure.go
```diff
@@ -101,11 +102,13 @@ func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfu
 	token := req.Header.Get("authorization")
 	for _, allowedPath := range allowedPaths {
 		if req.URL.Path == allowedPath {
+			logger.FromContext(req.Context()).Info("path not allowed: " + req.URL.Path)
```
hannesg, line 105: The log message is the wrong way round isn't it?

hannesg (approved):
Looks great! Only nit: the rest endpoint in the frontend doesn't yet have a unit test.

pkg/auth/azure.go
sven-urbanski-freiheit-com: yes, removed

PR #905, opened 2023-09-01
Don't show the rollout status if it's disabled

hannesg:
The rollout status is currently always shown, regardles if its backend is deployed. This change makes the frontend aware of the services availability, displaying the rollout status only if the frontend service doesn't responds with "not implemented" error.

### DIFF
--- a/services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.tsx
+++ b/services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.tsx
@@ -74,8 +74,12 @@ const rolloutStatusPriority = [
 const calculateDeploymentStatus = (
     app: string,
     deployedAt: EnvironmentGroupExtended[],
+    rolloutEnabled: boolean,
     rolloutStatus: RolloutStatusApplication
 ): [Array<StreamStatusResponse>, StreamStatusResponse?] => {
+    if (!rolloutEnabled) {
+        return [[], undefined];
+    }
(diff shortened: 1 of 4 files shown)


services/frontend-service/src/ui/components/ReleaseCard/ReleaseCard.tsx
```diff
@@ -74,8 +74,12 @@ const rolloutStatusPriority = [
 const calculateDeploymentStatus = (
     app: string,
     deployedAt: EnvironmentGroupExtended[],
+    rolloutEnabled: boolean,
     rolloutStatus: RolloutStatusApplication
 ): [Array<StreamStatusResponse>, StreamStatusResponse?] => {
+    if (!rolloutEnabled) {
+        return [[], undefined];
```
sven-urbanski-freiheit-com, line 81: Ideally we could show a tooltip here that explains the situation. e.g.
`<div title='To show the rollout status, enable the rollout service (...)'>not configured</div>`
(somewhere in the releaseCard)

sven-urbanski-freiheit-com: but I'm fine to merge as it is

hannesg: I'm not sure if that's a good idea to put this very prominently. Maybe have an info pane somewhere that shows this along with the versions of the backend services?

PR #906, opened 2023-09-01
Add Display Version to create release request

gsandok:
Testing changes


pkg/api/api.proto
```diff
@@ -78,6 +78,7 @@ message CreateReleaseRequest {
   string sourceAuthor = 7;
   string sourceMessage = 8;
   string sourceRepoUrl = 9;
+  string displayVersion = 10;
```
sven-urbanski-freiheit-com, line 81: this must be optional

hannesg: yes, but I would keep it this way. the usual convention of treating an empty string as absent is fine for us.

infrastructure/scripts/create-testdata/create-release.sh
```diff
@@ -1,15 +1,15 @@
...
+set -x
 
 # usage
 # ./create-release.sh my-service-name [my-team-name]
 # Note that this just creates files, it doesn't push in git
 
 name=${1}
 applicationOwnerTeam=${2:-sreteam}
-commit_id=$(LC_CTYPE=C tr -dc a-f0-9 </dev/urandom | head -c 12 ; echo '')
+commit_id=$(echo $RANDOM)
```
hannesg, line 12: Can you make this look more look more like a git commit id? 12 hex characters are much better than just a number.

hannesg (commented):
Looks good. Please cleanup create-release.sh

infrastructure/scripts/create-testdata/create-release.sh
gsandok: this was just for testing with sven, removing this cahnge

pkg/api/api.proto
sven-urbanski-freiheit-com: as long as we don't expose this in a sort of "public api", I don't mind either way.

E.g. once we generate a documentation/swagger/something out of this, it may be relevant to set it to optional. Since we don't have that, we can keep it non-optional.

sven-urbanski-freiheit-com:
maybe make it a Draft-PR while in its early stages

services/frontend-service/pkg/handler/release.go
```diff
@@ -194,6 +194,17 @@ func (s Server) HandleRelease(w http.ResponseWriter, r *http.Request, tail strin
 		}
 	}
 
+	if displayVersion, ok := form.Value["displayVersion"]; ok {
+		if len(displayVersion) == 1 {
+			if len(displayVersion[0]) > 15 {
+				w.WriteHeader(400)
+				fmt.Fprintf(w, "DisplayVersion given should be <= 15")
```
sven-urbanski-freiheit-com, line 201:
```suggestion
				fmt.Fprintf(w, "DisplayVersion given should be <= 15 characters")
```

services/frontend-service/pkg/handler/release.go
```diff
@@ -194,6 +194,17 @@ func (s Server) HandleRelease(w http.ResponseWriter, r *http.Request, tail strin
 		}
 	}
 
+	if displayVersion, ok := form.Value["displayVersion"]; ok {
+		if len(displayVersion) == 1 {
```
sven-urbanski-freiheit-com, line 198:
```suggestion
		if len(displayVersion) != 1 {
		        w.WriteHeader(400)
			fmt.Fprintf(w, fmt.Sprintf("Invalid number of display versions provided: %d, ", len(displayVersion)))
		}
```

sven-urbanski-freiheit-com (approved):

pkg/api/api.proto
```diff
@@ -294,6 +295,7 @@ message Environment {
     bool undeployVersion = 6;
     ArgoCD argoCD = 7;
     DeploymentMetaData deploymentMetaData = 8;
+    string displayVersion = 9;
```
hannesg, line 298: Shouldn't this field be part of the `Release` message?

hannesg (changes requested):

pkg/api/api.proto
sven-urbanski-freiheit-com: yes, good point

hannesg (approved):

PR #910, opened 2023-09-04
Notify argocd about updates

hannesg:
This change adds sending refresh notifications to argocd when applications in kuberpult changes. Notifications are send by the rollout service so that they do not slow down the cd service in any way. The current implementation is rather simple at the moment. The rollout service now listens to the cd service for changes, compares the versions with argocd and issues refresh requests if the version differs.

The feature must be explicitly enabled in the helm chart using `argocd.refreshEnabled`

### DIFF
--- a/charts/kuberpult/run-kind.sh
+++ b/charts/kuberpult/run-kind.sh
@@ -298,6 +298,7 @@ argocd:
   token: "$token"
   server: "https://argocd-server.${ARGO_NAMESPACE}.svc.cluster.local:443"
   insecure: true
+  refreshEnabled: true
 pgp:
   keyRing: |
 $(sed -e "s/^/    /" <./kuberpult-keyring.gpg)
--- a/services/rollout-service/pkg/cmd/server.go
+++ b/services/rollout-service/pkg/cmd/server.go
@@ -29,6 +29,7 @@ import (
 	"github.com/freiheit-com/kuberpult/pkg/logger"
 	"github.com/freiheit-com/kuberpult/pkg/setup"
 	"github.com/freiheit-com/kuberpult/pkg/tracing"
+	"github.com/freiheit-com/kuberpult/services/rollout-service/pkg/notifier"
 	"github.com/freiheit-com/kuberpult/services/rollout-service/pkg/service"
 	"github.com/freiheit-com/kuberpult/services/rollout-service/pkg/versions"
 	"github.com/kelseyhightower/envconfig"
@@ -47,9 +48,10 @@ type Config struct {
 	CdServer      string `default:"kuberpult-cd-service:8443"`
 	EnableTracing bool   `default:"false" split_words:"true"`
 
-	ArgocdServer   string `split_words:"true"`
-	ArgocdInsecure bool   `default:"false" split_words:"true"`
-	ArgocdToken    string `split_words:"true"`
+	ArgocdServer         string `split_words:"true"`
+	ArgocdInsecure       bool   `default:"false" split_words:"true"`
+	ArgocdToken          string `split_words:"true"`
+	ArgocdRefreshEnabled bool   `split_words:"true"`
 }
 
 func (config *Config) ClientConfig() (apiclient.ClientOptions, error) {
@@ -157,6 +159,33 @@ func runServer(ctx context.Context, config Config) error {
 	broadcast := service.New()
 	shutdownCh := make(chan struct{})
 	ready := false
+	versionC := versions.New(overview)
+
+	backgroundTasks := []setup.BackgroundTaskConfig{
+		{
+			Name: "consume argocd events",
+			Run: func(ctx context.Context) error {
+				return service.ConsumeEvents(ctx, appClient, versionC, broadcast, func() { ready = true })
+			},
+		},
+		{
+			Name: "consume kuberpult events",
+			Run: func(ctx context.Context) error {
+				return versionC.Subscribe(ctx, broadcast)
+			},
+		},
+	}
+
+	if config.ArgocdRefreshEnabled {
+		notify := notifier.New(appClient)
+		backgroundTasks = append(backgroundTasks, setup.BackgroundTaskConfig{
+			Name: "refresh argocd",
+			Run: func(ctx context.Context) error {
+				return notifier.Subscribe(ctx, notify, broadcast)
+			},
+		})
+	}
+
 	setup.Run(ctx, setup.ServerConfig{
 		HTTP: []setup.HTTPConfig{
 			{
@@ -172,13 +201,7 @@ func runServer(ctx context.Context, config Config) error {
 				},
 			},
 		},
-		Background: []setup.BackgroundTaskConfig{
-			{
-				Name: "consume events",
-				Run: func(ctx context.Context) error {
-					return service.ConsumeEvents(ctx, appClient, versions.New(overview), broadcast, func() { ready = true })
-				},
-			}},
+		Background: backgroundTasks,
 		GRPC: &setup.GRPCConfig{
 			Port: "8443",
 			Opts: []grpc.ServerOption{
--- a/services/rollout-service/pkg/cmd/server.go
+++ b/services/rollout-service/pkg/cmd/server.go
@@ -171,7 +171,7 @@ func runServer(ctx context.Context, config Config) error {
 		{
 			Name: "consume kuberpult events",
 			Run: func(ctx context.Context) error {
-				return versionC.Subscribe(ctx, broadcast)
+				return versionC.ConsumeEvents(ctx, broadcast)
 			},
 		},
 	}
--- a/services/rollout-service/pkg/notifier/notifier.go
+++ b/services/rollout-service/pkg/notifier/notifier.go
@@ -1,0 +1,66 @@
+/*This file is part of kuberpult.
+
+Kuberpult is free software: you can redistribute it and/or modify
+it under the terms of the Expat(MIT) License as published by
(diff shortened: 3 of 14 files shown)


services/rollout-service/pkg/notifier/subscriber.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (s *subscriber) maybeSend(ctx context.Context, ev *service.BroadcastEvent) {
+	// skip cases where we don't know the kuberpult version
+	if ev.KuberpultVersion == 0 {
+		return
+	}
+	// also don't notify when the version in argocd is already the right one
+	if ev.ArgocdVersion == ev.KuberpultVersion {
+	  return
```
sven-urbanski-freiheit-com, line 63: please check formatting

services/rollout-service/pkg/versions/versions_test.go
```diff
@@ -155,20 +202,100 @@ func TestVersionClient(t *testing.T) {
...
+			if err != nil {
+				t.Errorf("expected no error, but received %q", err)
 			}
+			mc.testAllConsumed(t)
+			testExpectedVersions(t, tc.ExpectedVersions, vc, mc)
 		})
 	}
+}
 
+func testExpectedVersions(t *testing.T, expectedVersions []expectedVersion, vc VersionClient, mc *mockOverviewClient) {
```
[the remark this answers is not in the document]
hannesg: Good idea!

sven-urbanski-freiheit-com:
type in the commit message: "an issues" -> "and issues"

services/rollout-service/pkg/notifier/notifier.go
```diff
@@ -0,0 +1,66 @@
...
+
+func New(client SimplifiedApplicationInterface) Notifier {
+	return &notifier{client}
+}
+
+type notifier struct {
+	client SimplifiedApplicationInterface
+}
+
+func (n *notifier) NotifyArgoCd(ctx context.Context, environment, application string) {
```
sven-urbanski-freiheit-com, line 49:
```suggestion

// NotifyArgoCd tells Argo CD to refresh one app in the given environment
func (n *notifier) NotifyArgoCd(ctx context.Context, environment, application string) {
```

services/rollout-service/pkg/versions/versions.go
```diff
@@ -72,6 +77,67 @@ func (v *versionClient) GetVersion(ctx context.Context, revision, environment, a
...
+}
+
+func (v *versionClient) Subscribe(ctx context.Context, processor VersionEventProcessor) error {
+	ctx = auth.WriteUserToGrpcContext(ctx, RolloutServiceUser)
+outer:
+	for {
+		client, err := v.client.StreamOverview(ctx, &api.GetOverviewRequest{})
+		if err != nil {
+			logger.FromContext(ctx).Warn("overview.connect", zap.Error(err))
+			continue outer
```
sven-urbanski-freiheit-com, line 97: should we really retry immediately? E.g. if we can't connect for a long time, shouldn't we log as error, not warn?

services/rollout-service/pkg/versions/versions.go
```diff
@@ -72,6 +77,67 @@ func (v *versionClient) GetVersion(ctx context.Context, revision, environment, a
...
+				if grpcErr != nil {
+					if grpcErr.Code() == codes.Canceled {
+						return nil
+					}
+
+					logger.FromContext(ctx).Warn("overview.stream", zap.Error(err), zap.String("grpc.code", grpcErr.Code().String()), zap.String("grpc.message", grpcErr.Message()))
+				} else {
+					logger.FromContext(ctx).Warn("overview.stream", zap.Error(err))
+				}
+				continue inner
```
sven-urbanski-freiheit-com, line 118: wouldn't just `continue` do the same thing here?

services/rollout-service/pkg/service/broadcast.go
```diff
@@ -33,38 +34,101 @@ type key struct {
...
+		Application:      application,
+		ArgocdVersion:    a.argocdVersion,
+		RolloutStatus:    rs,
+		KuberpultVersion: a.kuberpultVersion,
+	}
+}
+
 type Broadcast struct {
-	state    map[key]Event
+	state    map[key]*appState
```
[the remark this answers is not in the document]
hannesg: We still do need to lock things because we also have to consider the subscribers to that state AND we also need to merge the state of argocd and kuberpult. So even if we could have two seperate stores that would not require locking we would still need to have a common state that needs locking.

services/rollout-service/pkg/cmd/server.go
```diff
@@ -157,6 +159,33 @@ func runServer(ctx context.Context, config Config) error {
...
+		{
+			Name: "consume argocd events",
+			Run: func(ctx context.Context) error {
+				return service.ConsumeEvents(ctx, appClient, versionC, broadcast, func() { ready = true })
+			},
+		},
+		{
+			Name: "consume kuberpult events",
+			Run: func(ctx context.Context) error {
+				return versionC.Subscribe(ctx, broadcast)
```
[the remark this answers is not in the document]
hannesg: That is a good question. Perhaps yes. Though, the other ready flag is much more valuable as it ensures that the rollout service doesn't come up when the argocd config is wrong. For the connection to the cd service there is simply a lot less that can go wrong.

services/rollout-service/pkg/versions/versions.go
```diff
@@ -72,6 +77,67 @@ func (v *versionClient) GetVersion(ctx context.Context, revision, environment, a
...
+	Version     uint64
+}
+
+type VersionEventProcessor interface {
+	ProcessKuberpultEvent(ctx context.Context, ev KuberpultEvent)
+}
+
+func (v *versionClient) Subscribe(ctx context.Context, processor VersionEventProcessor) error {
+	ctx = auth.WriteUserToGrpcContext(ctx, RolloutServiceUser)
+outer:
```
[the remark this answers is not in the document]
hannesg: Okay, I will give it a try

services/rollout-service/pkg/cmd/server.go
sven-urbanski-freiheit-com, line 174: can we rename `Subscribe` to `ConsumeEvents` or vice versa?

services/rollout-service/pkg/cmd/server.go
```diff
@@ -157,6 +159,33 @@ func runServer(ctx context.Context, config Config) error {
 	broadcast := service.New()
 	shutdownCh := make(chan struct{})
 	ready := false
+	versionC := versions.New(overview)
```
sven-urbanski-freiheit-com, line 162: What happens if the overview service is down for a minute?
Do we then lose updates from Argo Cd?

charts/kuberpult/run-kind.sh
```diff
@@ -298,6 +298,7 @@ argocd:
   token: "$token"
   server: "https://argocd-server.${ARGO_NAMESPACE}.svc.cluster.local:443"
   insecure: true
+  refreshEnabled: true
```
sven-urbanski-freiheit-com, line 301: https://github.com/freiheit-com/kuberpult/blob/8cb4e6e1d6b08f57329a343279b52f3576fddd29/services/frontend-service/pkg/cmd/server.go#L465

"Unimplemented" is not accurate, right? It's just not enabled.
Maybe "424"? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/424

sven-urbanski-freiheit-com: or maybe just change the message to 
"rollout status not enabled"

services/rollout-service/pkg/service/broadcast.go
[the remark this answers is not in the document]
hannesg: Follow up: https://github.com/freiheit-com/kuberpult/issues/915

services/rollout-service/pkg/versions/versions.go
hannesg: No

services/rollout-service/pkg/cmd/server.go
sven-urbanski-freiheit-com: Discussed, it's true but we can live with this.
The rollout-service does on restart send everything necessary TO argoCd, though.

services/rollout-service/pkg/versions/versions.go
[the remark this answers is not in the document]
hannesg: Done

services/rollout-service/pkg/cmd/server.go
hannesg: Done

services/rollout-service/pkg/versions/versions.go
hannesg: Yes. Done as well.

services/rollout-service/pkg/cmd/server.go
hannesg: Yes we do but we immediately fully recover once it's back.

charts/kuberpult/run-kind.sh
hannesg: I created a follow up issue in github.

services/rollout-service/pkg/versions/versions.go
sven-urbanski-freiheit-com: can be solved in a later PR

sven-urbanski-freiheit-com (approved):

PR #917, opened 2023-09-04
Added spinner for submitting planned actions

sven-urbanski-freiheit-com:
This adds a Spinner when submitting Actions:
Screenshot from 2023-09-05 10-11-48

### DIFF
--- a/services/frontend-service/src/ui/components/Spinner/Spinner.scss
+++ b/services/frontend-service/src/ui/components/Spinner/Spinner.scss
@@ -24,6 +24,15 @@ Copyright 2023 freiheit.com*/
     .spinner-animation {
     }
     .spinner-message {
+        @extend .sub-headline1;
+
         margin-top: 20px;
+        background-color: white;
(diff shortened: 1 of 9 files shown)


services/frontend-service/src/ui/components/Spinner/Spinner.scss
```diff
@@ -25,5 +25,13 @@ Copyright 2023 freiheit.com*/
     }
     .spinner-message {
         margin-top: 20px;
+        font-size: 18px;
```
jaypontes, line 28: I understand we don't follow the figma design anymore but it would be nice to still use the global typography definitions for a cleaner look

jaypontes (approved):

services/frontend-service/src/ui/components/Spinner/Spinner.scss
sven-urbanski-freiheit-com: done

PR #923, opened 2023-09-05
Rollout service: dedup version notifications and notify for app deletions

hannesg:
Currently the rollout service needs to obtain the shared lock for every deployed app. This is rather costly. This patch changes that, so that we do send fewer events to that store by deduplicating the notifications in the versions.go locally.

Furthermore, the versions client never cared about removed apps.  This is also handled properly now.

### DIFF
--- a/services/rollout-service/pkg/versions/versions.go
+++ b/services/rollout-service/pkg/versions/versions.go
@@ -87,6 +87,11 @@ type VersionEventProcessor interface {
 	ProcessKuberpultEvent(ctx context.Context, ev KuberpultEvent)
 }
 
+type key struct {
+	Environment string
+	Application string
+}
+
 func (v *versionClient) ConsumeEvents(ctx context.Context, processor VersionEventProcessor) error {
 	ctx = auth.WriteUserToGrpcContext(ctx, RolloutServiceUser)
 outer:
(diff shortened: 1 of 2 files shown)


services/rollout-service/pkg/versions/versions.go
```diff
@@ -132,6 +144,16 @@ outer:
 					}
 				}
 			}
+			for k := range versions {
+				if seen[k] == 0 {
+					processor.ProcessKuberpultEvent(ctx, KuberpultEvent{
```
sven-urbanski-freiheit-com, line 149: What are really achieving here? Calling processor.ProcessKuberpultEvent(version=0) for all apps previously processed?
Maybe add a comment to explain why that is necessary. Or rename `seen`, it seems to be used only for the 0-version. Maybe `processZeroVersion`.

hannesg: Ah, right, this is for taking care of removed applications. If we don't send a zero value, the rollout service would consider  them deployed for-ever and would always reconsider them for notification. I will write a comment on that.

sven-urbanski-freiheit-com: Maybe also mention this in the committe message

hannesg: Yes, I added this to the message and added a comment.

Issue #878, opened 2023-08-21
Warn in the UI when there are too many items in the cart


sven-urbanski-freiheit-com:
Currently kuberpult only warns after submitting. That is very inconvenient if you have already ~150 items  in there

sven-urbanski-freiheit-com:
This was implemented in 1.4.1

PR #930, opened 2023-09-08
Show error about too many items before submit

sven-urbanski-freiheit-com:
The server always allowed only 100 actions.
Now the UI also checks if there are too many items, and stops the user before adding the 101st action.

This fixes issue https://github.com/freiheit-com/kuberpult/issues/878

### DIFF
--- a/services/frontend-service/src/ui/utils/store.test.tsx
+++ b/services/frontend-service/src/ui/utils/store.test.tsx
@@ -16,15 +16,25 @@ Copyright 2023 freiheit.com*/
 import { act, renderHook } from '@testing-library/react';
 import {
     AllLocks,
+    appendAction,
     FlushRolloutStatus,
+    UpdateAction,
     updateActions,
     UpdateOverview,
     UpdateRolloutStatus,
+    UpdateSnackbar,
     useLocksSimilarTo,
     useNavigateWithSearchParams,
     useRolloutStatus,
 } from './store';
-import { BatchAction, EnvironmentGroup, Priority, RolloutStatus, StreamStatusResponse } from '../../api/api';
+import {
+    BatchAction,
+    EnvironmentGroup,
+    LockBehavior,
+    Priority,
+    RolloutStatus,
+    StreamStatusResponse,
+} from '../../api/api';
 import { makeDisplayLock, makeLock } from '../../setupTests';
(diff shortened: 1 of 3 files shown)


services/frontend-service/src/ui/utils/store.test.tsx
```diff
@@ -452,3 +462,61 @@ describe('Rollout Status', () => {
...
+        {
+            name: 'at limit',
+            inputActionsLen: 100,
+            expectedLen: 100,
+            expectedShowError: false,
+        },
+        {
+            name: 'over limit',
+            inputActionsLen: 101,
+            expectedLen: 100,
```
leandro-freiheit, line 490: I am a bit hesitant here. Because if the snackbar errors, I am actually expecting only a error and not an action list of size, for example, 100. Or am I missing something?

services/frontend-service/src/ui/utils/store.test.tsx
```diff
@@ -452,3 +462,61 @@ describe('Rollout Status', () => {
...
+        {
+            name: 'below limit',
+            inputActionsLen: 99,
+            expectedLen: 99,
+            expectedShowError: false,
+        },
+        {
+            name: 'at limit',
+            inputActionsLen: 100,
+            expectedLen: 100,
```
leandro-freiheit, line 484: Won't this fail? We are adding 100 actions which is fine but then in 
```
    if (actions.length + 1 > maxBatchActions) {
        showSnackbarError('Maximum number of actions is ' + String(maxBatchActions));
        return;
    }
```
this will fail because 101 is larger than the 100 allowed actions

services/frontend-service/src/ui/utils/store.test.tsx
sven-urbanski-freiheit-com: the test puts in 101 actions one by one. So at the end there will be only 100 items in it, because the last one is not allowed to be added.

services/frontend-service/src/ui/utils/store.test.tsx
sven-urbanski-freiheit-com: no, at that point `actions.length` is 99. The actual "adding" comes later in the function.

leandro-freiheit (approved):

PR #936, opened 2023-09-11
Add Release Version to the frontend SideBar

gsandok:
Example of how this would look like for a deployment
Screenshot 2023-09-11 at 10 28 27 AM

and 
Screenshot 2023-09-11 at 11 53 28 AM

### DIFF
--- a/services/frontend-service/src/ui/components/ReleaseCardMini/ReleaseCardMini.scss
+++ b/services/frontend-service/src/ui/components/ReleaseCardMini/ReleaseCardMini.scss
@@ -28,9 +28,9 @@ Copyright 2023 freiheit.com*/
     width: inherit;
     margin: 10px;
     font-style: normal;
-    font-weight: 700;
-    font-size: 18px;
-    line-height: 32px;
+    font-weight: 500;
+    font-size: 14px;
+    line-height: 24px;
     color: #4a4a4a;
 }
 
--- a/services/frontend-service/src/ui/components/ReleaseCardMini/ReleaseCardMini.scss
+++ b/services/frontend-service/src/ui/components/ReleaseCardMini/ReleaseCardMini.scss
@@ -27,10 +27,7 @@ Copyright 2023 freiheit.com*/
     height: 100%;
     width: inherit;
     margin: 10px;
-    font-style: normal;
-    font-weight: 500;
-    font-size: 14px;
-    line-height: 24px;
+    @extend .text-regular;
     color: #4a4a4a;
 }
 
--- a/services/frontend-service/src/ui/components/SideBar/SideBar.tsx
+++ b/services/frontend-service/src/ui/components/SideBar/SideBar.tsx
@@ -202,12 +202,12 @@ type SideBarListItemProps = {
 export const SideBarListItem: React.FC<{ children: BatchAction }> = ({ children: action }: SideBarListItemProps) => {
     const { environmentLocks, appLocks } = useAllLocks();
     const actionDetails = getActionDetails(action, appLocks, environmentLocks);
-    const DisplayVersion = (): string => {
+    const DisplayVersion = () => {
         try {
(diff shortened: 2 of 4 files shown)


services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -200,6 +202,13 @@ type SideBarListItemProps = {
 export const SideBarListItem: React.FC<{ children: BatchAction }> = ({ children: action }: SideBarListItemProps) => {
     const { environmentLocks, appLocks } = useAllLocks();
     const actionDetails = getActionDetails(action, appLocks, environmentLocks);
+    var displayVersion = '';
+    try {
+        const release = useReleaseOrThrow(actionDetails.application ?? '', actionDetails.version ?? 0);
+        displayVersion = release.displayVersion;
+    } catch (error) {
+        // continue without displayVersion
+    }
```
jaypontes, line 211:
```suggestion
    var displayVersion = useRelease(actionDetails.application ?? '', actionDetails.version ?? 0).displayVersion | '';
```
Would something like this be possible?
Why do we have useRelease throwing errors?

leandro-freiheit: I agree with João. Since we do not want to throw an error, we can simply call useRelease

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -200,6 +202,13 @@ type SideBarListItemProps = {
 export const SideBarListItem: React.FC<{ children: BatchAction }> = ({ children: action }: SideBarListItemProps) => {
     const { environmentLocks, appLocks } = useAllLocks();
     const actionDetails = getActionDetails(action, appLocks, environmentLocks);
+    var displayVersion = '';
```
jaypontes, line 205: Also I'd rather have this as a `const` if possible

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
gsandok: Unfortunately, useRelease is not exported but I can create a helper function to move this logic

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
leandro-freiheit: This can be omitted if later on you specify `const displayVersion = useRelease(actionDetails.application ?? '', actionDetails.version ?? 0).displayVersion | '';`

gsandok: Since useRelease() is not exported I am unable to make the change specified

services/frontend-service/src/ui/components/ReleaseCardMini/ReleaseCardMini.scss
```diff
@@ -28,9 +28,9 @@ Copyright 2023 freiheit.com*/
     width: inherit;
     margin: 10px;
     font-style: normal;
```
jaypontes, line 30: use typography!
example: `@extend .text-regular;`

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
jaypontes:
```suggestion
    const displayVersion => {
        try {
            const release = useReleaseOrThrow(actionDetails.application ?? '', actionDetails.version ?? 0);
            return release.displayVersion;
        } catch (error) {
            return ''
        }
    }
```
I'd go with something like this. Not entirely sure the syntax is right or if there's some typescript error that might arise from this but I think it's cleaner and keeps the best of both worlds (const and error handling)

leandro-freiheit (approved):

PR #941, opened 2023-09-13
Add networking timeouts to git pushes

hannesg:
The solution take is highly unelegant but kinda unavoidable given the options available. Basically, libgit as most c based libs doesn't allow cancelling anything. The only thing we can do is to run the push operation in a seperate go routine and return our own timeout after some time while throwing a cancellation at the lib to stop. This is now done in the `useRemote` function. The network timeout is now 1 minute which is probably good enough for most consumers.

### DIFF
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -93,7 +93,6 @@ type repository struct {
 	writeLock    sync.Mutex
 	writesDone   uint
 	queue        queue
-	remote       *git.Remote
 	config       *RepositoryConfig
 	credentials  *credentialsStore
 	certificates *certificateStore
@@ -120,6 +119,8 @@ type RepositoryConfig struct {
 	CommitterName  string
 	// default branch is master
 	Branch string
+	// network timeout
+	NetworkTimeout time.Duration
 	//
 	GcFrequency    uint
 	StorageBackend StorageBackend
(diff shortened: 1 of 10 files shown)


services/cd-service/pkg/repository/repository.go
```diff
@@ -362,7 +384,9 @@ type PushActionCallbackFunc func(git.PushOptions, *repository) PushActionFunc
 // DefaultPushActionCallback is public for testing reasons only.
 func DefaultPushActionCallback(pushOptions git.PushOptions, r *repository) PushActionFunc {
 	return func() error {
-		return r.remote.Push([]string{fmt.Sprintf("refs/heads/%s:refs/heads/%s", r.config.Branch, r.config.Branch)}, &pushOptions)
+		return r.useRemote(context.Background(), func(remote *git.Remote) error {
```
leandro-freiheit, line 387: What exactly is the difference here?

services/cd-service/pkg/repository/repository_test.go
```diff
@@ -1511,3 +1515,71 @@ func TestProcessQueueOnce(t *testing.T) {
...
+			err = repo.Apply(testutil.MakeTestContext(),
+				&CreateEnvironment{Environment: "stg"},
+			)
+			if err == nil {
+				t.Errorf("expected an error, but didn't get one")
+			}
+			if status.Code(err) != codes.Canceled {
+				t.Errorf("expected status code cancelled, but got %q", status.Code(err))
+			}
+			// This will prevent the next push from working
```
leandro-freiheit, line 1575: Will it be prevented? Technically there is no delay or am I missing something?

services/cd-service/pkg/repository/repository.go
hannesg: The difference is that `useRemote` opens a new connection every time. Also it ensures that the connection is closed at the end and, most importantly, it is guarateed to return after `NetworkTimeout` so that calls don't stack up.

services/cd-service/pkg/repository/repository_test.go
hannesg: No, you are right. The wording must be the other way round.

PR #955, opened 2023-09-20
UI: Show warning before deploying over a lock

sven-urbanski-freiheit-com:
If the user wants to deploy an app that has a lock, or if the user wants to deploy to an environment that has a lock, we now show a warning (when submitting the actions), because generally locks should not be ignored.

The dialog shows which locks are the "offenders".

The user can still choose to ignore this, but now they have been warned.

New Dialog with offending locks:
image

New Dialog without offending locks:
image

### DIFF
--- a/services/frontend-service/src/ui/components/EnvironmentLockDisplay/EnvironmentLockDisplay.tsx
+++ b/services/frontend-service/src/ui/components/EnvironmentLockDisplay/EnvironmentLockDisplay.tsx
@@ -26,7 +26,7 @@ export const DisplayLockInlineRenderer: React.FC<{ lock: DisplayLock }> = (props
 
     const description = lock.application ? (
         <span>
-            App <b>{lock.application}</b> locked by <b>{author}</b> on environment <b>{lock.environment}</b>
+            Application <b>{lock.application}</b> locked by <b>{author}</b> on environment <b>{lock.environment}</b>
         </span>
     ) : (
         <span>
--- a/services/frontend-service/src/ui/components/EnvironmentLockDisplay/EnvironmentLockDisplay.tsx
+++ b/services/frontend-service/src/ui/components/EnvironmentLockDisplay/EnvironmentLockDisplay.tsx
@@ -19,6 +19,22 @@ import { Tooltip } from '../tooltip/tooltip';
 import { Locks, LocksWhite } from '../../../images';
 import { Button } from '../button';
 
+export const DisplayLockInlineRenderer: React.FC<{ lock: DisplayLock }> = (props) => {
+    const { lock } = props;
+    const hasAuthor = lock.authorEmail || lock.authorName;
+    const author = hasAuthor ? lock.authorName + '<' + lock.authorEmail + '>' : 'unknown';
+    return (
+        <span title={lock.lockId}>
+            <span>
+                locked by <b>{author}</b>
+            </span>
+            <span>
+                {' '}
+                with message: <b>{lock.message}</b>{' '}
+            </span>
+        </span>
+    );
+};
 export const DisplayLockRenderer: React.FC<{ lock: DisplayLock }> = (props) => {
     const { lock } = props;
     const hasAuthor = lock.authorEmail || lock.authorName;
--- a/services/frontend-service/src/ui/components/EnvironmentLockDisplay/EnvironmentLockDisplay.tsx
+++ b/services/frontend-service/src/ui/components/EnvironmentLockDisplay/EnvironmentLockDisplay.tsx
@@ -23,14 +23,22 @@ export const DisplayLockInlineRenderer: React.FC<{ lock: DisplayLock }> = (props
     const { lock } = props;
     const hasAuthor = lock.authorEmail || lock.authorName;
     const author = hasAuthor ? lock.authorName + '<' + lock.authorEmail + '>' : 'unknown';
+
+    const description = lock.application ? (
+        <span>
+            App <b>{lock.application}</b> locked by <b>{author}</b> on environment <b>{lock.environment}</b>
+        </span>
+    ) : (
+        <span>
+            Environment <b>{lock.environment}</b> locked by <b>{author}</b>
+        </span>
+    );
     return (
         <span title={lock.lockId}>
-            <span>
-                locked by <b>{author}</b>
-            </span>
+            {description}
             <span>
                 {' '}
-                with message: <b>{lock.message}</b>{' '}
+                with message: <b>'{lock.message}'</b>{' '}
             </span>
         </span>
     );
--- a/services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.tsx
+++ b/services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.tsx
@@ -112,26 +112,23 @@ export const EnvironmentListItem: React.FC<EnvironmentListItemProps> = ({
             },
         });
     }, [app, env.name]);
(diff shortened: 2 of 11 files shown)


services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -372,6 +385,76 @@ export const SideBar: React.FC<{ className?: string; toggleSidebar: () => void }
...
+                headerLabel={'Please Confirm the Deployment over Locks'}
+                onConfirm={applyActions}
+                confirmLabel={'Confirm Deployment'}
+                onCancel={cancelConfirmation}
+                open={dialogState.showConfirmationDialog}>
+                <div>
+                    You are attempting to deploy apps, although there are locks present. Please check the locks and be
+                    sure you really want to ignore them.
+                    <div className={'locks'}>
+                        {appLocksRendered}
```
leandro-freiheit, line 436: In the locks page, we display the environment locks first and then the application locks. I think we should display the environment locks first and then application locks for consistency

services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.tsx
```diff
@@ -112,23 +112,27 @@ export const EnvironmentListItem: React.FC<EnvironmentListItemProps> = ({
...
-                        environment: env.name,
-                        application: app,
-                        version: release.version,
-                        ignoreAllLocks: false,
-                        lockBehavior: LockBehavior.Ignore,
-                    },
+    const onConfirm = useCallback((): void => {
+        addAction({
+            action: {
+                $case: 'deploy',
```
leandro-freiheit, line 118: Since the function now is onConfirm, it makes more sense to me that the case is passed as a parameter. Either that or maybe rename the function to `onDeployConfirm`

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -315,6 +315,41 @@ export const useFilteredApplicationLocks = (appNameParam: string | null): Displa
...
 };
 
+export const useLocksConflictingWithActions = (): AllLocks => {
+    const allActions = useActions();
+    const locks = useAllLocks();
+    return {
+        environmentLocks: locks.environmentLocks.filter((envLock: DisplayLock) => {
+            const actions = allActions.filter((action) => {
+                if (action.action?.$case === 'deploy') {
+                    // const app = action.action.deploy.application;
```
leandro-freiheit, line 325: This can be removed

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -315,6 +315,15 @@ export const SideBar: React.FC<{ className?: string; toggleSidebar: () => void }
...
     const [showSpinner, setShowSpinner] = useState(false);
+    const [dialogState, setDialogState] = useState({
+        showConfirmationDialog: false,
+    });
+    const cancelConfirmation = useCallback((): void => {
+        setDialogState({ showConfirmationDialog: false });
+    }, []);
+
+    const conflictingLocks = useLocksConflictingWithActions();
+    const hasLocks = conflictingLocks.environmentLocks.length > 0 || conflictingLocks.appLocks.length > 0;
```
leandro-freiheit, line 326:
```suggestion
    const hasLocks = conflictingLocks.environmentLocks.length > 0 && conflictingLocks.appLocks.length > 0;
```
We can have only app locks or only env locks (and it would not be an error case). But here we say that we can have only env locks (for example) and we would get the dialog with conflicting locks

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -360,6 +395,28 @@ export type AllLocks = {
     appLocks: DisplayLock[];
 };
 
+export const useEnvLocks = (thisEnv: string): DisplayLock[] => {
```
leandro-freiheit, line 398: This function was not used right?

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
leandro-freiheit: Also the name of the variable should be different

sven-urbanski-freiheit-com: > here we say that we can have only env locks (for example) and we would get the dialog with conflicting locks

This is intended. If there is only one env lock, we show the dialog. Same if there is only one app lock.

services/frontend-service/src/ui/utils/store.tsx
sven-urbanski-freiheit-com: done

services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.tsx
sven-urbanski-freiheit-com: I combined and renamed the functions

services/frontend-service/src/ui/utils/store.tsx
sven-urbanski-freiheit-com: removed it.

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
sven-urbanski-freiheit-com: done

services/frontend-service/src/ui/components/EnvironmentLockDisplay/EnvironmentLockDisplay.tsx
```diff
@@ -19,6 +19,30 @@ import { Tooltip } from '../tooltip/tooltip';
...
 import { Button } from '../button';
 
+export const DisplayLockInlineRenderer: React.FC<{ lock: DisplayLock }> = (props) => {
+    const { lock } = props;
+    const hasAuthor = lock.authorEmail || lock.authorName;
+    const author = hasAuthor ? lock.authorName + '<' + lock.authorEmail + '>' : 'unknown';
+
+    const description = lock.application ? (
+        <span>
+            App <b>{lock.application}</b> locked by <b>{author}</b> on environment <b>{lock.environment}</b>
```
hannesg, line 29:
```suggestion
            Application <b>{lock.application}</b> locked by <b>{author}</b> on environment <b>{lock.environment}</b>
```

hannesg (approved):

services/frontend-service/src/ui/components/SideBar/SideBar.tsx
leandro-freiheit: But if we only have one environment lock, we would be expecting the release dialog to only display the non-conflicting dialog. However, when we check `hasLocks` it only renders the non-conflicting dialog if the value is false. But `hasLocks` is set to true if we have either app locks or env locks, when I expect that it's set to true only if we have appLocks and envLocks. (services/frontend-service/src/ui/components/SideBar/SideBar.tsx line 420)

leandro-freiheit: Nevermind, the filter for actions is inside a filter for the environment locks/app locks. It is completely correct 👍

leandro-freiheit (approved):

services/frontend-service/src/ui/components/EnvironmentLockDisplay/EnvironmentLockDisplay.tsx
sven-urbanski-freiheit-com: done

PR #965, opened 2023-09-27
testdata: do not include hash in commit message

sven-urbanski-freiheit-com:


hannesg (approved):
But we should definitely have testcases that contain veeeeeeeeeery nasty commit messages

PR #967, opened 2023-09-27
Fix shell invocation for license check

sven-urbanski-freiheit-com:
We are using bash features in the shell script, so we cannot call it with sh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ analyze/pull-request: $(CODE_REVIEWER_LOCATION)
 
 .PHONY: check-license
 check-license:
-	@sh check.sh || (echo run "bash check.sh" locally, commit the result and push; exit 1)
+	@bash check.sh || (echo run "bash check.sh" locally, commit the result and push; exit 1)
 
 .PHONY: version
 version:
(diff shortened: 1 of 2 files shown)


ahmed-nour-fdc (approved):

Makefile
```diff
@@ -76,7 +76,7 @@ analyze/pull-request: $(CODE_REVIEWER_LOCATION)
 
 .PHONY: check-license
 check-license:
-	@sh check.sh || (echo run "bash check.sh" locally, commit the result and push; exit 1)
+	@bash check.sh || (echo run "bash check.sh" locally, commit the result and push; exit 1)
```
leandro-freiheit, line 79: We should use the SHELL variable that is set before or remove it entirely. I would prefer to use `SHELL:=bash` and reference it here

leandro-freiheit (approved):

Makefile
sven-urbanski-freiheit-com: I think in general it's better to use sh in Makefiles, because those scripts are simpler. SHELL is also used to invoke any command lines, so that would have potentially a big impact.
But just for this script, we *need* bash, because this script uses some advanced features.

PR #972, opened 2023-09-27
Add pod annotations to the rollout pod

hannesg:
This adds a new option to the rollout service to add annotations to the pod directly. The current annotations will be used as a default and will be merge with the specified ones.

### DIFF
--- a/charts/kuberpult/values.yaml.tpl
+++ b/charts/kuberpult/values.yaml.tpl
@@ -89,6 +89,7 @@ rollout:
     requests:
       cpu: 500m
       memory: 250Mi
+  podAnnotations: {}
 
 ingress:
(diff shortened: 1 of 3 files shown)


charts/kuberpult/values.yaml.tpl
```diff
@@ -89,6 +89,7 @@ rollout:
     requests:
       cpu: 500m
       memory: 250Mi
+  podAnnotations: {}
```
sven-urbanski-freiheit-com, line 92:
```suggestion
  # annotations given here will take precedence over the defaults defined in _helpers.tpl
  podAnnotations: {}
```

sven-urbanski-freiheit-com (approved):

PR #990, opened 2023-10-09
Add a rest endpoint for obtaining the rollout status of a whole environment group

hannesg:
The new rest endpoint is available under `/environment-groups/<name>/rollout-status`

A script is available under `./infrastructure/script/env-group-status.sh`


services/frontend-service/pkg/handler/handleEnvironments.go
```diff
@@ -35,6 +35,8 @@ func (s Server) HandleEnvironmentGroups(w http.ResponseWriter, req *http.Request
 	switch function {
 	case "locks":
 		s.handleEnvironmentGroupLocks(w, req, envGroup, tail)
+	case "rollout-status":
+		s.handleEnvironmentGroupRolloutStatus(w, req, envGroup)
```
sven-urbanski-freiheit-com, line 39: please create a simple script like `https://github.com/freiheit-com/kuberpult/blob/main/infrastructure/scripts/create-testdata/create-app-lock.sh` for the new endpoint, so it's easier to run manually

services/frontend-service/pkg/cmd/server.go
```diff
@@ -486,3 +487,10 @@ func (p *GrpcProxy) StreamStatus(in *api.StreamStatusRequest, stream api.Rollout
 		}
 	}
 }
+
+func (p *GrpcProxy) GetStatus(ctx context.Context, in *api.GetStatusRequest) (*api.GetStatusResponse, error) {
+	if p.RolloutServiceClient == nil {
+		return nil, status.Error(codes.Unimplemented, "rollout status not implemented")
```
sven-urbanski-freiheit-com, line 493:
```suggestion
		return nil, status.Error(codes.Unimplemented, "rollout status not implemented, service client not configured")
```
(or something like that)

services/frontend-service/pkg/handler/rolloutstatus.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/ProtonMail/go-crypto/openpgp"
+	pgperrors "github.com/ProtonMail/go-crypto/openpgp/errors"
+	"github.com/freiheit-com/kuberpult/pkg/api"
+	"github.com/freiheit-com/kuberpult/pkg/logger"
+	"go.uber.org/zap"
+)
+
+func (s *Server) handleEnvironmentGroupRolloutStatus(w http.ResponseWriter, req *http.Request, environmentGroup string) {
+	if s.RolloutClient == nil {
+		http.Error(w, "not implemented", http.StatusNotImplemented)
```
sven-urbanski-freiheit-com, line 34: same here

services/rollout-service/pkg/service/broadcast.go
```diff
@@ -246,5 +275,43 @@ func rolloutStatus(ev *ArgoEvent) api.RolloutStatus {
...
+	api.RolloutStatus_RolloutStatusProgressing,
+	api.RolloutStatus_RolloutStatusUnhealthy,
+	api.RolloutStatus_RolloutStatusPending,
+	api.RolloutStatus_RolloutStatusUnknown,
+
+	// This is the only successful state
+	api.RolloutStatus_RolloutStatusSuccesful,
+}
+
+func statusPriority(a api.RolloutStatus) int {
```
sven-urbanski-freiheit-com, line 297:
```suggestion
// 0 is the highest priority - (RolloutStatusSuccesful) is the lowest priority
func statusPriority(a api.RolloutStatus) int {
```

sven-urbanski-freiheit-com (approved):

services/rollout-service/pkg/service/broadcast_test.go
```diff
@@ -437,3 +483,140 @@ func TestBroadcastDoesntGetStuck(t *testing.T) {
...
+			for _, s := range tc.ArgoEvents {
+				bc.ProcessArgoEvent(context.Background(), s)
+			}
+			for _, s := range tc.KuberpultEvents {
+				bc.ProcessKuberpultEvent(context.Background(), s)
+			}
+
+			ctx, cancel := context.WithCancel(context.Background())
+			resp, err := bc.GetStatus(ctx, tc.Request)
+			cancel()
```
leandro-freiheit, line 612: I'm not really familiar with this function here

services/rollout-service/pkg/service/broadcast_test.go
```diff
@@ -437,3 +483,140 @@ func TestBroadcastDoesntGetStuck(t *testing.T) {
...
+					{
+						Environment:   "dev",
+						Application:   "foo",
+						RolloutStatus: api.RolloutStatus_RolloutStatusPending,
+					},
+				},
+			},
+		},
+		{
+			Name: "exludes succesful applications",
```
leandro-freiheit, line 552: I have my doubts here, I think I am not understanding everything here. Because the test data is the same and the response does not differ from the previous test case.

services/rollout-service/pkg/service/broadcast_test.go
hannesg: Hmm, right, It's also not really needed here. I'll remove it

services/rollout-service/pkg/service/broadcast_test.go
hannesg: Ah, yes, the result is the same but the test data is slightly different. In the second case, both applications are in the same env group, so the successful one is exlcuded.

PR #995, opened 2023-10-11
Close dialogs on click outside the dialog

sven-urbanski-freiheit-com:


services/frontend-service/src/ui/components/dialog/ConfirmationDialog.tsx
```diff
@@ -73,7 +73,19 @@ export const PlainDialog: React.FC<PlainDialogProps> = (props) => {
                 onClose();
             }
         });
-    }, [onClose]);
+        document.addEventListener('click', (event) => {
+            if (open) {
+                if (event.target instanceof HTMLElement) {
+                    // eslint-disable-next-line no-type-assertion/no-type-assertion
+                    const target = event.target as HTMLElement;
```
jaypontes, line 80: Is there no way to get around the `as HTMLElement`? :(

sven-urbanski-freiheit-com: I am not aware of one.

`event.target`s type is very generic.

jaypontes: Having the condition in the line 78 doesn't remove the need for the cast?

jaypontes (approved):

services/frontend-service/src/ui/components/dialog/ConfirmationDialog.tsx
sven-urbanski-freiheit-com: You are right, it works :smiley: 
I simplified it

PR #1026, opened 2023-10-18
Make it easier to run integration tests locally

hannesg:
I increased the run-kind timeout to infinity and I made the integration tests repeatable locally.

### DIFF
--- a/tests/integration-tests/main_test.go
+++ b/tests/integration-tests/main_test.go
@@ -1,5 +1,4 @@
-/*
-This file is part of kuberpult.
+/*This file is part of kuberpult.
 
 Kuberpult is free software: you can redistribute it and/or modify
 it under the terms of the Expat(MIT) License as published by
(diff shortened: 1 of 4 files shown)


tests/integration-tests/main_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	nextRun := nr + 1
+	os.WriteFile(".run-number", []byte(fmt.Sprintf("%d", nextRun)), 0644)
+	runNumber = nr
+	appSuffix = base26(runNumber)
+	return nil
+}
+
+func TestMain(m *testing.M) {
+	err := calculateRunNumber()
```
sven-urbanski-freiheit-com, line 66: I'm fine with this change overall.
Please add a bit more explanation why the run-number is making it easier to run locally.
You only seem to be writing it to a file, but not actually using it (or I don't see where you use it).

sven-urbanski-freiheit-com (approved):

tests/integration-tests/main_test.go
hannesg: I renamed the method. The run number is indeed not really used but just the app suffix.

PR #1045, opened 2023-10-23
Added the first two endpoints for the repository service

hannesg:
The first two endpoints `GenerateManifest` and `ResolveRevision` are the absolute minimum that is  needed to make  the integration test  work. I have made sure  that our implementation is correct by starting the argocd repo server in the unit tests and tested that they give the  same responses.

### DIFF
--- a/services/cd-service/pkg/fs/fs.go
+++ b/services/cd-service/pkg/fs/fs.go
@@ -484,7 +484,7 @@ func (t *TreeBuilderFS) MkdirAll(dir string, perm os.FileMode) error {
 }
 
 func (t *TreeBuilderFS) Lstat(path string) (os.FileInfo, error) {
-	// TODO(HVG): implement this to support actual symlinks
+	// TODO(HVG): implement this to support actual symlinkk (https://github.com/freiheit-com/kuberpult/issues/1046)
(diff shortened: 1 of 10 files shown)


services/cd-service/pkg/fs/fs.go
```diff
@@ -484,7 +484,8 @@ func (t *TreeBuilderFS) MkdirAll(dir string, perm os.FileMode) error {
 }
 
 func (t *TreeBuilderFS) Lstat(path string) (os.FileInfo, error) {
-	return nil, billy.ErrNotSupported
+	// TODO(HVG): implement this to support actual symlinks
```
sven-urbanski-freiheit-com, line 487: when will this TODO be done? Please add a reference to a story that handles it (or do it now)

sven-urbanski-freiheit-com (approved):

services/cd-service/pkg/fs/fs.go
hannesg: Yes, I will!

hannesg: I can probably do it in the same story but I will put it in a seperate PR.

PR #1062, opened 2023-10-26
Deleting empty directories in the manifest repository

sven-urbanski-freiheit-com:
This is a workaround for a BitBucket issue, where it does not display directories, if they have an empty directory in them.
With this change kuberpult deletes empty lock directories.

Note that it's rather unlikely to run into this issue locally or on another git hosting provider.
The git command line client handles empty directories transparently: They are deleted automatically if empty, and they are added automatically, if a file gets added. However, this is done in the git command line, but not in libgit2, which is more low-level and which is what we use in kuberpult.

### DIFF
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -1183,8 +1183,20 @@ func readLock(fs billy.Filesystem, lockDir string) (*Lock, error) {
 	return lock, nil
 }
 
+func (s *State) GetEnvLocksDir(environment string) string {
+	return s.Filesystem.Join("environments", environment, "locks")
+}
+
+func (s *State) GetEnvLockDir(environment string, lockId string) string {
+	return s.Filesystem.Join(s.GetEnvLocksDir(environment), lockId)
+}
+
+func (s *State) GetAppLocksDir(environment string, application string) string {
+	return s.Filesystem.Join("environments", environment, "applications", application, "locks")
+}
+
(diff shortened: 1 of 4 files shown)


services/cd-service/pkg/repository/repository.go
```diff
@@ -1252,6 +1264,48 @@ func (s *State) GetDeploymentMetaData(ctx context.Context, environment, applicat
...
+const (
+	NoReason SuccessReason = iota
+	DirDoesNotExist
+	DirNotEmpty
+)
+
+// DeleteDirIfEmpty if it's empty. If the dir does not exist or is not empty, nothing happens.
+// Errors are only returned if the read or delete operations fail.
+// Returns SuccessReason for unit testing.
+func (s *State) DeleteDirIfEmpty(ctx context.Context, directoryName string) (error, SuccessReason) {
```
hannesg, line 1290: `error` should be the last return value.

services/cd-service/pkg/repository/repository.go
```diff
@@ -1252,6 +1264,48 @@ func (s *State) GetDeploymentMetaData(ctx context.Context, environment, applicat
...
+	fileInfos, err := s.Filesystem.ReadDir(directoryName)
+	if err != nil {
+		return fmt.Errorf("DeleteDirIfEmpty: failed to read directory %q: %w", directoryName, err), 0
```
hannesg, line 1293: `NoReason` would be better  than `0` here.

hannesg (approved):

PR #1064, opened 2023-10-26
Add DEX helm chart to Kuberpult helm chart 

bernardo-bastos:
If Dex is enabled, installs the DEX helm chart together with Kuberpult.

REV: DSN-0Z3L73

### DIFF
--- a/charts/kuberpult/Chart.yaml.tpl
+++ b/charts/kuberpult/Chart.yaml.tpl
@@ -23,14 +23,14 @@ version: $VERSION
 # It is recommended to use it with quotes.
 appVersion: "$VERSION"
 
-# This is the DEX helm chart which will only be installed if `dex.installDex.enabled` is true.
+# This is the DEX helm chart which will only be installed if `auth.dexAuth.installDex.enabled` is true.
 # Dex is an identity service that uses OpenID Connect to drive authentication through other 
 # identity providers.  
 # For more information please check: https://github.com/dexidp/dex
 dependencies:
 - name: dex
-condition: dex.installDex.enabled
-version: dex.installDex.version
+condition: auth.dexAuth.installDex.enabled
+version: auth.dexAuth.installDex.version
 repository: https://charts.dexidp.io
 
 maintainers:
--- a/charts/kuberpult/Chart.yaml.tpl
+++ b/charts/kuberpult/Chart.yaml.tpl
@@ -23,14 +23,14 @@ version: $VERSION
 # It is recommended to use it with quotes.
 appVersion: "$VERSION"
 
-# This is is the DEX helm chart only to installed it RBAC is enabled.
+# This is the DEX helm chart which will only be installed if `dex.installDex.enabled` is true.
 # Dex is an identity service that uses OpenID Connect to drive authentication through other 
 # identity providers.  
 # For more information please check: https://github.com/dexidp/dex
 dependencies:
 - name: dex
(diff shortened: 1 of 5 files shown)


charts/kuberpult/Chart.yaml.tpl
```diff
@@ -23,6 +23,16 @@ version: $VERSION
 # It is recommended to use it with quotes.
 appVersion: "$VERSION"
 
+# This is is the DEX helm chart only to installed it RBAC is enabled.
```
sven-urbanski-freiheit-com, line 26:
```suggestion
# This is the DEX helm chart which will only be installed it RBAC is enabled.
```

charts/kuberpult/Chart.yaml.tpl
```diff
@@ -23,6 +23,16 @@ version: $VERSION
...
+# Dex is an identity service that uses OpenID Connect to drive authentication through other 
+# identity providers.  
+# For more information please check: https://github.com/dexidp/dex
+dependencies:
+- name: dex
+condition: dex.enabled
+version: "0.14.2"
```
sven-urbanski-freiheit-com, line 33: Can we make the version configurable?

charts/kuberpult/values.yaml.tpl
```diff
@@ -218,3 +218,26 @@ dex:
   baseURL: ""
   # List of scopes to validate the token. Please add them as comma separated values.
   scopes: ""
+  # List of environment variables to be added to the dex service pod.
```
sven-urbanski-freiheit-com, line 221: @hannesg please review this PR too

charts/kuberpult/Chart.yaml.tpl
hannesg: :thinking: Maybe also make it possible to use an external dex?

hannesg (commented):
I think this is tricky because you are re-using the `dex` namespace in the values. I think it would be better to move our own config to a seperate top-level value like `auth.dex` to seperate concerns. But apart from that, I'm fine.

charts/kuberpult/values.yaml.tpl
```diff
@@ -199,22 +199,48 @@ auth:
...
+    # List of environment variables to be added to the dex service pod.
+    # Example, if you want your DEX service to have access to to the OAUTH_CLIENT_ID, you can specify
+    # it the following way: 
+    # 
+    #  - name: OAUTH_CLIENT_ID
+    #    valueFrom:
+    #      secretKeyRef:
+    #      name: kuberpult-oauth-client-id
+    #        key: kuberpult-oauth-client-id
+    envVars: []
```
hannesg, line 231: This config must be under `dex`, right? Otherwise the dex chart won't pick it up.

bernardo-bastos: Yes, you are right. Made the necessary changes and re tested everything locally on my team dummy cluster.

charts/kuberpult/Chart.yaml.tpl
bernardo-bastos: Corrected the comment.

charts/kuberpult/Chart.yaml.tpl
bernardo-bastos: Added an option to omit Dex installation

hannesg (approved):

PR #1075, opened 2023-11-01
Use earthly to build kuberpult locally

ahmed-nour-fdc:
* Kuberpult can now be build using [earthly](https://github.com/earthly/earthly). Download the earthly binary and run `make kuberpult-earthly` to build the services in a containerised environment and run the services using the built images.


Earthfile
```diff
@@ -0,0 +1,51 @@
...
+    RUN OS=Linux ARCH=$(uname -m) && \
+        SHA=$(cat buf_sha256.txt | grep "buf-${OS}-${ARCH}$" | cut -d ' ' -f1) && \
+        echo "${SHA}  ${BUF_BIN_PATH}/buf" | sha256sum -c
+    SAVE ARTIFACT go.mod
+    SAVE ARTIFACT go.sum
+    SAVE ARTIFACT buf_sha256.txt
+
+cd-service:
+    ARG UID=1000
+    ARG target=docker
```
sven-urbanski-freiheit-com, line 23: is it possible to define this only once? E.g. put it in the "deps" and depend on it?

ahmed-nour-fdc: Not yet supported, unfortunately.

pkg/path/Earthfile
```diff
@@ -0,0 +1,9 @@
+VERSION 0.7
```
sven-urbanski-freiheit-com, line 1: that's a lot of files for everything in `pkg`.
Could we instead have one file `pkg/Earthfile` to take care of all of it?

services/cd-service/.arg
```diff
@@ -0,0 +1 @@
+service=cd-service
```
sven-urbanski-freiheit-com, line 1: Why do we need this?

services/cd-service/Earthfile
```diff
@@ -0,0 +1,65 @@
+VERSION 0.7
+
+deps:
+    FROM ../../+deps
+    ARG service
+    ARG tag="deps"
+    COPY ../../pkg/api/+artifacts/* pkg/api/
```
sven-urbanski-freiheit-com, line 7:
```suggestion
    COPY ../../pkg/*/+artifacts/* pkg/*/
```
?

services/cd-service/.arg
ahmed-nour-fdc: Removed

pkg/path/Earthfile
ahmed-nour-fdc: It turned out to be easier than I thought. Now there is only one Earthfile in pkg

sven-urbanski-freiheit-com (approved):

PR #1076, opened 2023-11-01
Integration test for frontend main page

sven-urbanski-freiheit-com:


services/cd-service/Buildfile.yaml
```diff
@@ -12,3 +12,4 @@ spec:
   - ../../pkg
   - ../../go.mod
   - ../../go.sum
+  - ../../tests/integration-tests/
```
hannesg, line 15: This dependency looks the wrong way round, doesn't it?

hannesg (approved):

services/cd-service/Buildfile.yaml
sven-urbanski-freiheit-com: I wanted to express `if integration-tests changed, then build all services`.

I would like to refactor this dependency handling more in story SRX-8HW1TP. I would leave it as it is for now.

PR #1094, opened 2023-11-03
Always add /metrics endpoint

hannesg:
This makes the /metrics endpoint available on all services and to all background tasks. I need to have this available in my next MR to register a metric there that is 0/1 for all background tasks depending on wether they are running.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -17,6 +17,7 @@ Copyright 2023 freiheit.com*/
 package metrics
 
 import (
+	"context"
 	"net/http"
 
 	"go.opentelemetry.io/otel/metric"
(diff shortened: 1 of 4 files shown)


pkg/metrics/metrics.go
```diff
@@ -41,3 +42,15 @@ func Init() (metric.MeterProvider, http.Handler, error) {
...
+
+type ctxKeyType struct{}
+
+var ctxKey ctxKeyType = ctxKeyType{}
+
+func FromContext(ctx context.Context) metric.MeterProvider {
+	return ctx.Value(ctxKey).(metric.MeterProvider)
+}
+
+func With(ctx context.Context, pv metric.MeterProvider) context.Context {
```
sven-urbanski-freiheit-com, line 54: Maybe call it this:
```suggestion
func WithProvider(ctx context.Context, pv metric.MeterProvider) context.Context {
```
?

sven-urbanski-freiheit-com (approved):

Issue #1043, opened 2023-10-20
Incosistent Coloring Scheme on Environment Groups


gmestre-freiheit:
When you have 2 chains of environment groups, the color scheme is inconsistent.
Example on our case:
- 1 chain for Alicloud Environments that gets the following colors:
   - Dev (Green)
   - Testing (Grey)
   - Staging (Yellow)
   - Production (Red)
- 1 chain for GCP Environments that gets the following colors:
   - Dev (Green)
   - Staging (Grey)
   - Production (Yellow)
   
We would expect that for the GCP ones we would get instead:
- Dev (Green)
- Staging (Yellow)
- Production (Red)

PR #1101, opened 2023-11-07
SRX-FZ6M63/#1043: fix colors for environments

bjoern-michaelsen-freiheit:
- make explicit the definition of "Priority" and their precedence
- change the precedence: UPSTREAM is now over PRE_PROD
- modify existing tests (slightly breaking backwards compatibility):
  * if the chain of environments contains loops, all environments are
    marked OTHER. This is likely for the better.
  * multiple environments with no upstream will all get marked the same
    way as PROD.
  * if there are two environments and one is the upstream of the other,
    then they are marked as Priority "UPSTREAM" and "PROD"

This PR intends to fix https://github.com/freiheit-com/kuberpult/issues/1043 .


bjoern-michaelsen-freiheit:
> * note this means if there are n > 1 environments between UPSTREAM and
>   PROD, only the last environment before PROD will be marked PRE_PROD

Do we really want that? Note that if we change this to all environments between UPSTREAM and PROD are PRE_PROD, then OTHER will almost never appear (AFAICS only with broken configs containing loops)

pkg/api/api.proto
```diff
@@ -218,9 +218,16 @@ message EnvironmentGroup {
 }
 
 enum Priority {
+  // PROD is an environment that no environment names as its upstream
+  // PROD takes precedence over UPSTREAM and PRE_PROD
+  // (e.g. if an environment is qualifies as both PROD and UPSTREAM, it shall be marked as PROD)
```
sven-urbanski-freiheit-com, line 223:
```suggestion
  // (e.g. if an environment qualifies as both PROD and UPSTREAM, it shall be marked as PROD)
```

services/cd-service/pkg/mapper/environments_config.go
```diff
@@ -152,22 +152,55 @@ func DeriveGroupName(env config.EnvironmentConfig, envName string) string {
...
+				var upstream = env.Config.Upstream.Environment
+				if upstream != nil {
+					var upstreamChildsBefore = childsByName[*upstream]
+					childsByName[*upstream] = append(upstreamChildsBefore, env.Name)
+				}
+			}
+		}
+	}
+	// remaining childless envs can now be identified as PROD
+	for i := 0; i < len(environments); i++ {
```
leandro-freiheit, line 180: I see no reason for having two for loops, I think they can be merged into one. Or is there a reason behind this decision? :)

services/cd-service/pkg/mapper/environments_config.go
```diff
@@ -152,22 +152,55 @@ func DeriveGroupName(env config.EnvironmentConfig, envName string) string {
...
-			env.Priority = api.Priority_PROD
-		} else if env.DistanceToUpstream == maxDistance-1 {
-			env.Priority = api.Priority_PRE_PROD
-		} else if env.DistanceToUpstream == 0 {
+		envsByName[env.Name] = env
+		if env.Config.Upstream.GetLatest() {
 			env.Priority = api.Priority_UPSTREAM
 		} else {
 			env.Priority = api.Priority_OTHER
+			// should we barf about a config that neither has is upstream nor has one?
```
sven-urbanski-freiheit-com, line 169: > // should we barf about a config that neither has is upstream nor has one?

No, we should not, at least not here.
It is an usual configuration, so we could consider adding this sort of check into `GetEnvironmentConfigsAndValidate`, which is only called on startup and stops the process in case of an error.

services/cd-service/pkg/mapper/environments_config.go
bjoern-michaelsen-freiheit: In the current implementation ... not. But I think we should want to have all envs between UPSTREAM and PROD to be PRE_PROD, and then it makes sense ...

bjoern-michaelsen-freiheit: After discussion: Third loop is removed.

sven-urbanski-freiheit-com (approved):

leandro-freiheit (approved):

PR #1109, opened 2023-11-08
Add a unified retry mechanism for background tasks

hannesg:
The new retry mechanism is implemented in the HealthReporter.Retry function. This function retries the function a few times with randomized backoff until the context is cancelled or the number of retries is exhausted. This is necessary because the grpc stream sometimes disconnect when their pods get restarted. Before this change the reconnect logic was reinplemented in multiple places and not consistent. This change adds a consistent method for that and introduces it to various jobs. This also gets rid of several warning logs.

### DIFF
--- a/pkg/setup/health.go
+++ b/pkg/setup/health.go
@@ -95,6 +95,29 @@ func (r *HealthReporter) ReportHealth(health Health, message string) {
 	}
 }
 
+// Retry allows background services to setup reliable streaming with backoff.
+//
+// This can be used to create background tasks that look like this:
+//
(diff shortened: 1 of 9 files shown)


pkg/setup/health.go
```diff
@@ -20,17 +20,23 @@ Copyright 2023 freiheit.com*/
 package setup
 
 import (
```
sven-urbanski-freiheit-com, line 22: typo in commit message:
> This function retries the function some times
-> This function retries the function several times

sven-urbanski-freiheit-com (approved):

PR #1120, opened 2023-11-10
Add filtering by team to the frontend-service

hannesg:
Follow up of https://github.com/freiheit-com/kuberpult/pull/1112

### DIFF
--- a/services/frontend-service/pkg/handler/handle_test.go
+++ b/services/frontend-service/pkg/handler/handle_test.go
@@ -952,6 +952,106 @@ func TestServer_Handle(t *testing.T) {
 	}
 }
 
+type mockRolloutClient struct {
+	api.RolloutServiceClient
+	request  *api.GetStatusRequest
+	response *api.GetStatusResponse
+}
+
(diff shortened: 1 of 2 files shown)


services/frontend-service/pkg/handler/handle_test.go
```diff
@@ -952,6 +952,106 @@ func TestServer_Handle(t *testing.T) {
...
+			req: &http.Request{
+				Method: http.MethodPost,
+				URL: &url.URL{
+					Path: "/environment-groups/development/rollout-status",
+				},
+				Header: http.Header{
+					"Content-Type": []string{"application/json"},
+				},
+				Body: io.NopCloser(strings.NewReader(`{}`)),
+			},
```
sven-urbanski-freiheit-com, line 989: nitpick:
I think it would be easier to read if instead of `req *http.Request`, you just define a `requestBody string`, and make it either empty or fill it with the team.

leandro-freiheit (approved):

services/frontend-service/pkg/handler/handle_test.go
hannesg: At the moment yes, however this test should be extended soon to cover more cases and then it's needed.

PR #1129, opened 2023-11-13
Fix make dependency (again)

sven-urbanski-freiheit-com:


Makefile
```diff
@@ -76,7 +76,7 @@ builder:
 	IMAGE_TAG=latest make -C infrastructure/docker/builder build
 
 kuberpult:
-	make -C pkg/api all
```
ahmed-nour-fdc, line 79: I think we need both of them. The frontend-service has a sim-link to the `pkg/api` directory.

Makefile
```diff
@@ -76,7 +76,7 @@ builder:
...
+	make -C services/frontend-service src/api/api.ts
```
jaypontes, line 79: Should we change the name of the `src/api/api.ts` target to a something like `build-protos`/`build-deps`?

ahmed-nour-fdc (approved):

PR #1131, opened 2023-11-13
UI: Lock without deploying

sven-urbanski-freiheit-com:
Adds a new button "Deploy only" that pops up when the user clicks the dropdown.
This makes it easier for users to deploy without creating a lock. But the default is still to create a lock with every manual deployment.

image

### DIFF
--- a/services/frontend-service/src/ui/components/button/ExpandButton.tsx
+++ b/services/frontend-service/src/ui/components/button/ExpandButton.tsx
@@ -24,13 +24,14 @@ import { PlainDialog } from '../dialog/ConfirmationDialog';
  * Displays one normal button on the left, and one arrow on the right to select a different option.
  */
 export const ExpandButton = (props: {
-    onClickSubmit: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+    onClickSubmit: (shouldLockToo: boolean) => void;
     defaultButtonLabel: string;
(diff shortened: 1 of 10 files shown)


jaypontes (approved):

services/frontend-service/src/ui/components/button/ExpandButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { Button } from './button';
+import { PlainDialog } from '../dialog/ConfirmationDialog';
+
+/**
+ * Two buttons combined into one.
+ * Inspired by GitHubs merge button.
+ * Displays one normal button on the left, and one arrow on the right to select a different option.
+ */
+export type ExpandButtonProps = {
+    onClickSubmit: (shouldLockToo: boolean) => void;
```
hannesg, line 27: I would have imagined a more generic solution but this actually works as well.

hannesg (approved):

services/frontend-service/src/ui/components/button/ExpandButton.tsx
sven-urbanski-freiheit-com: yeah, currently it's very specifically build to this one use case of "deploy / deploy&lock".
E.g. we also only support one button that "pops up". But we can extend this once it's needed.

Issue #1034, opened 2023-10-19
Inconsistent behaviour of filters


jefrww:
When I am on `ui/home` and select a team from `Teams`, I only see that teams applications.
When I do the same on `ui/locks` nothing happens

Expected behaviour:
- I only see that teams locks
- OR no option to filter by team

sven-urbanski-freiheit-com:
We ended up removing the teams option from the locks page. This was already implemented a while ago.

Issue #1033, opened 2023-10-19
Reset filter when going to /ui/locks


jefrww:
When you search for a specific service in the deployments, that filter persists even if you switch to `/ui/locks` where you most likely want to see all existing locks.
So a better UX probably would be, to clear the `Application Name` filter when switching pages.
In case someone wants to search the locks by application, they could still do that explicitly on `/ui/locks` afterwards.

sven-urbanski-freiheit-com:
I agree the current UX is not ideal.

We could for example:
* Keep the filter, but add a button to clear it more easily (and make the user aware of the filter), like "You are filtering for app X, reset?"
* Reset the filter every time we switch from locks to home or back, but allow filtering on the locks page still
* Store a separate query parameter for the filter in the locks page and the home page, so they both filter, but are not linked together like they are now.
* ...

PR #1133, opened 2023-11-14
SRC-K0RM01: improve UX for app and team filters (#1033, #1034)

bjoern-michaelsen-freiheit:
- move TopAppBar from outside main content router to inside it
- add props to TopAppBar to allow toggling the display of the app and
  team filters
- hide these filters on the locks page


leandro-freiheit (approved):

bjoern-michaelsen-freiheit:
Example with new colors:
kuberpult-colors
](url)

PR #1143, opened 2023-11-16
Implement waiting in the rollout service

hannesg:
The frontend part will come in a second MR. The configurable limit will also be implemented there

### DIFF
--- a/services/rollout-service/pkg/service/broadcast.go
+++ b/services/rollout-service/pkg/service/broadcast.go
@@ -90,7 +90,7 @@ type Broadcast struct {
 	mx       sync.Mutex
 	listener map[chan *BroadcastEvent]struct{}
 
-	// used for testing
+	// The waiting function is used in tests to trigger events after the subscription is set up.
 	waiting func()
 }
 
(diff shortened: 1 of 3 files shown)


services/rollout-service/pkg/service/broadcast.go
```diff
@@ -192,29 +196,85 @@ func (b *Broadcast) StreamStatus(req *api.StreamStatusRequest, svc api.RolloutSe
...
-			if s.RolloutStatus == api.RolloutStatus_RolloutStatusSuccesful {
-				continue
+		s := filterApplication(req, r)
+		if s != nil {
+			apps[r.Key] = s
+		}
+	}
+	status := aggregateStatus(apps)
+	if wait != nil {
+		// The waiting function is used in testing to make sure, we are really processing delayed events.
```
sven-urbanski-freiheit-com, line 214: The comment makes it sound a bit like waiting is only used in testing? Is that accurate? Maybe rephrase.

services/rollout-service/pkg/service/broadcast.go
```diff
@@ -192,29 +196,85 @@ func (b *Broadcast) StreamStatus(req *api.StreamStatusRequest, svc api.RolloutSe
...
+	}
+
 	return &api.GetStatusResponse{
 		Status:       status,
-		Applications: apps,
+		Applications: appList,
 	}, nil
 }
 
+// Removes irrelevant apps from
```
sven-urbanski-freiheit-com, line 254: from what?

services/rollout-service/pkg/service/broadcast_test.go
```diff
@@ -671,5 +765,33 @@ func TestGetStatus(t *testing.T) {
 			}
 		})
 
+		// This runs all test-cases again but delays all argoevents.
+		// The efect is that all apps will start as "unknown" and the will eventually converge.
```
sven-urbanski-freiheit-com, line 769:
```suggestion
		// The effect is that all apps will start as "unknown" and then will eventually converge.
```

sven-urbanski-freiheit-com (approved):

services/rollout-service/pkg/service/broadcast.go
hannesg: Yes, that is correct.

PR #1159, opened 2023-11-21
Kuberpult Page to Display the version of each service for a git tag

gsandok:
Simplified product summary endpoint to move logic to frontend. Display product summary based on environment in table format.

Screenshot 2023-11-23 at 1 49 34 PM

REV: DSN-FE1RG6


pkg/api/api.proto
```diff
@@ -33,6 +33,7 @@ message GetProductSummaryResponse {
 message ProductSummary {
   string app = 1;
   string version = 2;
+  string link_version = 3;
```
hannesg, line 36: Why not just put the source commit id and display version here just like we do with the normal releases? Then you wouldn't need to duplicate the logic in the backend.

gsandok: I was just trying to keep the logic out of the frontend because based on this story https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-NQPLdkfXUNc4bE9KOiO/STORY/-NdArQWgtNTXtGiJFTVH/ there is really only one value to display. If it makes sense I can add all values but also its just a preference if this logic would be in the front end or the backend

hannesg: It's a bit more than a preference. Currently this logic is only in the frontend ( see https://github.com/freiheit-com/kuberpult/blob/ce320ba275b3902d0cc3e42afc369025535a843f/services/frontend-service/src/ui/components/ReleaseVersion/ReleaseVersion.tsx#L31 ). It's not too bad to duplicate this, but I think it will make your life easier to not do that and put it in the frontend in both cases.

gsandok: fixed and now all logic is dealt with in the front end

services/cd-service/pkg/service/tags.go
```diff
@@ -68,15 +68,12 @@ func (s *TagsServer) GetProductSummary(ctx context.Context, in *api.GetProductSu
...
 			if row.App == app.Name {
-				length := len(app.Releases) - 1
-				if app.Releases[length].DisplayVersion != "" {
-					productVersion = append(productVersion, &api.ProductSummary{App: row.App, Version: app.Releases[length].DisplayVersion})
-				} else if app.Releases[length].SourceCommitId != "" {
-					productVersion = append(productVersion, &api.ProductSummary{App: row.App, Version: app.Releases[length].SourceCommitId})
-				} else {
-					productVersion = append(productVersion, &api.ProductSummary{App: row.App, Version: row.Version})
+				for _, release := range app.Releases {
+					if strconv.Itoa(int(release.Version)) == row.Version {
```
sven-urbanski-freiheit-com, line 72: this typecast should be avoided because we are loosing information. See the implementation of strconv.Itoa. 
```suggestion
					if strconv.FormatUint(release.Version, 10) == row.Version {
```

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
```diff
@@ -15,40 +15,108 @@ along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>
...
+        setShowSpinner(false);
     }, []);
     const { environment } = props;
     const [open, setOpen] = React.useState(false);
-    const openClose = React.useCallback(() => {
-        setOpen(!open);
-    }, [open, setOpen]);
+    const openClose = React.useCallback(
+        (e: React.ChangeEvent<HTMLSelectElement>) => {
+            setShowSpinner(true);
```
sven-urbanski-freiheit-com, line 37: I cannot test it right now, but I doubt that this has the intended effect, because `getSummary` does not wait for the response (e.g. it's not returning a Promise). 
Did you try to enable network throttling in chrome to test this (manually)?

For comparison, we are doing something similar in the function `useFrontendConfig` with the `configReady` flag.

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
```diff
@@ -15,40 +15,108 @@ along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>
...
-                {
-                    'This page shows the version of the product for the selected environment based on tags to the repository. If there are no tags, then no data can be shown.'
-                }
+            {showSpinner && <Spinner message="Loading Tag Data" />}
+            <div>
+                {displaySummary ? (
+                    <div className="table_padding">
+                        <table className="table">
+                            <tr className="table_title">
+                                <th>App/Service Name</th>
```
sven-urbanski-freiheit-com, line 90:
```suggestion
                                <th>App Name</th>
```

services/cd-service/pkg/service/tags.go
sven-urbanski-freiheit-com: (not crucial because it's unlike that we have more than ~2^32 releases, but should be fixed anyway)

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
sven-urbanski-freiheit-com: I tested it locally, the spinner does not show up for me.

services/cd-service/pkg/service/tags.go
```diff
@@ -68,15 +68,12 @@ func (s *TagsServer) GetProductSummary(ctx context.Context, in *api.GetProductSu
...
+						productVersion = append(productVersion, &api.ProductSummary{App: row.App, Version: row.Version, CommitId: release.SourceCommitId, DisplayVersion: release.DisplayVersion})
+						break
+					}
 				}
-				break
 			}
 		}
 	}
```
sven-urbanski-freiheit-com, line 79: please sort the productionVersions, e.g. by app name, e.g. `sort.Strings(...)` or `sort.Sort(...)`.
As of now the order is undefined.

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
sven-urbanski-freiheit-com: (to be precise, the spinner does show up, but it's not the right one. It never says "loading tag data")

sven-urbanski-freiheit-com (dismissed):
.

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
gsandok: Fixed and tested with network throttling

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
```diff
@@ -15,40 +15,120 @@ along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>
...
+        return app.version;
+    };
+
+    const tagsResponse = useTags();
+    const [displaySummary, setDisplayVersion] = React.useState(false);
+
+    React.useEffect(() => {
+        if (tagsResponse.response.tagData.length > 0) {
+            setShowSummarySpinner(true);
+            getSummary(tagsResponse.response.tagData[0].commitId, environment);
```
sven-urbanski-freiheit-com, line 61: why are we calling `getSummary` twice?

hannesg (approved):

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
gsandok: two separate instances, one is for when a user selects an option in the dropdown and the other is when the user first reaches the product version page so that results are loaded

PR #1168, opened 2023-11-24
Fail health check when too many git pushes fail

sven-urbanski-freiheit-com:
From real life experience, it seems there are some cases where a restart of the cd-service helped to fix a hanging git push processes.

This automates the restart by adding the number of failures to the health check. We restart when we see more than 3 git push errors.

Reference: SRX-976IRJ

### DIFF
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -67,6 +67,7 @@ type Repository interface {
 	State() *State
 	StateAt(oid *git.Oid) (*State, error)
 	Notify() *notify.Notify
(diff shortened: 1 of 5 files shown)


services/cd-service/pkg/repository/repository.go
```diff
@@ -1098,6 +1114,21 @@ func (r *repository) Notify() *notify.Notify {
 	return &r.notify
 }
 
+func (r *repository) HealthCheckBackgroundJob(ctx context.Context, reporter *setup.HealthReporter) {
```
sven-urbanski-freiheit-com, line 1117: tests are still missing

PR #1183, opened 2023-12-01
feat: Support longer application names

### LINKED ISSUE #1182: Allow App names with length > 39
Hi,
I noted that there's an arbitrary length restriction on the length of the application name to 39 characters.
In our application we have microservices which have a name with length of 40-45 characters.
And who knows whether they get longer in the future.
I don't think it makes sense to trim a character of every service for just this application, as this could also impose `uniqueness` issues.

The change was introduced with #814.
From Argo-CD side it looks like this is possible to have quite long application names, so this should not conflict:
https://github.com/argoproj/argo-cd/pull/7468

Proposal:
So maybe just add 31 more characters as a buffer.

jdvgh:
Closes #1182

### DIFF
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -67,6 +67,12 @@ log:
   level: "WARN"
 cd:
   image: kuberpult-cd-service
+# By default, kuberpult allows names of apps to be max 39 characters.
+# This is usually enough. If you want to extend the length to up to 70 characters,
+# You need to ensure that ArgoCD has no issue with this. Set the Argo CD helm parameter:
+# `configs.cm.application.resourceTrackingMethod: annotation+label`
+# Note that there is no simple way in kuberpult to rename an app.
+  allowLongAppNames: false
   backendConfig:
     create: false   # Add backend config for health checks on GKE only
     timeoutSec: 300  # 30sec is the default on gcp loadbalancers, however kuberpult needs more with parallel requests. It is the time how long the loadbalancer waits for kuberpult to finish calls to the rest endpoint "release"
--- a/docker-compose-earthly.yml
+++ b/docker-compose-earthly.yml
@@ -10,6 +10,7 @@ services:
       - KUBERPULT_DEX_ENABLED=false
       - KUBERPULT_GIT_WRITE_COMMIT_DATA=true
       - KUBERPULT_MAXIMUM_QUEUE_SIZE=2
+      - KUBERPULT_ALLOW_LONG_APP_NAMES=true
     ports:
       - "8080:8080"
       - "8443:8443"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,7 @@ services:
       - KUBERPULT_DEX_ENABLED=false
       - KUBERPULT_GIT_NETWORK_TIMEOUT=3s
       - KUBERPULT_GIT_WRITE_COMMIT_DATA=true
+      - KUBERPULT_ALLOW_LONG_APP_NAMES=true
     ports:
       - "8080:8080"
       - "8443:8443"
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -193,14 +193,14 @@ type RepositoryConfig struct {
 	// if set, kuberpult will generate push events to argoCd whenever it writes to the manifest repo:
 	ArgoWebhookUrl string
 	// the url to the git repo, like the browser requires it (https protocol)
-	WebURL          string
-	DogstatsdEvents bool
-	WriteCommitData bool
-	WebhookResolver WebhookResolver
-
+	WebURL                string
+	DogstatsdEvents       bool
+	WriteCommitData       bool
+	WebhookResolver       WebhookResolver
 	MaximumCommitsPerPush uint
-
-	MaximumQueueSize uint
+	MaximumQueueSize      uint
+	// Extend maximum AppName length
+	AllowLongAppNames bool
 }
 
 func openOrCreate(path string, storageBackend StorageBackend) (*git.Repository, error) {
--- a/charts/kuberpult/templates/cd-service.yaml
+++ b/charts/kuberpult/templates/cd-service.yaml
@@ -193,6 +193,8 @@ spec:
           value: "{{ .Values.datadogProfiling.enabled }}"
         - name: KUBERPULT_MAXIMUM_QUEUE_SIZE
           value: "{{ .Values.cd.backendConfig.queueSize }}"
+        - name: KUBERPULT_ALLOW_LONG_APP_NAMES
+          value: "{{ .Values.cd.allowLongAppNames }}"
         volumeMounts:
         - name: repository
           mountPath: /repository
--- a/pkg/valid/validations.go
+++ b/pkg/valid/validations.go
@@ -17,17 +17,20 @@ Copyright 2023 freiheit.com*/
 package valid
 
 import (
+	"os"
 	"regexp"
 	"strings"
 )
 
 const (
-	MaxAppNameLen        = 39
-	AppNameRegExp        = `\A[a-z0-9]+(?:-[a-z0-9]+)*\z`
-	TeamNameRegExp       = AppNameRegExp
-	EnvNameRegExp        = AppNameRegExp
-	SHA1CommitIDLength   = 40
-	commitIDPrefixRegExp = `^[0-9a-fA-F]*$`
+	FallbackMaxAppNameLength       = 39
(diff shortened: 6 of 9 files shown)


services/cd-service/pkg/valid/validations.go
```diff
@@ -22,7 +22,7 @@ import (
 )
 
 const (
-	MaxAppNameLen  = 39
+	MaxAppNameLen  = 70
```
sven-urbanski-freiheit-com, line 25: We have to double check this.
The app name is also sent to argoCd via CRD and I vaguely remember we did this because of a technical limitations there. (maybe dns)

sven-urbanski-freiheit-com, line 25: also we have to check if this breaks the UI layout in any way.

ahmed-nour-fdc: [This](https://github.com/argoproj/argo-cd/issues/5595) specifies that the limit for app names is 63 characters.

hannesg: Yes, you can put 70 here but then argocd breaks.

jdvgh:
Sorry for the late feedback.
Yes, out-of-the-box ArgoCD has issues with application names > 63 chars:
image`

However, if the tracking method is changed e.g. if we use in the Helm chart:
```
configs:
  cm:
    application.resourceTrackingMethod: annotation+label
```
image
or `annotation`
```
configs:
  cm:
    application.resourceTrackingMethod: annotation
```
image

we do not get this issue.

services/cd-service/pkg/valid/validations.go
jdvgh: @ahmed-nour-fdc exactly - also see my comment to the PR. I tested it in my own instance.

jdvgh: I tested it with the application name `myappiswaytoolongdontyouthinkisnitormaybeitisnotmaybewejusthaveareally`


image

sven-urbanski-freiheit-com: How does it look when adding an action (like deploy or lock) to the planned actions? This should also render nice.

jdvgh: image
image
image
image


well it is not beautiful, as it is not trimmed anywhere - I'll have a look if I see some low hangig fruits to adapt in the UI

jdvgh: okay, i'll adapt it next week. Sorry for this torn apart review

jdvgh: @sven-urbanski-freiheit-com I updated it to use `KUBERPULT_MAX_APP_NAME_LENGTH` env variable.
As this is used inside the `validation.go` I had to somehow read the env there.
If you can think of a better approach - go ahead and sketch it and I'll implement it :)

services/cd-service/pkg/repository/repository.go
```diff
@@ -137,6 +137,8 @@ type RepositoryConfig struct {
 	ArgoWebhookUrl string
 	// the url to the git repo, like the browser requires it (https protocol)
 	WebURL string
+	// maximum length of kuberpult application name to be used
+	MaxAppNameLength int
```
jdvgh, line 141: so we also have it available in the config - if needed

services/cd-service/pkg/valid/validations.go
```diff
@@ -17,23 +17,39 @@ Copyright 2023 freiheit.com*/
...
 	applicationNameRx = regexp.MustCompile(AppNameRegExp)
 	teamNameRx        = regexp.MustCompile(TeamNameRegExp)
 	envNameRx         = regexp.MustCompile(EnvNameRegExp)
+	MaxAppNameLen     = setupMaxAppNameLen()
 )
 
+func setupMaxAppNameLen() int {
+	maxAppNameLength := FallbackMaxAppNameLength
+	res, ok := os.LookupEnv(KUBERPULT_MAX_APP_NAME_LENGTH)
+	if ok {
```
sven-urbanski-freiheit-com, line 44: to change it to `boolean` (see comment above), we also need to adapt this and some other places.
```suggestion
	res, ok := os.LookupEnv(KUBERPULT_MAX_APP_NAME_LENGTH)
	if ok && res == "true" {
```

services/cd-service/pkg/valid/validations.go
sven-urbanski-freiheit-com: Overall this looks ok, just the "planned actions" need to be adapted. There should not be a new scrollbar.

charts/kuberpult/values.yaml.tpl
```diff
@@ -46,6 +46,7 @@ cd:
 # In general, kuberpult only guarantees that running with the same version of frontend and cd service will work.
 # For testing purposes, we allow to overwrite the tags individually, to test an old frontend service with a new cd service.
   tag: "$VERSION"
+  maxAppNameLength: 39 # How long are kuberpult application names allowed to be
```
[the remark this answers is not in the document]
jdvgh: adapted

charts/kuberpult/values.yaml
```diff
@@ -61,6 +61,16 @@ log:
...
+# In MOST cases, do NOT overwrite the "tag".
+# In general, kuberpult only guarantees that running with the same version of frontend and cd service will work.
+# For testing purposes, we allow to overwrite the tags individually, to test an old frontend service with a new cd service.
+  tag: "$VERSION"
+# By default, kuberpult allows names of apps to be max 39 characters.
+# This is usually enough. If you want to extend the length to up to 70 characters,
+# You need to ensure that ArgoCD has no issue with this. Set the Argo CD helm parameter:
+# `configs.cm.application.resourceTrackingMethod: annotation+label`
+# Note that there is no simple way in kuberpult to rename an app.
+  allowLongAppNames: true
```
sven-urbanski-freiheit-com, line 73: The default should be false.
```suggestion
  allowLongAppNames: false
```

docker-compose-earthly.yml
```diff
@@ -9,6 +9,7 @@ services:
       - KUBERPULT_DEX_MOCK=false
       - KUBERPULT_DEX_ENABLED=false
       - KUBERPULT_GIT_WRITE_COMMIT_DATA=true
+      - KUBERPULT_ALLOW_LONG_APP_NAMES=false
```
sven-urbanski-freiheit-com, line 12:
```suggestion
      - KUBERPULT_ALLOW_LONG_APP_NAMES=true
```
generally, all feature flags should be enable for local development.

docker-compose.yml
```diff
@@ -15,6 +15,7 @@ services:
       - KUBERPULT_DEX_ENABLED=false
       - KUBERPULT_GIT_NETWORK_TIMEOUT=3s
       - KUBERPULT_GIT_WRITE_COMMIT_DATA=true
+      - ALLOW_LONG_APP_NAMES=false
```
sven-urbanski-freiheit-com, line 18:
```suggestion
      - KUBERPULT_ALLOW_LONG_APP_NAMES=false
```

charts/kuberpult/values.yaml
```diff
@@ -61,6 +61,16 @@ log:
   level: "WARN"
 cd:
   image: kuberpult-cd-service
+# In MOST cases, do NOT overwrite the "tag".
+# In general, kuberpult only guarantees that running with the same version of frontend and cd service will work.
+# For testing purposes, we allow to overwrite the tags individually, to test an old frontend service with a new cd service.
+  tag: "$VERSION"
```
sven-urbanski-freiheit-com, line 67:
```suggestion
```
This should be removed, looks like a git rebase conflict to me.

services/cd-service/pkg/valid/validations.go
sven-urbanski-freiheit-com: There is still a scrollbar in the planned actions:
image

Instead it should automatically add newlines. Maybe `word-wrap:break-word;` is enough to achieve that. @jdvgh

charts/kuberpult/values.yaml
jdvgh: updated

docker-compose-earthly.yml
jdvgh: updated

docker-compose.yml
jdvgh: changed but set to true, or should it be only true for the `docker-compose-earthly.yml`? @sven-urbanski-freiheit-com

services/cd-service/pkg/valid/validations.go
jdvgh: Got it to work with `word-break: break-word`
image

PR #1193, opened 2023-12-05
Restart pod when a queued action takes longer than 3*networkTimeout

sven-urbanski-freiheit-com:
This changes the cd-service so that it automatically exits when a queued action (like a release train), takes longer than 3 times the configured networkTimeout.

The problem is that sometimes "git push" actions take a long time (and every writing action needs to run git push). They cannot be run in parallel, and they can not properly be cancelled either. So as a last resort, we just exit, and let kubernetes restart the pod.


sven-urbanski-freiheit-com:
this replaces https://github.com/freiheit-com/kuberpult/pull/1184

services/cd-service/pkg/repository/repository_test.go
```diff
@@ -1022,6 +1023,15 @@ func (s *SlowTransformer) Transform(ctx context.Context, state *State) (string,
 	return "ok", &TransformerResult{}, nil
 }
 
+type SleepTransformer struct {
+	duration time.Duration
+}
+
+func (s *SleepTransformer) Transform(ctx context.Context, state *State) (string, *TransformerResult, error) {
+	time.Sleep(s.duration)
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: I was able to remove `sleep` from the test - there is still one case where it's required, but I think it's reasonable. Please check again @hannesg

pkg/setup/health.go
```diff
@@ -153,6 +168,7 @@ type HealthServer struct {
 	parts          map[string]report
 	mx             sync.Mutex
 	BackOffFactory func() backoff.BackOff
+	Clock          func() time.Time
 }
 
 func (h *HealthServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
sven-urbanski-freiheit-com, line 174: i think we need to consider the ttl here as well

leandro-freiheit (approved):

pkg/setup/health.go
hannesg: Yes, we do. We can do that pretty well in the reports method.

hannesg (commented):
In order to make this work, we would also need to get rid of the timeout in useRemote but we can already merge this now becuase it's an improvement already.

pkg/setup/health.go
sven-urbanski-freiheit-com: done

PR #1201, opened 2023-12-07
feat: validate the manifest is yaml and not changed (SRX-627DCD)

bjoern-michaelsen-freiheit:


services/cd-service/pkg/repository/transformer.go
```diff
@@ -329,13 +330,66 @@ func (c *CreateApplicationVersion) calculateVersion(bfs billy.Filesystem) (uint6
...
+		}
+	}
+	if c.SourceRepoUrl != "" {
+		existingSourceRepoUrl, err := util.ReadFile(fs, fs.Join(releaseDir, "source_repo_url"))
+		if err != nil || string(existingSourceRepoUrl) != c.SourceRepoUrl {
+			return false
+		}
+	}
+	for env, man := range c.Manifests {
+		envDir := fs.Join(releaseDir, "environments", env)
```
bjoern-michaelsen-freiheit, line 384: Should we compare canonical serializations instead?

sven-urbanski-freiheit-com: I think we should compare after "normalization by serialization".

But more importantly: if there is a diff, then we should give the user the diff (or at least the line where they diff first), so they can understand the issue - ideally in the http response.

services/frontend-service/pkg/handler/release.go
```diff
@@ -64,6 +64,17 @@ func isAuthor(value string) bool {
...
 
+func writeReleaseResponse(w http.ResponseWriter, r *http.Request, jsonBlob []byte, err error, status int) {
+	ctx := r.Context()
+	if err != nil {
+		logger.FromContext(ctx).Error(fmt.Sprintf("error in json.Marshal of /release: %s", err.Error()))
+		http.Error(w, "", http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(status)
+	w.Write(jsonBlob)
```
sven-urbanski-freiheit-com, line 75:
```suggestion
	w.Write(jsonBlob + "\n")
```

pkg/api/api.proto
```diff
@@ -132,7 +132,42 @@ message CreateReleaseRequest {
   string displayVersion = 10;
 }
 
+message CreateReleaseResponseSuccess {
+}
```
sven-urbanski-freiheit-com, line 136: overall fine, but actual "diff" is missing still

services/cd-service/pkg/repository/transformer.go
bjoern-michaelsen-freiheit: done

pkg/api/api.proto
bjoern-michaelsen-freiheit: success never has a diff (because no version existed previously in that case).

sven-urbanski-freiheit-com (approved):

PR #1213, opened 2023-12-08
Move Disconnect to the go func

hannesg:
The disconnect might get triggered while the push is still running and then segfaults sometimes.

### DIFF
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -423,6 +423,9 @@ func (r *repository) useRemote(ctx context.Context, callback func(*git.Remote) e
 	defer cancel()
 	errCh := make(chan error, 1)
 	go func() {
+                // Usually we call `defer` right after resource allocation (`CreateAnonymous`).
+                // The issue with that is that the `callback` requires the remote, and cannot be cancelled properly.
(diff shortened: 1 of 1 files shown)


services/cd-service/pkg/repository/repository.go
```diff
@@ -419,11 +419,11 @@ func (r *repository) useRemote(ctx context.Context, callback func(*git.Remote) e
 	if err != nil {
 		return fmt.Errorf("opening remote %q: %w", r.config.URL, err)
 	}
-	defer remote.Disconnect()
 	ctx, cancel := context.WithTimeout(context.Background(), r.config.NetworkTimeout)
 	defer cancel()
 	errCh := make(chan error, 1)
 	go func() {
+		defer remote.Disconnect()
```
sven-urbanski-freiheit-com, line 426: please comment this, it's not obvious. e.g.
```suggestion
                // Usually we call `defer` right after resource allocation (`CreateAnonymous`).
                // The issue with that is that the `callback` requires the remote, and cannot be cancelled properly.
                // So `callback` may run longer than `useRemote`, and if at that point `Disconnect` was already called, we get a `panic`.
		defer remote.Disconnect()
```

sven-urbanski-freiheit-com (approved):

PR #1240, opened 2023-12-13
perf: Add a new simpler protocol for calculating versions in the rollout service

hannesg:
Currently this relies on the much heavier GetOverview endpoint which is tricky because it slows down the rollout service in wawi. This new endpoint contains exactly what we need and nothing else.

### DIFF
--- a/pkg/api/api.proto
+++ b/pkg/api/api.proto
@@ -213,6 +213,20 @@ message GetFrontendConfigResponse {
   string manifestRepoUrl = 6;
 }
 
+message GetVersionRequest {
+  string git_revision = 1;
+  string application = 2;
+  string environment = 3;
+}
+
+message GetVersionResponse {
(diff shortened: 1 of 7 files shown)


leandro-freiheit (approved):

pkg/api/api.proto
```diff
@@ -213,6 +213,20 @@ message GetFrontendConfigResponse {
...
+  string environment = 3;
+}
+
+message GetVersionResponse {
+  uint64 version = 1;
+  google.protobuf.Timestamp deployed_at = 2;
+}
+
+service VersionService {
+  rpc GetVersion (GetVersionRequest) returns (GetVersionResponse) {}
```
sven-urbanski-freiheit-com, line 228: this looks similar to `GetProductSummaryRequest`, can we unify the two?

PR #1247, opened 2023-12-18
feat: Use new dedicated version protocol

hannesg:
This switches the rollout services internal client to use the new endpoint instead of relying on the overview endpoint


services/rollout-service/pkg/cmd/server.go
```diff
@@ -105,7 +105,7 @@ func RunServer() {
 	}
 }
 
-func getGrpcClient(ctx context.Context, config Config) (api.OverviewServiceClient, error) {
+func getGrpcClient(ctx context.Context, config Config) (api.OverviewServiceClient, api.VersionServiceClient, error) {
```
sven-urbanski-freiheit-com, line 108:
```suggestion
func getGrpcClients(ctx context.Context, config Config) (api.OverviewServiceClient, api.VersionServiceClient, error) {
```

sven-urbanski-freiheit-com (approved):

leandro-freiheit (approved):

services/rollout-service/pkg/cmd/server.go
hannesg: Done

PR #1260, opened 2023-12-22
fix: add missing author for new version endpoint

hannesg:
The author is missing for this endpoint.

### DIFF
--- a/services/rollout-service/pkg/versions/versions_test.go
+++ b/services/rollout-service/pkg/versions/versions_test.go
@@ -45,14 +45,15 @@ type step struct {
 }
 
 type expectedVersion struct {
-	Revision        string
-	Environment     string
-	Application     string
-	DeployedVersion uint64
-	DeployTime      time.Time
-	SourceCommitId  string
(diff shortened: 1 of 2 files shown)


leandro-freiheit (approved):

services/rollout-service/pkg/versions/versions_test.go
```diff
@@ -45,14 +45,15 @@ type step struct {
...
-	SourceCommitId  string
-	Metadata        metadata.MD
-	IsProduction    bool
+	Revision         string
+	Environment      string
+	Application      string
+	DeployedVersion  uint64
+	DeployTime       time.Time
+	SourceCommitId   string
+	OverviewMetadata metadata.MD
```
sven-urbanski-freiheit-com, line 54: it looks like you're only using `OverviewMetadata` and `VersionMetadata` in the tests. Sounds a bit like a changed file was not pushed?

hannesg: No, it is actually checked in the test. However, my change made it always empty

PR #1261, opened 2023-12-22
feat: create argo apps ourselves

leandro-freiheit:


services/cd-service/pkg/repository/repository.go
```diff
@@ -1194,6 +1196,8 @@ type State struct {
 	Commit                 *git.Commit
 	BootstrapMode          bool
 	EnvironmentConfigsPath string
+	RepoURL                string
```
hannesg, line 1199: You don't need this anymore.

services/rollout-service/pkg/argo/argo.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+}
+
+func (a *ArgoAppProcessor) Consume(ctx context.Context) error {
+	seen := map[Key]*api.Environment_Application{}
+
+	for {
+		select {
+		case <-a.trigger:
+			err := a.ConsumeArgo(ctx, a.HealthReporter)
```
hannesg, line 63: You can't call this here like this as `ConsumeArgo` might never finish.

services/rollout-service/pkg/argo/argo_test.go
```diff
@@ -0,0 +1,444 @@
...
+			hlth := &setup.HealthServer{}
+			argoProcessor := &mockArgoProcessor{
+				lastOverview:      tc.Overview,
+				ApplicationClient: as,
+				HealthReporter:    hlth.Reporter("consume-argo"),
+				trigger:           make(chan *v1alpha1.Application, 1),
+				argoApps:          make(chan *v1alpha1.ApplicationWatchEvent, 2),
+			}
+			hlth.BackOffFactory = func() backoff.BackOff { return backoff.NewConstantBackOff(0) }
+			go argoProcessor.Consume(ctx)
```
hannesg, line 408: Please also test that this actually shuts down.

services/rollout-service/pkg/versions/versions.go
```diff
@@ -218,21 +234,53 @@ func (v *versionClient) ConsumeEvents(ctx context.Context, processor VersionEven
...
+							annotations["com.freiheit.kuberpult/self-managed-filter"] = v.manageArgoAppsFilter
+							// This annotation is so that argoCd does not invalidate *everything* in the whole repo when receiving a git webhook.
+							// It has to start with a "/" to be absolute to the git repo.
+							// See https://argo-cd.readthedocs.io/en/stable/operator-manual/high_availability/#webhook-and-manifest-paths-annotation
+							annotations["argocd.argoproj.io/manifest-generate-paths"] = "/" + manifestPath
+							labels["com.freiheit.kuberpult/team"] = tm
+
+							deployApp := argo.CreateDeployApplication(overview, app, env, annotations, labels, manifestPath)
+
+							v.ArgoProcessor.Push(overview, deployApp)
```
hannesg, line 260: You only need to push the overview here. Also you are pushing for every application which is definitely too much.

services/rollout-service/pkg/versions/versions.go
```diff
@@ -218,21 +234,53 @@ func (v *versionClient) ConsumeEvents(ctx context.Context, processor VersionEven
...
+
+						} else {
+							processor.ProcessKuberpultEvent(ctx, KuberpultEvent{
```
hannesg, line 263: Why is that in an else clause now? That call is needed even for managed argo applications.

services/rollout-service/pkg/argocd/render.go
hannesg, line 1: This file is unused, right?

services/rollout-service/pkg/versions/versions.go
```diff
@@ -218,21 +234,53 @@ func (v *versionClient) ConsumeEvents(ctx context.Context, processor VersionEven
...
-							EnvironmentGroup: envGroup.EnvironmentGroupName,
-							Team:             tm,
-							IsProduction:     env.Priority == api.Priority_PROD,
-							Version: &VersionInfo{
-								Version:        app.Version,
-								SourceCommitId: sc,
-								DeployedAt:     dt,
-							},
-						})
+						if v.manageArgoAppsEnabled && len(v.manageArgoAppsFilter) > 0 && strings.Contains(v.manageArgoAppsFilter, app.Name) {
```
hannesg, line 237: The simple string filter here is probably a bit too simple. Also, I think this would be much better in the ArgoProcessor.

services/rollout-service/pkg/versions/versions.go
```diff
@@ -218,21 +234,53 @@ func (v *versionClient) ConsumeEvents(ctx context.Context, processor VersionEven
...
+							manifestPath := path.Join("environments", env.Name, "applications", app.Name, "manifests", "manifests.yaml")
```
hannesg, line 238: You have already most of the code that generates the applications in rollout-service/pkg/argo. I would move this code there as well for simplicity.

services/rollout-service/pkg/argo/argo_test.go
```diff
@@ -0,0 +1,534 @@
...
+				lastOverview:      tc.Overview,
+				ApplicationClient: as,
+				trigger:           make(chan *api.GetOverviewResponse, 10),
+				argoApps:          make(chan *v1alpha1.ApplicationWatchEvent, 10),
+			}
+			hlth.BackOffFactory = func() backoff.BackOff { return backoff.NewConstantBackOff(0) }
+			go argoProcessor.Consume(t, ctx)
+			go argoProcessor.ConsumeArgo(ctx, hlth.Reporter("consume-argo"))
+			argoProcessor.Push(ctx, tc.Overview)
+			//We add a delay so that all the events are reported by the application client
```
leandro-freiheit, line 485: I force a delay here to make sure the events created from the apps are matched before the consumed events test function is called

charts/kuberpult/values.yaml.tpl
```diff
@@ -264,3 +264,8 @@ revolution:
     token: ""
     # Maximum number of requests send in parallel.
     concurrency: 20
+
+# Whether the rollout service should self-manage applications
+manageArgoApplications:
+  enabled: false
+  filter: []
```
sven-urbanski-freiheit-com, line 271:
```suggestion
# list of apps that should be self managed. App names must be as they appear in Argo CD (usually "<microservicename>- + <environment>"
  filter: []
```

services/rollout-service/pkg/argo/argo.go
```diff
@@ -0,0 +1,341 @@
...
+			for _, envGroup := range overview.EnvironmentGroups {
+				for _, env := range envGroup.Environments {
+					appList, err := a.ApplicationClient.List(ctx, &application.ApplicationQuery{
+						Name: ptr.FromString(fmt.Sprintf("%s-", env)),
+					})
+
+					if err != nil {
+						fmt.Errorf("listing applications: %w", err)
+					}
+					err = a.DeleteArgoApps(ctx, appList, env.Applications)
```
sven-urbanski-freiheit-com, line 109: It's a bit counter-intuitive to call "delete" when "list" didn't work.
How about this:
```suggestion
						fmt.Errorf("listing applications: %w", err)
						continue
					}
					err = a.DeleteArgoApps(ctx, appList, env.Applications)
```

services/rollout-service/pkg/argo/argo.go
```diff
@@ -0,0 +1,341 @@
...
+
+					if err != nil {
+						fmt.Errorf("deleting applications: %w", err)
```
sven-urbanski-freiheit-com, line 112:
```suggestion
						fmt.Errorf("deleting applications: %w", err)
						continue;
```

services/rollout-service/pkg/argo/argo.go
```diff
@@ -0,0 +1,341 @@
...
+
+func (a *ArgoAppProcessor) Consume(ctx context.Context) error {
+	l := logger.FromContext(ctx).With(zap.String("event-consuming", "ready"))
+
+	for {
+		select {
+		case overview := <-a.trigger:
+			for _, envGroup := range overview.EnvironmentGroups {
+				for _, env := range envGroup.Environments {
+					appList, err := a.ApplicationClient.List(ctx, &application.ApplicationQuery{
```
sven-urbanski-freiheit-com, line 102: maybe rename
```suggestion
					appsKnownToArgo, err := a.ApplicationClient.List(ctx, &application.ApplicationQuery{
```

services/rollout-service/pkg/argo/argo.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+}
+
+func calculateFinalizers() []string {
+	return []string{
+		"resources-finalizer.argocd.argoproj.io",
+	}
+}
+
+func (a ArgoAppProcessor) DeleteArgoApps(ctx context.Context, appList *v1alpha1.ApplicationList, apps map[string]*api.Environment_Application) error {
```
sven-urbanski-freiheit-com, line 216: maybe rename
```suggestion
func (a ArgoAppProcessor) DeleteArgoApps(ctx context.Context, appsKnownToArgo *v1alpha1.ApplicationList, appsKnownToKuberpult map[string]*api.Environment_Application) error {
```

services/rollout-service/pkg/argo/argo.go
```diff
@@ -0,0 +1,341 @@
...
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func CreateDeployApplication(overview *api.GetOverviewResponse, app *api.Environment_Application, env *api.Environment) *v1alpha1.Application {
```
sven-urbanski-freiheit-com, line 241:
```suggestion
func CreateArgoApplication(overview *api.GetOverviewResponse, app *api.Environment_Application, env *api.Environment) *v1alpha1.Application {
```

services/rollout-service/pkg/argo/argo.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+}
+
+func (a ArgoAppProcessor) CreateOrUpdateApp(ctx context.Context, overview *api.GetOverviewResponse, app *api.Environment_Application, env *api.Environment) {
+	t := team(overview, app.Name)
+	if a.ManageArgoAppsEnabled && len(a.ManageArgoAppsFilter) > 0 && slices.Contains(a.ManageArgoAppsFilter, t) {
+		k := Key{AppName: app.Name, EnvName: env.Name, Application: app, Environment: env}
+
+		argoApp, err := a.ApplicationClient.Get(ctx, &application.ApplicationQuery{
+			Name: ptr.FromString(app.Name),
```
hannesg, line 149: The name is definitely  wrong here as it doesn't contain the environment. Also, why do you even retrieve the application? You did just run a listing operation in the other function which should have given you the exact same object if it exists. Also, when you do the create later,  you still  need to handle the case that the app was already created in the meantime gracefully.

services/rollout-service/pkg/argo/argo.go
```diff
@@ -0,0 +1,341 @@
...
+		})
+
+		if err != nil {
+			fmt.Errorf("getting application: %w", err)
```
hannesg, line 153: I think you want to use `logger.FromContext(ctx).Error`

charts/kuberpult/values.yaml.tpl
```diff
@@ -264,3 +264,9 @@ revolution:
     token: ""
     # Maximum number of requests send in parallel.
     concurrency: 20
+
+# Whether the rollout service should self-manage applications
+manageArgoApplications:
+  enabled: false
+  # List of teams that should be self managed by the rollout service. The apps that belong to the teams in the filter.
```
sven-urbanski-freiheit-com, line 271: either remove this sentence or rephrase it.

```suggestion
  # List of teams that should be self managed by the rollout service.
```

sven-urbanski-freiheit-com (approved):

services/rollout-service/pkg/argo/argo.go
```diff
@@ -0,0 +1,341 @@
...
+func (a *ArgoAppProcessor) Consume(ctx context.Context) error {
+	l := logger.FromContext(ctx).With(zap.String("event-consuming", "ready"))
+
+	for {
+		select {
+		case overview := <-a.trigger:
+			for _, envGroup := range overview.EnvironmentGroups {
+				for _, env := range envGroup.Environments {
+					appList, err := a.ApplicationClient.List(ctx, &application.ApplicationQuery{
+						Name: ptr.FromString(fmt.Sprintf("%s-", env)),
```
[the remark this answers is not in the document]
leandro-freiheit: In ArgoCD, the applications are prefixed with the environment name. The applications are always prefixed with the environment name.

services/rollout-service/pkg/argo/argo.go
leandro-freiheit: Argh, you are right with the duplicate operations. Changed this.

For the graceful handle, not sure how this would be handled.

leandro-freiheit: Addressed the handle of the application

services/rollout-service/pkg/argo/argo.go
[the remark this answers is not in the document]
leandro-freiheit: Actually, you are right on this, they are not necessarily prefixed with the env name. Changed the query.

PR #1272, opened 2024-01-10
feat: Add `--max-remote-cache` option to earthly buids

khaled-ismaeel:
This is not the intended final behvaior; we would like to upload the maximum cache only upon pushedto main. This is merely a test commit.

Edit: added new commits, so now the behavior is correct.

### DIFF
--- a/services/cd-service/Makefile
+++ b/services/cd-service/Makefile
@@ -62,10 +62,10 @@ bin/:
 
 build: bin/main
 
-build-pr: test release
+build-pr: test release-pr
 	echo "build on pull request"
 
-build-main: test release
+build-main: test release-main
 	echo "build on main"
 
 .PHONY: cleanup-pr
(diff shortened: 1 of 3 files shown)


khaled-ismaeel:
Me and @ahmed-nour-fdc verified this actually caches all intermediate layers. This resulted in an overall speedup in the build times for rollout and frontend services, while it caused a slight slowdown in the build time for the CD service.

The slowdown for the CD service is due to a behavior by Earthly that we didn't fully understand: it caches `apk add` commands in all cases except one.

### Further work:

- The current setup means we only receive upstream dependency updates when a service is updated. This is a security vulnerability and worth investigating new ways for caching dependencies.
- The behavior we didn't understand with Earthly might be a bug and might be worth a check or an issue on Earthly repository.
- Throughout our work on this PR we noticed quite a bit of duplication in the Makefiles, might be worth a refactor later.

ahmed-nour-fdc (approved):

PR #1279, opened 2024-01-18
fix: release endpoint reports conflicts for an idempotent call

bjoern-michaelsen-freiheit:
- the release endpoint should be idempotent
- however, it did not compare author, message, display version, team and repo correctly (all were compared to the commit_id errorously)
- team also had a wrong path (releaseDir instead of app dir)
- also introduce constants for the field names to minimize typo errors and further inconsistencies

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -21,15 +21,16 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


sven-urbanski-freiheit-com (approved):

sven-urbanski-freiheit-com (changes requested):
double check if this is a breaking change

sven-urbanski-freiheit-com (approved):

PR #1287, opened 2024-01-22
fix: exit early when the git protocol is wrong

sven-urbanski-freiheit-com:


bjoern-michaelsen-freiheit (approved):

ahmed-nour-fdc:
Can we add a test for it?

PR #1310, opened 2024-02-01
feat: Add TLS support for

ahmed-nour-fdc:
* Add support for  environment variable `KUBERPULT_CD_SERVER_SECURE` to enable TLS communication to the cd-service.

### DIFF
--- a/services/frontend-service/pkg/cmd/server.go
+++ b/services/frontend-service/pkg/cmd/server.go
@@ -18,6 +18,8 @@ package cmd
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"io"
 	"net/http"
@@ -47,6 +49,8 @@ import (
 	"google.golang.org/api/idtoken"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
 	grpctrace "gopkg.in/DataDog/dd-trace-go.v1/contrib/google.golang.org/grpc"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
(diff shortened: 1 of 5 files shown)


services/frontend-service/pkg/cmd/server.go
```diff
@@ -114,8 +119,23 @@ func runServer(ctx context.Context) error {
 		grpc_zap.UnaryServerInterceptor(grpcServerLogger),
 	}
 
+	logger.FromContext(ctx).Info("config.tls_enabled: " + strconv.FormatBool(c.TlsEnabled) + "\n")
```
sven-urbanski-freiheit-com, line 122: the whole config is printed in line 90 already, do we really need this log too?

services/frontend-service/pkg/cmd/server.go
```diff
@@ -114,8 +119,23 @@ func runServer(ctx context.Context) error {
...
+
+	var cred credentials.TransportCredentials = insecure.NewCredentials()
+	if c.TlsEnabled {
+		systemRoots, err := x509.SystemCertPool()
+		if err != nil {
+			msg := "failed to read CA certificates"
+			logger.FromContext(ctx).Error(msg)
+			return fmt.Errorf(msg)
```
sven-urbanski-freiheit-com, line 130: this looks like it would log twice. Once here and once in the calling function that prints the error again.
-> Remove the log here.

services/frontend-service/pkg/config/config.go
```diff
@@ -20,6 +20,7 @@ import "time"
 
 type ServerConfig struct {
 	CdServer            string        `default:"kuberpult-cd-service:8443"`
+	TlsEnabled          bool          `default:"false" split_words:"true"`
```
hannesg, line 23: We  will need this on   the  rollout service as well. Also, I'm not too happy with the config name "TlsEnabled" could also  mean that the http server creates a  tls listener instead of a plaintext http listener. Wouldn't it be easier to use a proper url for the CdServer variable like this `http://kuberpult-cd-service:8443` -> plaintext, `https://kuberpult-cd-service:8443` ->  tls?

ahmed-nour-fdc: > We will need this on the rollout service as well.

You are right. I will add this to the rollout-service as well.

> Also, I'm not too happy with the config name "TlsEnabled" could also mean that the http server creates a tls listener instead of a plaintext http listener.

Yes, this could be confusing. How about `GrpcTlsEnabled`?

> Wouldn't it be easier to use a proper url for the CdServer variable like this `http://kuberpult-cd-service:8443` -> plaintext, `https://kuberpult-cd-service:8443` -> tls?

AFAIK, grpc expects the `domain:port` without the protocol. If you supply the protocol as well, you get `too many colons in address` error.

hannesg: Yes,  grpc expects the host without protocol. However, we can easily extract this from the url ourself.

ahmed-nour-fdc: As agreed, I updated the name of the environment variable.

PR #1312, opened 2024-02-02
feat: Display ArgoCD rollout status in release dialog

hego-fdc:
Shows an icon & description of the ArgoCD rollout status of the current app in each environment.

Also changes the ArgoCD link to directly go to the details of the app instead of the search page.

image


services/frontend-service/src/assets/_components.scss
```diff
@@ -41,5 +41,6 @@ Copyright 2023 freiheit.com*/
 @import 'src/ui/components/Spinner/Spinner';
 @import 'src/ui/components/EnvironmentLockDisplay/EnvironmentLockDisplay';
 @import 'src/ui/components/ReleaseVersion/ReleaseVersion';
+@import 'src/ui/components/RolloutStatusDescription/RolloutStatusDescription.scss';
```
sven-urbanski-freiheit-com, line 44: is there a reason why we need to add ".scss" here?
If not, please remove the ending, so that the imports look uniform.
If yes, please add a comment why this is necessary.

hego-fdc: No, there's no reason, just an oversight. Removed it.

services/frontend-service/src/ui/components/ReleaseDialog/ReleaseDialog.test.tsx
```diff
@@ -291,6 +310,17 @@ describe('Release Dialog', () => {
                 },
             ],
         });
+        const status = testcase.rolloutStatus;
+        if (status !== undefined) {
```
sven-urbanski-freiheit-com, line 314: nitpick:
ideally, this should be a different test array, to avoid `if` statements in test code.
But since this would be against how we do it in this file with the other tests, I would be fine to keep it as it is now.

sven-urbanski-freiheit-com (approved):

services/frontend-service/src/assets/_components.scss
```diff
@@ -41,5 +41,6 @@ Copyright 2023 freiheit.com*/
 @import 'src/ui/components/Spinner/Spinner';
 @import 'src/ui/components/EnvironmentLockDisplay/EnvironmentLockDisplay';
 @import 'src/ui/components/ReleaseVersion/ReleaseVersion';
```
sven-urbanski-freiheit-com, line 43: please include a picture in the PR description of the new look (and if it makes sense also one screenshot for how it looked before)

PR #1315, opened 2024-02-02
feat: store event "newRelease" in manifest repo

sven-urbanski-freiheit-com:
This stores a kuberpult event of type "newRelease' in the manifest repo.
This will be used later to retrieve data about one commit.


services/cd-service/pkg/repository/transformer.go
```diff
@@ -374,9 +373,79 @@ func (c *CreateApplicationVersion) Transform(ctx context.Context, state *State)
...
+	if err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
+			// anything except "not exists" is an error:
+			return fmt.Errorf("event directory %s already existed: %v", eventDir, err)
+		}
+	}
+	if err := filesystem.MkdirAll(eventDir, 0777); err != nil {
+		return fmt.Errorf("could not create directory %s: %v", eventDir, err)
+	}
+	for i := range envs {
```
hannesg, line 423: Do we actually need to store a list of all environments here again? Wouldn't it be much more interesting to store the release version?

services/cd-service/pkg/repository/transformer.go
```diff
@@ -374,9 +373,79 @@ func (c *CreateApplicationVersion) Transform(ctx context.Context, state *State)
...
+	}
+	const (
+		EventTypeNewRelease = "new-release"
+	)
+	eventTypePath := filesystem.Join(eventDir, "eventType")
+	if err := util.WriteFile(filesystem, eventTypePath, []byte(EventTypeNewRelease), 0666); err != nil {
+		return fmt.Errorf("could not write file %s: %v", eventTypePath, err)
+	}
+
+	createdAtPath := filesystem.Join(eventDir, "createdAt")
```
hannesg, line 442: I we would use uuids as event ids, we could just get this from the event id.

services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: I would like to make the rpc call return all required information to render if possible.
In this case, this list of envs correlates with "which envs have manifests", which I think is a relevant question.
We could add the release version as well, but I don't require it for this story.

services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: This would require us to use a certain kind of uuid. I think it's simpler to just write the file.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -214,6 +219,7 @@ type CreateApplicationVersion struct {
 	SourceRepoUrl  string
 	Team           string
 	DisplayVersion string
+	EventId        string // the eventId is only there to make testing simpler
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: I did create an interface and put it in the context.
I'm a bit hesitant to implement our own UUID generator, I'm not convinced that it is really necessary.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -374,9 +373,79 @@ func (c *CreateApplicationVersion) Transform(ctx context.Context, state *State)
...
+}
+
+func writeEvent(ctx context.Context, eventId string, sourceCommitId string, filesystem billy.Filesystem, envs []string) error {
+	if eventId == "" {
+		fullUuid, err := uuid.NewUUID()
+		if err != nil {
+			return fmt.Errorf("could not create uuid: %v", err)
+		}
+		// format of this "uuid" is the same as the lock IDs, see "randomLockId" in store.tsx
+		eventId = "cd-v1-" + string(fullUuid.String()[:7])
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: done.

services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: It's easier for the client like this. We have one call that returns all required data.
We can later add the release version, once we need it.

services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: This depends on the UUID implementation, so I'd prefer to have the data explicitely.

hannesg: This is not accurate. UUIDv1 is an independant spec that is not depending on any implementation

sven-urbanski-freiheit-com: as discussed, we are now using time based UUIDs in this PR

hannesg (approved):

PR #1319, opened 2024-02-05
feat: Add filtering based on team to Product Version Page

gsandok:
- Adding a column to table that says team value
- Add filtering for the table based on selected teams

Image showing the new Team column
Screenshot 2024-02-05 at 16 49 12

Same image as before but filtering for a team
Screenshot 2024-02-05 at 16 49 22


services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
```diff
@@ -93,11 +144,18 @@ export const ProductVersion: React.FC = () => {
...
+                setSearchParams(searchParams);
+                getSummary(tagsResponse.response.tagData[0].commitId, env[0], env[1]);
+            } else {
+                getSummary(selectedTag, env[0], env[1]);
+            }
             setDisplayVersion(true);
-            setSelectedTag(tagsResponse.response.tagData[0].commitId);
         }
-    }, [tagsResponse, envGroupResponse, searchParams, environment]);
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
leandro-freiheit, line 157: Not seeing the need to disable this rule here. Technically we have less dependencies than before, no?

leandro-freiheit:
Since this is a very visual feature, I would really like a screenshot of what the feature looks like 😄 I think it really helps one to make suggestions

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
gsandok: There's some dependencies that disrupt with testing that I was unable to work around and be able to test properly. Without disabling the rule, I can't test the page and it is forever stuck in a loading state.

sven-urbanski-freiheit-com: @gsandok assuming we do not find a way to set the deps properly: Your comment here (in GitHub) should be part of the source code comment, otherwise it will get lost.

I've never seen a case, where the deps could not be set correctly, let's talk directly @gsandok .

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.test.tsx
```diff
@@ -146,3 +146,89 @@ describe('Product Version Data', () => {
...
+                    team: 'others',
+                },
+            ],
+            teams: ['sre-team', 'others'],
+        },
+    ];
+    describe.each(data)(`Displays Product Version Table`, (testCase) => {
+        render(<TableFiltered productSummary={testCase.productSummary} teams={testCase.teams} />);
+        expect(document.querySelector('.table')?.textContent).toContain('App Name');
+        expect(document.body).toMatchSnapshot();
```
sven-urbanski-freiheit-com, line 232: please remove "toMatchSnapshot" and replace it with individual "expect" invocations.

sven-urbanski-freiheit-com (changes requested):
.

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
```diff
@@ -91,27 +135,22 @@ export const ProductVersion: React.FC = () => {
...
         }
-    }, [summaryResponse]);
-    if (showTagsSpinner) {
+    }, [tagsResponse, envGroupResponse, environment, searchParams, setSearchParams]);
+    if (!tagsResponse.tagsReady) {
         return <Spinner message="Loading Tag Data" />;
     }
-    if (showSummarySpinner) {
+    if (!summaryResponse.summaryReady) {
         return <Spinner message="Loading Summary Data" />;
```
sven-urbanski-freiheit-com, line 154:
```suggestion
        return <Spinner message="Loading Production Version" />;
```

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
```diff
@@ -21,6 +21,62 @@ import { Spinner } from '../Spinner/Spinner';
 import { EnvironmentGroup, ProductSummary } from '../../../api/api';
 import { useSearchParams } from 'react-router-dom';
 
+export type TableProps = {
+    productSummary: ProductSummary[];
+    teams: string[];
+};
+
+export const TableFiltered: React.FC<TableProps> = (props) => {
+    var versionToDisplay = (app: ProductSummary): string => {
```
sven-urbanski-freiheit-com, line 30: Please always try to use `const`. If that doesn't work, use `let`.
```suggestion
    const versionToDisplay = (app: ProductSummary): string => {
```

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -104,6 +104,7 @@ export const [useTag, updateTag] = createStore<TagsResponse>({ response: tagsRes
 
 const summaryResponse: GetProductSummaryResponse = { productSummary: [] };
 export const getSummary = (commitHash: string, environment: string, environmentGroup: string): void => {
+    updateSummary.set({ summaryReady: false });
```
sven-urbanski-freiheit-com, line 107: please add some instructions to the readme about setting up the manifest repo locally.
Something like "git tag foo-my-tag1 && git push --tags", so that we have it locally as well.

sven-urbanski-freiheit-com (approved):
approved with some comments

PR #1321, opened 2024-02-06
fix: environment cards colors

khaled-ismaeel:
In this PR we change the rules for environment cards colors in the environments page.

Environments have a [`priority`](https://github.com/freiheit-com/kuberpult/blob/19d0f430f3948526694b099e4f9be295afc89754/pkg/api/v1/api.proto#L337) property that indicates where they lie in a chain of upstream relations. Currently the color of an environment card is [determined](https://github.com/freiheit-com/kuberpult/blob/4a3eaa34cd55732df7b16da61503fe54cbb0de9f/services/frontend-service/src/ui/components/EnvironmentCard/EnvironmentCard.tsx#L40) based on this property.

The color of environment groups is determined by the priority of the first environment in that group, which is erroneous: environments in one group have the same _distance to upstream_, but not necessarily the same _priority_. This leads to situations of an environment group its environments having different colors which looks awkward in the UI.

|  |
| :---: |
| _Figure 1: awkward UI colors, notice that both `testing-1` and `testing-2` have different colors. This is because `testing-2` has not downstream, so it is assigned the `PRODUCTION` priority level._|

In this PR we:
1. Introduce a new field to the environment group struct, called `priority` of type `Priority`. Despite the similarity, this field for environment groups is calculated in a different way from that of environments such that the priority is determined based on the distance to upstream and not the priorities of environments. The reason for reusing the same name and the same type is to keep changes to the frontend to a minimum.
2. Change the frontend logic for determining the colors of environment and environment group cards to be based on the newly introduced `EnvironmentGroup.priority` field.

Now it looks like this:

|  |
| :---: |
| _Figure 2: fixed colors in the UI_|


services/cd-service/pkg/mapper/environments_config_test.go
```diff
@@ -525,6 +569,105 @@ func TestMapEnvironmentsToGroup(t *testing.T) {
...
+						Environment: nameTestGlobal,
+					},
+					EnvironmentGroup: &nameStaging,
+				},
+				nameStagingUS: {
+					Upstream: &config.EnvironmentConfigUpstream{
+						Environment: nameTestGlobal,
+					},
+					EnvironmentGroup: &nameStaging,
+				},
```
sven-urbanski-freiheit-com, line 608: it seems redundant to have 3 envs in the same group

services/frontend-service/api
```diff
@@ -1 +1 @@
-../../pkg/api
\ No newline at end of file
+../../pkg/api/v1
```
sven-urbanski-freiheit-com, line 1: is this correct?

khaled-ismaeel: No, nice catch. That was specific to my local setup.

khaled-ismaeel: Yes it is. It does not affect CI builds anyhow, it is however needed for local development. It was needed after 

The reason this change is here is because this branch was created before the fix on the main branch #1323.

pkg/api/v1/api.proto
```diff
@@ -323,6 +323,15 @@ message EnvironmentGroup {
   repeated Environment environments = 2;
   // note that the distance_to_upstream should usually be configured to be the same for all envs in this group, but this is not enforced.
   uint32 distance_to_upstream = 3;
+  /*
+    Even though this field has the same name and type as Environment.priority, it does have difference semantics.
+
+    Environment priority is calculated based on the location of an environment in a chain. Environment group priority is calculated based on the distance to upstream **and** the maximum global distance to upstream.
+    This field therefore characterizes the "layer" of an environment group. The reason it is reusing the name and type of Environment.priority is to keep the calculation of environment colors in the frontned untouched.
```
sven-urbanski-freiheit-com, line 330:
```suggestion
    This field therefore characterizes the "layer" of an environment group. The reason it is reusing the name and type of Environment.priority is to keep the calculation of environment colors in the frontend untouched.
```

services/cd-service/pkg/mapper/environments_config_test.go
```diff
@@ -525,6 +563,98 @@ func TestMapEnvironmentsToGroup(t *testing.T) {
 						makeEnv(nameProdFr, nameProd, makeUpstreamEnvironment(nameStagingFr), 2, api.Priority_PROD),
 					},
 					DistanceToUpstream: 2,
+					Priority:           api.Priority_PROD,
+				},
+			},
+		},
+		{
+			Name: "Environments with different environment priorities",
+			InputEnvs: map[string]config.EnvironmentConfig{
```
leandro-freiheit, line 572: As the test scenario is quite complex, please add a comment specifying the conditions of the test. It increases readability and makes the test easier to understand

services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -720,6 +721,7 @@ export const useCurrentlyExistsAtGroup = (application: string): EnvironmentGroup
                     environments: envs,
                     distanceToUpstream: group.distanceToUpstream,
                     numberOfEnvsInGroup: group.environments.length,
+                    priority: group.priority,
```
sven-urbanski-freiheit-com, line 724: please provide screenshots and a little more details in the PR description

sven-urbanski-freiheit-com (approved):
approved with 2 small comments

pkg/api/v1/api.proto
leandro-freiheit, line 330:
```suggestion
    This field therefore characterizes the "layer" of an environment group. The reason it is reusing the name and type of Environment.priority is to keep the calculation of environment colors in the frontend untouched.
```

services/cd-service/pkg/mapper/environments_config.go
```diff
@@ -140,9 +140,47 @@ func MapEnvironmentsToGroups(envs map[string]config.EnvironmentConfig) []*api.En
...
 
+func calculateGroupPriority(distanceToUpstream, downstreamDepth uint32) api.Priority {
+	lookup := [][]api.Priority {
+		[]api.Priority {api.Priority_YOLO},
+		[]api.Priority {api.Priority_UPSTREAM, api.Priority_PROD},
+		[]api.Priority {api.Priority_UPSTREAM, api.Priority_PRE_PROD, api.Priority_PROD},
+		[]api.Priority {api.Priority_UPSTREAM, api.Priority_PRE_PROD, api.Priority_CANARY, api.Priority_PROD},
+		[]api.Priority {api.Priority_UPSTREAM, api.Priority_OTHER, api.Priority_PRE_PROD, api.Priority_CANARY, api.Priority_PROD},
+	}
+	if downstreamDepth > 4 {
```
leandro-freiheit, line 166:
```suggestion
	if downstreamDepth > len(lookup) - 1 {
```
Please do not use static values, think about the extensibility of the code 😄 The lookup already provides de maximum chain of possible unique priorities

khaled-ismaeel: That's a good point.

services/frontend-service/src/ui/utils/store.tsx
khaled-ismaeel: Done, is it detailed enough?

services/cd-service/pkg/mapper/environments_config_test.go
khaled-ismaeel: Fair enough. Added a diagram :D Is it good?

PR #1324, opened 2024-02-07
feat: reading and returning events of type newRelease from GetCommitInfo

sven-urbanski-freiheit-com:
Adds a new event type "newRelease" and returns the events in the grpc call "GetCommitInfo"

The event is not yet rendered in the UI.

### DIFF
--- a/pkg/uuid/uuid.go
+++ b/pkg/uuid/uuid.go
@@ -18,7 +18,6 @@ package uuid
 
 import (
 	"github.com/onokonem/sillyQueueServer/timeuuid"
-
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
--- a/services/cd-service/pkg/service/git.go
+++ b/services/cd-service/pkg/service/git.go
@@ -24,10 +24,15 @@ import (
 	"strings"
 
 	api "github.com/freiheit-com/kuberpult/pkg/api/v1"
-	"github.com/freiheit-com/kuberpult/pkg/valid"
 	grpcErrors "github.com/freiheit-com/kuberpult/pkg/grpc"
+	"github.com/freiheit-com/kuberpult/pkg/logger"
+	"github.com/freiheit-com/kuberpult/pkg/uuid"
+	"github.com/freiheit-com/kuberpult/pkg/valid"
+	eventmod "github.com/freiheit-com/kuberpult/services/cd-service/pkg/event"
 	"github.com/freiheit-com/kuberpult/services/cd-service/pkg/repository"
+	billy "github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-billy/v5/util"
+	"github.com/onokonem/sillyQueueServer/timeuuid"
 )
 
 type GitServer struct {
@@ -120,7 +125,7 @@ func (s *GitServer) GetProductSummary(ctx context.Context, in *api.GetProductSum
 
 func (s *GitServer) GetCommitInfo(ctx context.Context, in *api.GetCommitInfoRequest) (*api.GetCommitInfoResponse, error) {
 	fs := s.OverviewService.Repository.State().Filesystem
-	
+
 	commitID := in.CommitHash
 
 	if !valid.SHA1CommitID(commitID) {
@@ -132,18 +137,17 @@ func (s *GitServer) GetCommitInfo(ctx context.Context, in *api.GetCommitInfoRequ
 	commitPath := fs.Join("commits", commitID[:2], commitID[2:])
 
 	if _, err := fs.Stat(commitPath); err != nil {
-		return nil, grpcErrors.NotFoundError(ctx, fmt.Errorf("commit %s was not found in the manifest repo", commitID));
+		return nil, grpcErrors.NotFoundError(ctx, fmt.Errorf("commit %s was not found in the manifest repo", commitID))
 	}
 
 	sourceMessagePath := fs.Join(commitPath, "source_message")
 	var commitMessage string
 	if dat, err := util.ReadFile(fs, sourceMessagePath); err != nil {
-		return nil, fmt.Errorf("could not open the source message file at %s, err: %w", sourceMessagePath, err);
+		return nil, fmt.Errorf("could not open the source message file at %s, err: %w", sourceMessagePath, err)
 	} else {
 		commitMessage = string(dat)
 	}
 
-	
 	commitApplicationsDirPath := fs.Join(commitPath, "applications")
 	dirs, err := fs.ReadDir(commitApplicationsDirPath)
 	if err != nil {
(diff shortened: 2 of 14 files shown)


services/cd-service/pkg/service/git.go
```diff
@@ -153,10 +157,95 @@ func (s *GitServer) GetCommitInfo(ctx context.Context, in *api.GetCommitInfoRequ
...
+		oneEventDir := potentialEventDirs[i]
+		if oneEventDir.IsDir() {
+			fileName := oneEventDir.Name()
+			rawUUID, err := timeuuid.ParseUUID(fileName)
+			if err != nil {
+				return nil, errorTemplate(fmt.Sprintf("could not read event, because it is not a UUID '%s'", fileName), err)
+			}
+
+			var event *api.Event = nil
+			event, err = s.ReadEvent(ctx, fs, fs.Join(allEventsPath, oneEventDir.Name()), rawUUID)
```
leandro-freiheit, line 194: Hm, to me it would make more sense that the events are stored in `events/<event-UUID>` instead of `events/<event-name>/<event-UUID>`. What is the benefit of specifying the event name?

services/cd-service/pkg/service/git.go
```diff
@@ -153,10 +157,95 @@ func (s *GitServer) GetCommitInfo(ctx context.Context, in *api.GetCommitInfoRequ
...
+	for i := range potentialEventDirs {
+		oneEventDir := potentialEventDirs[i]
+		if oneEventDir.IsDir() {
+			fileName := oneEventDir.Name()
+			rawUUID, err := timeuuid.ParseUUID(fileName)
+			if err != nil {
+				return nil, errorTemplate(fmt.Sprintf("could not read event, because it is not a UUID '%s'", fileName), err)
+			}
+
+			var event *api.Event = nil
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: that is exactly what's happening. 
The event Type is stored like this:
`commits/8b/406da2f1d4abc2f7724692e20514b7513314b5/events/ec91b88b-c50c-11ee-33d1-17bf681a28d2/eventType`

pkg/uuid/uuid.go
```diff
@@ -16,7 +16,13 @@ Copyright 2023 freiheit.com*/
 
 package uuid
 
-import "4d63.com/uuid/uuidv1"
+import (
+	"github.com/onokonem/sillyQueueServer/timeuuid"
```
hego-fdc, line 20: Is there a reason why you're not using the more standard github.com/google/uuid? Seems like it also supports UUIDv1.

pkg/uuid/uuid.go
```diff
@@ -26,5 +32,11 @@ type RealUUIDGenerator struct {
 }
 
 func (t RealUUIDGenerator) Generate() string {
-	return uuidv1.Now().String()
+	return timeuuid.TimeUUID().String()
+}
+
+func GetTime(id *UUID) *timestamppb.Timestamp {
```
hego-fdc, line 38: Since this relies on a property of the UUID package that is not immediately obvious, maybe it would be good to have a few unit tests that make sure that the UUID package actually behaves in the way we expect it to? Otherwise it would be quite easy for someone to accidentially break functionality, e.g. by switching to a different UUID version.

services/cd-service/pkg/service/git.go
```diff
@@ -153,10 +157,95 @@ func (s *GitServer) GetCommitInfo(ctx context.Context, in *api.GetCommitInfoRequ
...
 	}, nil
-}
\ No newline at end of file
+}
+
+func (s *GitServer) GetEvents(ctx context.Context, fs billy.Filesystem, hash string, commitPath string) ([]*api.Event, error) {
+	errorTemplate := func(message string, err error) error {
+		return fmt.Errorf("could not get events for hash '%s' in path '%s' - error %w", hash, commitPath, err)
+	}
+	var result []*api.Event = nil
```
hego-fdc, line 178: `nil` is already the zero value

services/cd-service/pkg/service/git.go
hego-fdc, line 194: ` oneEventDir.Name()` is already assigned to `fileName`, so use that instead.

services/cd-service/pkg/service/git.go
```diff
@@ -24,10 +24,15 @@ import (
...
 
 	api "github.com/freiheit-com/kuberpult/pkg/api/v1"
-	"github.com/freiheit-com/kuberpult/pkg/valid"
 	grpcErrors "github.com/freiheit-com/kuberpult/pkg/grpc"
+	"github.com/freiheit-com/kuberpult/pkg/logger"
+	"github.com/freiheit-com/kuberpult/pkg/uuid"
+	"github.com/freiheit-com/kuberpult/pkg/valid"
+	eventmod "github.com/freiheit-com/kuberpult/services/cd-service/pkg/event"
 	"github.com/freiheit-com/kuberpult/services/cd-service/pkg/repository"
+	billy "github.com/go-git/go-billy/v5"
```
hego-fdc, line 33: Have you checked if `io/fs` also fulfills your needs for this? Would be more standard and save a dependency.

services/cd-service/pkg/service/git.go
```diff
@@ -153,10 +157,95 @@ func (s *GitServer) GetCommitInfo(ctx context.Context, in *api.GetCommitInfoRequ
...
+		TouchedApps:   touchedApps,
+		Events:        events,
 	}, nil
-}
\ No newline at end of file
+}
+
+func (s *GitServer) GetEvents(ctx context.Context, fs billy.Filesystem, hash string, commitPath string) ([]*api.Event, error) {
+	errorTemplate := func(message string, err error) error {
+		return fmt.Errorf("could not get events for hash '%s' in path '%s' - error %w", hash, commitPath, err)
```
hego-fdc, line 176: Maybe this is standard in the codebase, but it feels like this is an antipattern: You should not automatically wrap all errors that you return from a given function. The caller of the function should decide whether it should be wrapped to provide more context. Otherwise you'll end up with very long, redundant error messages in a lot of cases.

services/cd-service/pkg/service/git.go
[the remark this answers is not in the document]
leandro-freiheit: But we read the event from `fs.Join(allEventsPath, oneEventDir.Name()), rawUUID)`. What is `oneEventDir.Name()` then? I may not be getting it 😄

services/cd-service/pkg/service/git.go
```diff
@@ -153,10 +157,95 @@ func (s *GitServer) GetCommitInfo(ctx context.Context, in *api.GetCommitInfoRequ
...
+				CreateReleaseEvent: &api.CreateReleaseEvent{
+					EnvironmentNames: envs,
+				},
+			},
+		}
+		return result, nil
+	}
+	err = fmt.Errorf("could not read event, did not recognize event type '%s'", eventType)
+	logger.FromContext(ctx).Sugar().Warnf("%v", err)
+	return nil, err
```
hego-fdc, line 250: Do not log & return errors, this is an antipattern

services/cd-service/pkg/service/git.go
sven-urbanski-freiheit-com: > The caller of the function should decide whether it should be wrapped

I would argue that *I am* the caller here, wrapping errors from called functions.
If I have the choice between "sometimes too little information" and "sometimes too much information", I'll take the second one.
How would you do this?

services/cd-service/pkg/service/git.go
sven-urbanski-freiheit-com: good question, I suspect it would work too. However, we do use billy everywhere, so it would be odd to change it just here. We should discuss this in a refinement.

pkg/uuid/uuid.go
sven-urbanski-freiheit-com: I got the recommendation to not use it.
I just now tried it out briefly, and it does not seem as easy to use, e.g. it brings it's own "time" type.

pkg/uuid/uuid.go
sven-urbanski-freiheit-com: good point, I added a roundtrip test.

services/cd-service/pkg/service/git.go
sven-urbanski-freiheit-com: fair point, fixed

services/cd-service/pkg/service/git.go
hego-fdc: Sure, if you're already using it, then it's fine for now.

services/cd-service/pkg/service/git.go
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: these are 2 parameters.
`oneEventDir.Name()` is the name of the dir, which is indeed the same string as `rawUUID`.

first parameter: full path including the event id, so `commits/8b/406da2f1d4abc2f7724692e20514b7513314b5/events/ec91b88b-c50c-11ee-33d1-17bf681a28d2`

second parameter: just the UUID itself - this is not used as a directory name, but to extract the time. See call to `uuid.GetTime`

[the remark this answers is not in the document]
leandro-freiheit: Ah now I see. The `oneEventDir.Name()` is the identifier after the `commits/`. I was not aware 😄

leandro-freiheit (approved):

PR #1328, opened 2024-02-08
feat: Allow finding commits by prefix

hego-fdc:


pkg/valid/validations.go
```diff
@@ -22,18 +22,19 @@ import (
...
-	TeamNameRegExp = AppNameRegExp
-	EnvNameRegExp  = AppNameRegExp
-	CommitIDRegExp = `^[0-9a-fA-F]{40}$`
+	MaxAppNameLen        = 39
+	AppNameRegExp        = `\A[a-z0-9]+(?:-[a-z0-9]+)*\z`
+	TeamNameRegExp       = AppNameRegExp
+	EnvNameRegExp        = AppNameRegExp
+	SHA1CommitIDLength   = 40
+	commitIDPrefixRegExp = `^[0-9a-fA-F]*$`
 )
```
khaled-ismaeel, line 31: Perhaps this is a better way for writing this regex.

```suggestion
	commitIDPrefixRegExp = `^[0-9a-fA-F]{1,40}$`
```

pkg/valid/validations.go
```diff
@@ -22,18 +22,19 @@ import (
...
-	AppNameRegExp  = `\A[a-z0-9]+(?:-[a-z0-9]+)*\z`
-	TeamNameRegExp = AppNameRegExp
-	EnvNameRegExp  = AppNameRegExp
-	CommitIDRegExp = `^[0-9a-fA-F]{40}$`
+	MaxAppNameLen        = 39
+	AppNameRegExp        = `\A[a-z0-9]+(?:-[a-z0-9]+)*\z`
+	TeamNameRegExp       = AppNameRegExp
+	EnvNameRegExp        = AppNameRegExp
+	SHA1CommitIDLength   = 40
+	commitIDPrefixRegExp = `^[0-9a-fA-F]*$`
```
khaled-ismaeel, line 30: Perhaps this is a better way for writing this regex.

```suggestion
	commitIDPrefixRegExp = `^[0-9a-fA-F]{1,40}$`
```

services/cd-service/pkg/service/git_test.go
```diff
@@ -393,6 +395,32 @@ func TestGetCommitInfo(t *testing.T) {
...
+				Events: []*api.Event{
+					{
+						CreatedAt: timestamppb.New(fixedTime()),
+						EventType: &api.Event_CreateReleaseEvent{
+							CreateReleaseEvent: &api.CreateReleaseEvent{},
+						},
+					},
+				},
+			},
+		},
```
khaled-ismaeel, line 423: Perhaps we need to add test for the case when the commit ID length is less than 7 and greater than 40?

pkg/valid/validations.go
hego-fdc: I think it's better to check the length separately, as this gives you the ability to generate more precise error messages.

services/cd-service/pkg/service/git_test.go
hego-fdc: I don't think that those are especially important cases. Even if we would accidentally give out a commit if the length is <7 , there's no harm in it.

sven-urbanski-freiheit-com (approved):

PR #1330, opened 2024-02-09
feat: Display events of commits

hego-fdc:
image

### DIFF
--- a/services/frontend-service/src/ui/components/CommitInfo/CommitInfo.test.tsx
+++ b/services/frontend-service/src/ui/components/CommitInfo/CommitInfo.test.tsx
@@ -35,7 +35,17 @@ test('CommitInfo component renders commit info when the response is valid', () =
 Commit message body line 1
 Commit message body line 2`,
         touchedApps: ['google', 'windows'],
-        events: [],
+        events: [
+            {
+                createdAt: new Date('2024-02-09T09:46:00Z'),
+                eventType: {
+                    $case: 'createReleaseEvent',
+                    createReleaseEvent: {
+                        environmentNames: ['dev', 'staging'],
+                    },
+                },
+            },
+        ],
     };
     render(
         <MemoryRouter>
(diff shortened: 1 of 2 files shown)


services/frontend-service/src/ui/components/CommitInfo/CommitInfo.tsx
```diff
@@ -61,7 +61,48 @@ export const CommitInfo: React.FC<CommitInfoProps> = (props) => {
...
+                        <td>{description}</td>
+                        <td>{environments.join(', ')}</td>
+                    </tr>
+                );
+            })}
+        </tbody>
+    </table>
+);
+
+const eventDescription = (event: Event): [string, string[]] => {
```
sven-urbanski-freiheit-com, line 96: I think in the long run we need to return a JSXElement here (instead of the first string), because there should be more formatting that just "stirng". However, for this PR I'm fine with how it is.

services/frontend-service/src/ui/components/CommitInfo/CommitInfo.test.tsx
```diff
@@ -50,4 +60,6 @@ Commit message body line 2`,
     expect(screen.getAllByRole('row', { name: /Commit message body line 1/ })).toHaveLength(1);
     expect(screen.getAllByRole('row', { name: /Commit message body line 2/ })).toHaveLength(1);
     expect(screen.getAllByRole('row', { name: /google, windows/ })).toHaveLength(1);
+    expect(screen.getAllByRole('row', { name: /2024-02-09T09:46:00/ })).toHaveLength(1);
+    expect(screen.getAllByRole('row', { name: /received data about this commit for the first time/ })).toHaveLength(1);
```
sven-urbanski-freiheit-com, line 64: I think there should be a test that checks the list of environmentNames.

sven-urbanski-freiheit-com (approved):

services/frontend-service/src/ui/components/CommitInfo/CommitInfo.test.tsx
hego-fdc: Good point, done

PR #1343, opened 2024-02-15
fix: set range strategy=pin to nudge renovate into creating separate PRs

sven-urbanski-freiheit-com:


bjoern-michaelsen-freiheit (approved):
lgtm

PR #1348, opened 2024-02-16
WIP: events for deployments

sven-urbanski-freiheit-com:


hannesg:
I would have imaged a completely different solution to be honest. Instead of sending a 1 when there is a deployment, I would have assumed that you send the age of the last deployment so that your monitor can do a simple difference.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -135,7 +135,16 @@ func GaugeEnvAppLockMetric(fs billy.Filesystem, env, app string) {
 	}
 }
 
-func UpdateDatadogMetrics(state *State, changes *TransformerResult) error {
+func GaugeDeploymentMetric(_ context.Context, env, app string, timeInMinutes float64) error {
+	if ddMetrics != nil {
+		// store the time since the last deployment in minutes:
+		err := ddMetrics.Gauge("lastDeployed", timeInMinutes, []string{"app:" + app, "env:" + env}, 1)
```
hannesg, line 141: Can you please use the same tag names as in https://github.com/freiheit-com/kuberpult/blob/main/services/rollout-service/pkg/metrics/metrics.go#L152

hannesg (approved):

PR #1355, opened 2024-02-20
fix: do not expect /commit directories to always exist

sven-urbanski-freiheit-com:
It is ok if they do not exist, especially during clean-up.

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -709,6 +709,12 @@ func removeCommit(fs billy.Filesystem, commitID, application string) error {
 
 	commitApplicationDir := commitApplicationDirectory(fs, commitID, application)
 	if err := fs.Remove(commitApplicationDir); err != nil {
+		if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -708,6 +708,13 @@ func removeCommit(fs billy.Filesystem, commitID, application string) error {
...
 
 	commitApplicationDir := commitApplicationDirectory(fs, commitID, application)
+	_, err := fs.ReadDir(commitApplicationDir)
+	if err != nil {
+		// could not read the directory commitApplicationDir - but that's ok, because we don't know
+		// if the kuberpult version that accepted this commit in the release endpoint, did already have commit writing enabled.
+		// So there's no guarantee that this file ever existed
+		return nil
+	}
 	if err := fs.Remove(commitApplicationDir); err != nil {
```
hannesg, line 718: Wouldn't it be enough to check if the error here is `fs.ErrNotExist`?

sven-urbanski-freiheit-com: done

hannesg (approved):

PR #1358, opened 2024-02-21
fix: Adapt Release train endpoint to take commitHash as optional input

gsandok:
Release Train now run release train to specific commit hash instead of always to latest changes

REV: DSN-BADARR

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -21,8 +21,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/DataDog/datadog-go/v5/statsd"
-	"github.com/freiheit-com/kuberpult/pkg/metrics"
 	"io"
 	"io/fs"
 	"os"
@@ -32,8 +30,12 @@ import (
 	"strings"
 	"time"
 
+	"github.com/DataDog/datadog-go/v5/statsd"
+	"github.com/freiheit-com/kuberpult/pkg/metrics"
+
 	"github.com/freiheit-com/kuberpult/pkg/uuid"
 	"github.com/freiheit-com/kuberpult/services/cd-service/pkg/event"
+	git "github.com/libgit2/git2go/v34"
 
 	"github.com/freiheit-com/kuberpult/pkg/grpc"
 	"github.com/freiheit-com/kuberpult/pkg/valid"
@@ -1650,8 +1652,10 @@ func (c *DeployApplicationVersion) Transform(
 
 type ReleaseTrain struct {
 	Authentication
-	Target string
-	Team   string
+	Target     string
+	Team       string
+	CommitHash string
+	Repo       Repository
 }
 
 func getEnvironmentGroupsEnvironmentsOrEnvironment(configs map[string]config.EnvironmentConfig, targetGroupName string) map[string]config.EnvironmentConfig {
@@ -1773,18 +1777,9 @@ func (c *envReleaseTrain) Transform(
 			"Target Environment '%s' is locked - skipping.",
 			c.Env), nil
 	}
-	var apps []string
-	if upstreamLatest {
-		apps, err = state.GetApplications()
-		if err != nil {
-			return "", grpc.InternalError(ctx, fmt.Errorf("could not get all applications for %q: %w", source, err))
-		}
-	} else {
-		apps, err = state.GetEnvironmentApplications(upstreamEnvName)
-		if err != nil {
-			return "", grpc.PublicError(ctx, fmt.Errorf("upstream environment (%q) does not have applications: %w", upstreamEnvName, err))
-		}
-	}
+	apps, overrideVersions, err := c.getUpstreamLatestApp(upstreamLatest, state, ctx, upstreamEnvName, source, c.CommitHash)
+	if err != nil {
+		return "", nil, err
 	sort.Strings(apps)
 	// now iterate over all apps, deploying all that are not locked
 	numServices := 0
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -1657,6 +1657,105 @@ type ReleaseTrain struct {
 	CommitHash string
 	Repo       Repository
 }
+type Overview struct {
+	App     string
+	Version uint64
+}
+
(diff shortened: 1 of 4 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -1620,6 +1624,108 @@ func generateReleaseTrainResponse(envDeployedMsg, envSkippedMsg map[string]strin
...
+			if err != nil {
+				return nil, nil, grpc.PublicError(ctx, fmt.Errorf("unable to find findOldApplicationVersions for app %s: %w", app.App, err))
+			}
+			if len(versions) > 0 && versions[0] > app.Version {
+				return nil, nil, grpc.PublicError(ctx, fmt.Errorf("Version for app %s is older than 20 commits when running release train to commitHash %s: %w", app.App, c.CommitHash, err))
+			}
+
+		}
+		return apps, appVersions, nil
+	} else if upstreamLatest {
```
sven-urbanski-freiheit-com, line 1715: nitpick: no need for an `else` here:
```suggestion
	}
	if upstreamLatest {
```

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -1926,13 +1927,13 @@ func TestReleaseTrainErrors(t *testing.T) {
 			expectedError: "",
 			expectedCommitMsg: `Release Train to environment/environment group 'acceptance':
 
-Release Train to 'acceptance-ca' environment:
+Release Train to 'acceptance-ca' environment for 0 apps:
```
sven-urbanski-freiheit-com, line 1930: I think this change would get you into conflicts with @hego-fdc 
I would suggest to undo the "for n apps" addition to make merging simpler.

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -2093,6 +2094,255 @@ Environment "acceptance-de" has both upstream.latest and upstream.environment co
...
+				},
+			},
+			ReleaseTrainEnv:    "staging",
+			shouldSucceed:      true,
+			expectedError:      "",
+			expectedCommitMsg:  "Release Train to environment/environment group 'staging':\n\nRelease Train to 'staging' environment for 1 apps:\n\nThe release train deployed 1 services from 'dev' to 'staging'\ndeployed version 1 of \"test\" to \"staging\"\n\n\n\n\n",
+			overrideCommitHash: "",
+		},
+		{
+			Name: "Release train done with commit Hash",
```
sven-urbanski-freiheit-com, line 2139: I don't see a commit hash in this test, please rename it.

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -2093,6 +2094,255 @@ Environment "acceptance-de" has both upstream.latest and upstream.environment co
...
+				&DeployApplicationVersion{
+					Application: "test",
+					Environment: "dev",
+					Version:     1,
+				},
+			},
+			ReleaseTrainEnv:    "dev",
+			shouldSucceed:      false,
+			expectedError:      "rpc error: code = InvalidArgument desc = error: could not get app version for commitHash aa for dev: Error creating new oid for commitHash aa: invalid oid",
+			overrideCommitHash: "aa",
```
sven-urbanski-freiheit-com, line 2264: I'd like to see 1 additional test where `overrideCommitHash!=""` and the release train is successful.

sven-urbanski-freiheit-com (changes requested):
see other comments.

sven-urbanski-freiheit-com:
Please change "feat" to "fix" when merging (I did it for the title), because this is not a new feature for a kuberpult user/operator.

services/cd-service/pkg/repository/transformer_test.go
gsandok: The commit hash has to be generated once the setupTransformers are run here: https://github.com/freiheit-com/kuberpult/pull/1358/files#diff-894f09b2c5166abb03bb01b74cfbe3a0c17322a4fe3a938f2acf819d40bd3269R2310

services/cd-service/pkg/repository/transformer.go
```diff
@@ -1620,6 +1624,109 @@ func generateReleaseTrainResponse(envDeployedMsg, envSkippedMsg map[string]strin
...
+						} else {
+							app.Version = *version
+						}
+					}
+					resp = append(resp, Overview{App: app.Name, Version: app.Version})
+				}
+			}
+		}
+	}
+	return resp, nil
```
sven-urbanski-freiheit-com, line 1693: this is maybe a nitpick, but we should stay consistent with indentation style/early exits.
Instead of
```
if envs, err := foo(); err != nil {
  return ...
} else { ...
```

Applying this here will save 3 indentation levels.
Please do
```
envs, err := foo()
if err != nil {
  return ...
} // no else here
 ...
```

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -2093,6 +2094,257 @@ Environment "acceptance-de" has both upstream.latest and upstream.environment co
...
+				Target:     tc.ReleaseTrainEnv,
+				Repo:       repo,
+			}
+			if tc.overrideCommitHash != "TO_BE_REPLACED" {
+				releaseTrain = &ReleaseTrain{
+					CommitHash: tc.overrideCommitHash,
+					Target:     tc.ReleaseTrainEnv,
+					Repo:       repo,
+				}
+			}
```
sven-urbanski-freiheit-com, line 2323: This can be simplified:
```suggestion
			commitHash:=strings.TrimSpace(string(out2))
			if tc.overrideCommitHash != "TO_BE_REPLACED" {
        			commitHash:=tc.overrideCommitHash
			}
		        releaseTrain = &ReleaseTrain{
			        CommitHash: commitHash,
			        Target:     tc.ReleaseTrainEnv,
			        Repo:       repo,
	        	}
```

sven-urbanski-freiheit-com (approved):

services/cd-service/pkg/repository/transformer.go
gsandok: will do, thanks for noticing that!

sven-urbanski-freiheit-com (approved):

PR #1359, opened 2024-02-21
fix: Refactor transformer architecture

hego-fdc:
Remove the manual accumulation of commit messages & changes, which makes the transformers easier to understand & debug. It also makes future changes easier to implement.

It slightly changes the commit message for skipped services, making it shorter while still conveing the same information.

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -38,7 +38,6 @@ import (
 
 	"github.com/DataDog/datadog-go/v5/statsd"
 	"github.com/freiheit-com/kuberpult/pkg/logger"
-	"k8s.io/utils/strings/slices"
 
 	yaml3 "gopkg.in/yaml.v3"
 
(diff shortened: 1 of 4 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -201,13 +200,77 @@ func GetRepositoryStateAndUpdateMetrics(repo Repository) {
...
+		Commits:         runner.Commits,
+	}, nil
+}
 
-func (t TransformerFunc) Transform(ctx context.Context, state *State) (string, *TransformerResult, error) {
-	return (t)(ctx, state)
+type transformerRunner struct {
+	Context         context.Context
+	State           *State
+	Stack           [][]string
```
sven-urbanski-freiheit-com, line 231: Why do we need a `[][]string` for the stack instead of just `[]string`?
if we really need it, please add a small comment explaining the dimensions, like
 `// outer array is for foo, inner array is for bar`

sven-urbanski-freiheit-com (approved):

services/cd-service/pkg/repository/transformer.go
hego-fdc: Sure, I added an explanation. It's not the most obvious thing ever, but since it is basically invisible when you're writing a transformer, I think it's ok.

sven-urbanski-freiheit-com:
please wait for a review from @hannesg  too

hannesg:
I think you have some room to simplify it even more.

PR #1374, opened 2024-02-26
fix: only log internal errors as errors

sven-urbanski-freiheit-com:


hannesg:
Why even log them at all?

sven-urbanski-freiheit-com:
> Why even log them at all?

@hannesg fair point, removed it completely.

hannesg (approved):

leandro-freiheit (approved):

PR #1375, opened 2024-02-26
fix: handle invalid oid as invalidArgument

sven-urbanski-freiheit-com:


services/cd-service/pkg/service/overview.go
```diff
@@ -58,7 +58,7 @@ func (o *OverviewServiceServer) GetOverview(
 	if in.GitRevision != "" {
 		oid, err := git.NewOid(in.GitRevision)
 		if err != nil {
-			return nil, err
+			return nil, grpc.NotFoundError(ctx, fmt.Errorf("getOverview: could not find revision %v: %v", in.GitRevision, err))
```
hannesg, line 61: This is sadly wrong :|. We actually need to check the error here because it could also mean something else.

sven-urbanski-freiheit-com: Not sure I get it, what else could it mean and why is it important?
I think the difference between "not found" and "invalid parameter" is not crucial here, or are you thinking of something else?

sven-urbanski-freiheit-com: change it to be a different error @hannesg

hannesg (approved):

PR #1378, opened 2024-02-27
feat: use earthly to build (almost) everything

khaled-ismaeel:
Here we extend the use of Earthly to further build the images under `infrastructure/docker` and build the Helm chart under `charts/kuberpult`.


.github/workflows/execution-plan-snippet.yml
```diff
@@ -70,6 +70,10 @@ jobs:
         packages: write
         contents: read
     steps:
+      - name: Install Earthly
```
ahmed-nour-fdc, line 73: There is a github action that installs earthly. You can find it [here](https://github.com/freiheit-com/kuberpult/blob/main/.github/workflows/commitlint.yml#L12). You should use that instead.

charts/kuberpult/ci/lintconf.yaml
```diff
@@ -67,6 +67,6 @@ rules:
   new-line-at-end-of-file: enable
   new-lines:
     type: unix
-  trailing-spaces: enable
+  trailing-spaces: disable
```
ahmed-nour-fdc, line 70: Why are you disabling this?

charts/kuberpult/Earthfile
```diff
@@ -0,0 +1,104 @@
...
+
+test:
+    BUILD +ct-test
+    BUILD +test-helm
+
+clean:
+    LOCALLY
+
+    RUN rm -rf Chart.yaml
+    RUN rm -rf values.yaml
```
ahmed-nour-fdc, line 102: `values.yaml` is part of the repository and should not be deleted.

charts/kuberpult/Earthfile
```diff
@@ -0,0 +1,104 @@
+VERSION 0.7
+
+ARG --required --global VERSION
```
ahmed-nour-fdc, line 3: It is really annoying if I have to always pass the version when running earthly. You can set a default value for the argument as what is defined in [Makefile.variables](https://github.com/freiheit-com/kuberpult/blob/main/Makefile.variables#L17)
```suggestion
LOCALLY
ARG --global VERSION=$(git describe --always --long --tags || echo 0.0.1)
```

PR #1383, opened 2024-02-28
chore: check license in a separate workflow

khaled-ismaeel:
Here we license checking logic out of Makefiles (which will eventually be run via execution planner) into a separate GitHub Workflow.


.github/workflows/check-license.yml
```diff
@@ -0,0 +1,21 @@
+name: "Check license"
+'on':
+  pull_request:
+    branches:
+      - main
+  push:
+    tags-ignore:
+      - "**"
+    branches:
```
ahmed-nour-fdc, line 9: This feels redundant. The workflow should have already ran during the PR

ahmed-nour-fdc (approved):

PR #1384, opened 2024-02-29
fix: Event refactor

hego-fdc:
Create a reflection-based reader/writer for events which converts a Go-struct from/into a filesystem structure. Reflection is used here since it is the most straight-forward approach to handle generic en-/decoding functions.

### DIFF
--- a/services/cd-service/pkg/event/event_test.go
+++ b/services/cd-service/pkg/event/event_test.go
@@ -1,3 +1,18 @@
+/*This file is part of kuberpult.
+
+Kuberpult is free software: you can redistribute it and/or modify
+it under the terms of the Expat(MIT) License as published by
+the Free Software Foundation.
+
+Kuberpult is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+MIT License for more details.
+
+You should have received a copy of the MIT License
+along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>.
+
+Copyright 2023 freiheit.com*/
 package event
 
 import (
--- a/services/cd-service/pkg/event/event_test.go
+++ b/services/cd-service/pkg/event/event_test.go
@@ -1,0 +1,69 @@
+package event
+
+import (
+	"testing"
+
+	"github.com/go-git/go-billy/v5/memfs"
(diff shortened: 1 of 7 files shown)


services/cd-service/pkg/event/event_test.go
```diff
@@ -0,0 +1,84 @@
...
+				t.Fatal(err)
+			}
+			if diff := cmp.Diff(test.Event, result); diff != "" {
+				t.Error("wrong result:\n", diff)
+			}
+		})
+	}
+}
+
+func ptr[T any](x T) *T {
```
sven-urbanski-freiheit-com, line 82: nitpick: this could go to `ptr.go`

services/cd-service/pkg/event/writer.go
```diff
@@ -0,0 +1,139 @@
...
+	val any,
+) error {
+	return writeFile(filesystem, dir, reflect.ValueOf(val), encodingDefault)
+}
+
+func writeFile(
+	fs billy.Filesystem,
+	file string,
+	value reflect.Value,
+	encoding encoding,
```
sven-urbanski-freiheit-com, line 39: why is `encoding` needed? It seems to be passed only to `writeFile` recursively.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -547,43 +547,28 @@ func writeCommitData(ctx context.Context, sourceCommitId string, sourceMessage s
...
-		if err := util.WriteFile(filesystem, environmentNamePath, make([]byte, 0), 0666); err != nil {
-			return fmt.Errorf("could not write file %s: %v", environmentNamePath, err)
-		}
-	}
-	eventTypePath := filesystem.Join(eventDir, "eventType")
-	if err := util.WriteFile(filesystem, eventTypePath, []byte(event.NewReleaseEventName), 0666); err != nil {
-		return fmt.Errorf("could not write file %s: %v", eventTypePath, err)
-	}
-
 	// Note: we do not store the "createAt" date here, because we use UUIDs with timestamp information
```
sven-urbanski-freiheit-com, line 570: either move this comment somewhere else, or delete it, if it doesn't fit anymore.

sven-urbanski-freiheit-com (approved):

services/cd-service/pkg/event/event_test.go
hego-fdc: Since it's only used for testing, I'd rather not make it accessible for everything in that package.

PR #1390, opened 2024-03-01
feat: Add lock-prevented-deployment commit event

hego-fdc:
Whenever a lock (application or environment) blocks the deployment of an app, this event is created.

### DIFF
--- a/pkg/api/v1/api.proto
+++ b/pkg/api/v1/api.proto
@@ -62,6 +62,7 @@ message Event {
   oneof event_type {
     CreateReleaseEvent create_release_event = 3;
     DeploymentEvent deployment_event = 4;
+    LockPreventedDeploymentEvent lock_prevented_deployment_event = 5;
   }
 }
 
@@ -77,6 +78,18 @@ message DeploymentEvent {
   string application = 1;
   string target_environment = 2;
   optional ReleaseTrainSource release_train_source = 3;
+}
+
+message LockPreventedDeploymentEvent {
+  enum LockType {
+    LOCK_TYPE_UNKNOWN = 0;
+    LOCK_TYPE_ENV = 1;
+    LOCK_TYPE_APP = 2;
+  }
+  string application = 1;
+  string environment = 2;
+  string lock_message = 3;
+  LockType lock_type = 4;
 }
 
 message TagData {
--- a/services/cd-service/pkg/repository/transformer_test.go
+++ b/services/cd-service/pkg/repository/transformer_test.go
@@ -1575,6 +1575,89 @@ func TestApplicationDeploymentEvent(t *testing.T) {
 				},
 			},
 		},
+		{
+			Name: "Block deployments using locks",
+			Transformers: []Transformer{
+				&CreateEnvironment{
+					Environment: "dev",
+					Config: config.EnvironmentConfig{
+						Upstream: &config.EnvironmentConfigUpstream{
+							Latest: true,
+						},
+					},
+				},
+				&CreateEnvironment{
+					Environment: "staging",
+					Config: config.EnvironmentConfig{
+						Upstream: &config.EnvironmentConfigUpstream{
+							Latest: true,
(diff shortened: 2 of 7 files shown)


services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -1575,6 +1575,89 @@ func TestApplicationDeploymentEvent(t *testing.T) {
...
+				},
+				&CreateEnvironmentApplicationLock{
+					Environment: "dev",
+					Application: "myapp",
+					LockId:      "lock2",
+					Message:     "lock myapp",
+				},
+				&CreateApplicationVersion{
+					Application:    "myapp",
+					SourceCommitId: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaab",
```
khaled-ismaeel, line 1610: nitpick: you probably meant to have `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa` as commit ID.

pkg/api/v1/api.proto
```diff
@@ -79,6 +80,18 @@ message DeploymentEvent {
   optional ReleaseTrainSource release_train_source = 3;
 }
 
+message LockPreventedDeploymentEvent {
+  enum LockType {
+    LOCK_TYPE_UNKNOWN = 0;
```
khaled-ismaeel, line 85: Do we really need to have the `LOCK_TYPE_UNKNOWN`? I think we can cross it out and we should.

services/frontend-service/src/ui/components/CommitInfo/CommitInfo.tsx
```diff
@@ -134,5 +134,26 @@ const eventDescription = (event: Event): [JSX.Element, string] => {
                     );
             }
             return [description, de.targetEnvironment];
+        case 'lockPreventedDeploymentEvent':
+            const inner = tp.lockPreventedDeploymentEvent;
+            return [
+                <span>
+                    Application <b>{inner.application}</b> was blocked from deploying due to{' '}
```
khaled-ismaeel, line 141: It's probably useful to indicate the environment here as well.

khaled-ismaeel (commented):
See file comments.

services/frontend-service/src/ui/components/CommitInfo/CommitInfo.tsx
hego-fdc: Isn't that duplicating the information that is already in the "environment" column?

services/cd-service/pkg/repository/transformer_test.go
hego-fdc: Doesn't really matter, right?

pkg/api/v1/api.proto
hego-fdc: It's standard practice in protobuf to always map 0 to a special "unknown" value (https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum). Unless you know exactly how your system is evolving in the future, it's a good idea to stick to that.

services/frontend-service/src/ui/components/CommitInfo/CommitInfo.tsx
sven-urbanski-freiheit-com: I don't mind either way for now. We'll unify this later for all lock types.

services/cd-service/pkg/repository/transformer_test.go
khaled-ismaeel: Yup, it doesn't, hence the "nitpick" disclaimer.

pkg/api/v1/api.proto
khaled-ismaeel: Fair enough, though now I can think of many places where this practice needs to be applied.

khaled-ismaeel (approved):

PR #1391, opened 2024-03-01
fix: Lint rollout-service

ahmed-nour-fdc:


services/rollout-service/pkg/service/broadcast.go
```diff
@@ -176,7 +176,10 @@ func (b *Broadcast) StreamStatus(req *api.StreamStatusRequest, svc api.RolloutSe
 	resp, ch, unsubscribe := b.Start()
 	defer unsubscribe()
 	for _, r := range resp {
-		svc.Send(streamStatus(r))
+		err := svc.Send(streamStatus(r))
+		if err != nil {
+			return err
```
hannesg, line 181: I have the feeling this one needs a test. But we can merge this now and add the test later on.

hannesg (approved):

services/rollout-service/pkg/service/broadcast.go
ahmed-nour-fdc: From the looks of it, this entire for loop makes some changes to syncWindows variable that is not used again anywhere.
I will talk with Leandro about this before merging.

PR #1397, opened 2024-03-04
fix: activate self heal and sync prune for self managed apps

leandro-freiheit:


sven-urbanski-freiheit-com (approved):

hannesg (approved):
Please create a follow-up with a test for that.

PR #1402, opened 2024-03-04
fix: introduce TransformerBatchApplyError ...

bjoern-michaelsen-freiheit:
- to allow to forward, which transformer of a batch failed
- use it in ApplyTransformers and ApplyTransformersInternal
- remove grcp errors in low level repository: grpc errors should only be
  created on a high level at the endpoint
- try to remove (now?) rather pointless internal error type
- close the result channel and ensure only one error or nil is send
- rename too generic type element to transformerBatchrename, and applyElement to applyTransformerBatches

### DIFF
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -64,10 +64,25 @@ import (
 type Repository interface {
 	Apply(ctx context.Context, transformers ...Transformer) error
 	Push(ctx context.Context, pushAction func() error) error
-	ApplyTransformersInternal(ctx context.Context, transformers ...Transformer) ([]string, *State, []*TransformerResult, error)
+	ApplyTransformersInternal(ctx context.Context, transformers ...Transformer) ([]string, *State, []*TransformerResult, *TransformerBatchApplyError)
 	State() *State
 	StateAt(oid *git.Oid) (*State, error)
 	Notify() *notify.Notify
+}
+
+type TransformerBatchApplyError struct {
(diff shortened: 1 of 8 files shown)


services/cd-service/pkg/repository/repository.go
```diff
@@ -835,16 +850,20 @@ type ArgoWebhookData struct {
 	Commits       []commit
 }
 
-func (r *repository) ApplyTransformersInternal(ctx context.Context, transformers ...Transformer) ([]string, *State, []*TransformerResult, error) {
+func (r *repository) ApplyTransformersInternal(ctx context.Context, transformers ...Transformer) ([]string, *State, []*TransformerResult, *TransformerBatchApplyError) {
 	if state, err := r.StateAt(nil); err != nil {
-		return nil, nil, nil, grpc.InternalError(ctx, fmt.Errorf("%s: %w", "failure in StateAt", err))
+		return nil, nil, nil, &TransformerBatchApplyError{TransformerError: fmt.Errorf("%s: %w", "failure in StateAt", err), Index: -1}
```
sven-urbanski-freiheit-com, line 860: I think `nil` would be more clear, instead of `-1`

bjoern-michaelsen-freiheit: Index is a a value type, not at reference/pointer type though

sven-urbanski-freiheit-com (approved):

sven-urbanski-freiheit-com (approved):

PR #1404, opened 2024-03-04
chore: initialize all struct in `pkg`

khaled-ismaeel:
This is a prerequisite PR for #1396 

Here we address or silence all the errors generated by exhaustruct in the CD service. The decision for whether to address or silence an error was the mere convenience; if a struct initialization seemed too large or too complicated, the error was silenced, otherwise it was addressed.

### DIFF
--- a/tests/integration-tests/main_test.go
+++ b/tests/integration-tests/main_test.go
@@ -1,5 +1,4 @@
-/*
-This file is part of kuberpult.
+/*This file is part of kuberpult.
 
 Kuberpult is free software: you can redistribute it and/or modify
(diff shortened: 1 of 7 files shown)


tests/integration-tests/main_test.go
```diff
@@ -1,4 +1,5 @@
-/*This file is part of kuberpult.
+/*
+This file is part of kuberpult.
```
sven-urbanski-freiheit-com, line 2: why is this change necessary?

sven-urbanski-freiheit-com (approved):

PR #1421, opened 2024-03-08
feat: make batchClient timeout configurable

leandro-freiheit:
The batchClient can timeout with the static value we use. Therefore we should make the timeout configurable in the chart.

### DIFF
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -102,6 +102,7 @@ frontend:
 # This MUST be lower than the combined timeouts of ALL http proxies in use.
   maxWaitDuration: 10m
   batchClient:
+    # This value needs to be higher than the network timeout for git
     timeout: 2m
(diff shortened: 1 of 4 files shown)


charts/kuberpult/values.yaml
```diff
@@ -91,6 +91,8 @@ frontend:
 # See frontend-service.yaml for automatically added annotations.
   service:
     annotations: {}
+    batchClient:
```
hannesg, line 94: This place is slightly wrong because it contains the annotations of the kubernetes service. Rather move it one level up where also `maxWaitDuration` is.

hannesg: Also add a documentation that this must be higher than the cd servers gitNetworkTimeout

hannesg (approved):

hannesg (approved):

PR #1427, opened 2024-03-13
feat: limit number of batched commits per push

hannesg:
Theoretically we could end up in a situation where we are trying to fit an infinite amount of commits in one push. That is probably not an ideal thing to do. There should be a limit on this, especially in environments with high concurrency.

This change adds a new config option `.git.maximumCommitsPerPush` to the helm chart that is used as an upper limit for the number of commits per push.

### DIFF
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -53,6 +53,9 @@ git:
   enableWritingCommitData: false
 
   # Kuberpult tries to reduce the number of pushes and can bundle concurrent commits into a single push.
+  # This can reduce the time it takes to process requests ariving at the same time and improve throughput.
+  # The correct number largely depends on the performance of the git host and repository size. For small to medium sized deployments the default is good.
+  # We recommend values between 1 and 20.
   maximumCommitsPerPush: 5
 
 hub: europe-west3-docker.pkg.dev/fdc-public-docker-registry/kuberpult
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -52,6 +52,9 @@ git:
   # Disabling this option does not delete the `/commit` directory.
   enableWritingCommitData: false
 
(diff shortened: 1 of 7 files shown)


charts/kuberpult/values.yaml
```diff
@@ -52,6 +52,9 @@ git:
   # Disabling this option does not delete the `/commit` directory.
   enableWritingCommitData: false
 
+  # Kuberpult tries to reduce the number of pushes and can bundle concurrent commits into a single push.
```
sven-urbanski-freiheit-com, line 55: Please check at startup of the cd-service, that the value is in a sane range.
```suggestion
  # Kuberpult tries to reduce the number of pushes and can bundle concurrent commits into a single push.
  # Must be 1 <= maximumCommitsPerPush <= 100
```

services/cd-service/pkg/repository/repository.go
```diff
@@ -534,22 +534,26 @@ func (r *repository) useRemote(ctx context.Context, callback func(*git.Remote) e
 }
 
 func (r *repository) drainQueue() []transformerBatch {
+	if r.config.MaximumCommitsPerPush == 0 {
```
sven-urbanski-freiheit-com, line 537: nitpick: from the naming it's a bit counter intuitive that 0 is valid value.
Maybe `maximumAdditionalCommitsPerPush`?

sven-urbanski-freiheit-com (approved):

charts/kuberpult/values.yaml
hannesg: Not sure if we want to set a hard upper bound for this value all. Currently the value is essentially infite and we only see issues for the very largest deployments. Good values for the limit depend on the performance of the git provider and I wouldn't want to set a fixed limit there. I will still add a comment and check for lower bound.

services/cd-service/pkg/repository/repository.go
hannesg: Yes, agreed. I changed the code so that the default is set when we initialize the repository.

PR #1433, opened 2024-03-14
fix: reduce flakiness by refactoring unit test error/proto comparisons

obreitwi:
* aim to reduce flakiness of tests due to protojson instabilities

Rev: SRX-UG7OJV


README_DEVELOPERS.md
```diff
@@ -120,6 +120,59 @@ cd services/cd-service
 go test ./... -v
 ```
 
+### Best practices for unit tests
+
+#### Always use `cmp.Diff`
+
+When writing unit tests, aim to always compare with `cmp.Diff` and print the result.
+For errors, the the test should check via:
```
sven-urbanski-freiheit-com, line 128:
```suggestion
For errors, the test should check via:
```

pkg/auth/azure_test.go
```diff
@@ -28,40 +28,54 @@ import (
...
+	msg string
+}
+
+func (e errMatcher) Error() string {
+	return e.msg
+}
+
+func (e errMatcher) Is(err error) bool {
+	return e.Error() == err.Error()
+}
```
sven-urbanski-freiheit-com, line 45: this code seems quite duplicated. Can we move it into testutil.go?

sven-urbanski-freiheit-com (changes requested):
(see other comments)

pkg/auth/azure_test.go
obreitwi: Was not sure if I should introduce coupling between tests, but can move the code, yes.

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -5553,12 +5586,18 @@ func TestAllErrorsHandledDeleteEnvironmentLock(t *testing.T) {
 func TestAllErrorsHandledDeleteEnvironmentApplicationLock(t *testing.T) {
 	t.Parallel()
 	collector := &testfs.UsageCollector{}
+	fixtureWrapTransformError := func(err error) *TransformerBatchApplyError {
```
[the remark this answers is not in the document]
obreitwi: Was not sure if I should already add them there, keeping them local to the test makes them easier to change, but can move to `testutil.go`.

sven-urbanski-freiheit-com (approved):

services/cd-service/pkg/repository/transformer_test.go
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: let's keep it as it is, to get this merged quickly

PR #1434, opened 2024-03-14
feat: frontend implementation of previous and next buttons on commit info

miguel-crespo-fdc:
UI elements and tests for next and previous commit buttons on Commit Info Page. Functionality only fully visible once [this PR](https://github.com/freiheit-com/kuberpult/pull/1435) is accepted.

Screenshot 2024-03-15 at 08 41 53


services/frontend-service/src/ui/components/CommitInfo/CommitInfo.test.tsx
```diff
@@ -209,6 +299,29 @@ test('CommitInfo component renders commit info when the response is valid', () =
...
+                    if (testCase.commitInfo.nextCommitHash !== '') {
+                        expect(target).toHaveTextContent('Next Commit');
+                    }
+                } else {
+                    expect(targetElements.length).toEqual(0);
+                }
+            }
+        } else {
+            expect(targetElements.length).toEqual(0);
+        }
```
sven-urbanski-freiheit-com, line 323: The test should be more specific.
In general, we should avoid generic test functions like this `verifyTable`. Even though they have their place, they are not the ideal way to test most cases. Also `next/previous` are not part of the html table.

For this PR, I expect a specific test with it's own test array, and that test only looks at the `next` and `previous` button. There should also be very few `if` statements in tests.
(this is a bit hard to explain via text only)

services/frontend-service/src/ui/components/CommitInfo/CommitInfo.tsx
```diff
@@ -15,15 +15,30 @@ along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>
...
+        if (commitInfo !== undefined) {
+            window.location.href = commitInfo?.previousCommitHash;
+        }
+    }, [commitInfo]);
+
+    const onClickNext = useCallback((): void => {
+        if (commitInfo !== undefined) {
+            window.location.href = commitInfo?.nextCommitHash;
+        }
+    }, [commitInfo]);
```
sven-urbanski-freiheit-com, line 40: Please use normal links (`<a href...`) instead of callbacks.
Links are there to reach places via a url. Buttons/callbacks are more complicated and not needed in this case.

services/frontend-service/src/ui/components/CommitInfo/CommitInfo.test.tsx
```diff
@@ -213,3 +215,151 @@ test('CommitInfo component renders commit info when the response is valid', () =
...
+        head: string[];
+        // NOTE: newlines, if there are any, will effectively be removed, since they will be checked using .toHaveTextContent
+        body: string[][];
+    };
+
+    type TestCase = {
+        commitInfo: GetCommitInfoResponse;
+        expectedTitle: string;
+        expectedCommitDescriptionTable: Table;
+        expectedEventsTable: Table;
```
sven-urbanski-freiheit-com, line 230:
```suggestion
```
Either remove and move to another test

services/frontend-service/src/ui/components/CommitInfo/CommitInfo.tsx
```diff
@@ -15,15 +15,16 @@ along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>
 Copyright 2023 freiheit.com*/
 
 import { TopAppBar } from '../TopAppBar/TopAppBar';
-import React from 'react';
 import { GetCommitInfoResponse, Event, LockPreventedDeploymentEvent_LockType } from '../../../api/api';
+import './CommitInfo.scss';
```
sven-urbanski-freiheit-com, line 19: this should not be necessary, because we import all scss here: https://github.com/freiheit-com/kuberpult/blob/main/services/frontend-service/src/assets/_components.scss

sven-urbanski-freiheit-com (approved):

PR #1453, opened 2024-03-20
feat: implement endpoint to retrieve manifests

obreitwi:
* add new endpoint `/api/application/<app>/release/<release>/manifests`
* `/api` is protected via IAP by default
* set `KUBERPULT_API_ENABLE_DESPITE_NO_AUTH=true` to disable IAP protection (only recommended for local dev)

Rev: SRX-MJL1MD

### DIFF
--- a/charts/kuberpult/templates/frontend-service.yaml
+++ b/charts/kuberpult/templates/frontend-service.yaml
@@ -118,6 +118,10 @@ spec:
           value: "https://{{ .Values.ingress.domainName }}"
         - name: KUBERPULT_GIT_BRANCH
           value: {{ .Values.git.branch | quote }}
+        - name: KUBERPULT_IAP_ENABLED
+          value: {{ .Values.ingress.iap.enabled | quote }}
+        - name: KUBERPULT_API_ENABLE_DESPITE_NO_AUTH
+          value: {{ .Values.auth.api.enableDespiteNoAuth | quote }}
 {{- if .Values.datadogTracing.enabled }}
         - name: DD_AGENT_HOST
           valueFrom:
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -270,7 +270,7 @@ auth:
     # As of now this applies only to the manifest endpoint <insert rest pattern here>
     # If you do not have Googles IAP enabled, but still want to use the API, be aware that it is publicly available,
     # so you would need a protection outside of kuberpult (e.g. a VPN).
-    # If `enableDespiteNoAuth=true`, then the API will only respond, if IAP is enabled in this helm chart (`ingress.iap.enabled=true`).
+    # If `enableDespiteNoAuth=true`, then the API will respond, even if IAP is disabled in this helm chart (`ingress.iap.enabled=true`).
     enableDespiteNoAuth: false
 
 # The Dex configuration values. For more information please check the Dex repository https://github.com/dexidp/dex
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -267,7 +267,7 @@ auth:
     scopes: ""
   api:
     # New api endpoints (starting with `/api/`), are by default only turned on when IAP is enabled.
-    # As of now this applies only to the manifest endpoint <insert rest pattern here>
+    # As of now this applies only to the manifest endpoint `/api/application/<application>/release/<release>/manifests`.
     # If you do not have Googles IAP enabled, but still want to use the API, be aware that it is publicly available,
     # so you would need a protection outside of kuberpult (e.g. a VPN).
     # If `enableDespiteNoAuth=true`, then the API will respond, even if IAP is disabled in this helm chart (`ingress.iap.enabled=true`).
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -265,6 +265,13 @@ auth:
     baseURL: ""
     # List of scopes to validate the token. Please add them as comma separated values.
     scopes: ""
+  api:
+    # New api endpoints (starting with `/api/`), are by default only turned on when IAP is enabled.
+    # As of now this applies only to the manifest endpoint <insert rest pattern here>
+    # If you do not have Googles IAP enabled, but still want to use the API, be aware that it is publicly available,
+    # so you would need a protection outside of kuberpult (e.g. a VPN).
+    # If `enableDespiteNoAuth=true`, then the API will only respond, if IAP is enabled in this helm chart (`ingress.iap.enabled=true`).
+    enableDespiteNoAuth: false
 
 # The Dex configuration values. For more information please check the Dex repository https://github.com/dexidp/dex
 dex:
--- a/services/cd-service/pkg/service/version.go
+++ b/services/cd-service/pkg/service/version.go
@@ -20,6 +20,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"strconv"
+
 	"github.com/freiheit-com/kuberpult/pkg/grpc"
 
 	"google.golang.org/grpc/codes"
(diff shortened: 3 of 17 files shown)


charts/kuberpult/templates/frontend-service.yaml
```diff
@@ -118,6 +118,8 @@ spec:
           value: "https://{{ .Values.ingress.domainName }}"
```
sven-urbanski-freiheit-com, line 118: please change the commit messages, e.g.
`feat: implement endpoint to retrieve manifests` (and then more details in the next line)

charts/kuberpult/values.yaml
```diff
@@ -253,6 +253,9 @@ auth:
     baseURL: ""
     # List of scopes to validate the token. Please add them as comma separated values.
     scopes: ""
+  iapAuth:
+    # enfore IAP authorization for API
```
sven-urbanski-freiheit-com, line 257: this requires a bit more explanation as to why it's there. I'm not sure if "enforce" is the best word to describe it.

charts/kuberpult/templates/frontend-service.yaml
obreitwi: done

services/cd-service/pkg/service/version.go
```diff
@@ -72,3 +75,58 @@ func (o *VersionServiceServer) GetVersion(
...
+
+		if len(allReleases) == 0 {
+			return nil, status.Errorf(codes.NotFound, "no releases found for application %s", req.Application)
+		}
+
+		for _, r := range allReleases {
+			if r > release {
+				release = r
+			}
+		}
```
sven-urbanski-freiheit-com, line 109: this looks like it can be simplified by calling `GetLastRelease`

services/cd-service/pkg/service/version.go
```diff
@@ -72,3 +75,58 @@ func (o *VersionServiceServer) GetVersion(
...
+	if err != nil {
+		return nil, wrapError("manifests", err)
+	}
+
+	return &api.GetManifestsResponse{
+		Release:   repoRelease.ToProto(),
+		Manifests: manifests,
+	}, nil
+
+	return nil, status.Error(codes.Unimplemented, "not implemented")
```
sven-urbanski-freiheit-com, line 131: why are there 2 returns here?

services/cd-service/pkg/service/version_test.go
```diff
@@ -166,3 +173,167 @@ func TestVersion(t *testing.T) {
...
+	type testCase struct {
+		name    string
+		setup   []repository.Transformer
+		req     *api.GetManifestsRequest
+		want    *api.GetManifestsResponse
+		wantErr error
+	}
+
+	appName := "app-" + rand.String(16)
+	appNameOther := "app-other-" + rand.String(16)
```
sven-urbanski-freiheit-com, line 187: do we really need random values here? Why can't the apps have the same name?

services/cd-service/pkg/service/version_test.go
```diff
@@ -166,3 +173,167 @@ func TestVersion(t *testing.T) {
...
+			setup: append(fixtureSetupEnv(),
+				fixtureRelease(appNameOther, 1),
+				fixtureRelease(appNameOther, 2),
+				fixtureRelease(appNameOther, 3),
+			),
+			req:     fixtureRequest(),
+			wantErr: status.Errorf(codes.NotFound, "no releases found for application %s", appName),
+		},
+	} {
+		tc := tc // TODO: Remove after switching to go v1.22
```
sven-urbanski-freiheit-com, line 313: no "TODO" comments without reference.
Please create a story to upgrade to 1.22 and share the story Id in the comment

services/frontend-service/pkg/cmd/server.go
```diff
@@ -313,32 +313,50 @@ func runServer(ctx context.Context) error {
...
 		AzureAuth:     c.AzureEnableAuth,
 	}
 	mux := http.NewServeMux()
-	mux.Handle("/environments/", http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+	restHandler := http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		defer readAllAndClose(req.Body, 1024)
 		if c.DexEnabled {
-			interceptors.DexLoginInterceptor(w, req, httpHandler, c.DexClientId, c.DexClientSecret)
+			interceptors.DexLoginInterceptor(w, req, httpHandler.Handle, c.DexClientId, c.DexClientSecret)
+			return
```
sven-urbanski-freiheit-com, line 326:
```suggestion
```
This `return` would mean that we exit whenever dex is enabled.

services/frontend-service/pkg/handler/applications.go
```diff
@@ -111,3 +112,63 @@ func (s Server) handleDeleteApplicationLock(w http.ResponseWriter, req *http.Req
...
+		Release:     releaseNum,
+	})
+	if err != nil {
+		handleGRPCError(req.Context(), w, err)
+		return
+	}
+	encoded, err := json.Marshal(resp)
+	if err != nil {
+		logger.FromContext(req.Context()).Error("GetManifests: encoding response")
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
```
sven-urbanski-freiheit-com, line 164:
```suggestion
		http.Error(w, "GetManifests: encoding response", http.StatusInternalServerError)
```

services/frontend-service/pkg/cmd/server.go
```diff
@@ -520,6 +538,8 @@ func (p *Auth) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 }
 
 // GrpcProxy passes through gRPC messages to another server.
+// This is needed for the UI to communicate with other services via gRPC over web.
+// The UI _only_ communicates via gRPC over web (+ static files), while the REST API is only inteded for build pipelines.
```
sven-urbanski-freiheit-com, line 542:
```suggestion
// The UI _only_ communicates via gRPC over web (+ static files), while the REST API is only intended for automated processes like build pipelines.
```

charts/kuberpult/values.yaml
```diff
@@ -253,6 +253,9 @@ auth:
...
+    enforce_api: true
```
sven-urbanski-freiheit-com, line 258:
```suggestion
  api:
    # New api endpoints (starting with `/api/`), are by default only turned on when IAP is enabled.
    # As of now this applies only to the manifest endpoint <insert rest pattern here>
    # If you do not have Googles IAP enabled, but still want to use the API, be aware that it is publicly available.
    # If `onlyWithIAP=true`, then the API will only respond, if IAP is enabled in this helm chart (`ingress.iap.enabled=true`).
    onlyWithIAP: true
```

services/cd-service/pkg/service/version_test.go
obreitwi: I typically do that to force myself to not use literals - you _have_ to refer to the value provided here and so make the coupling to other components of the test more explicit. I could change it though.

services/frontend-service/pkg/cmd/server.go
obreitwi: Which we should do, right? Because the `DexLoginInterceptor` [calls the underlying http handler in any case](https://github.com/freiheit-com/kuberpult/blob/006f7b4ec1cc0422416f89cc9467159c1f3b6397/services/frontend-service/pkg/interceptors/interceptors.go#L116) so there is no point calling it again.

Alternatively, we could just return the modified context and then let the http handler do the validation of correct headers.

services/cd-service/pkg/service/version.go
obreitwi: Did not see that function :see_no_evil:

obreitwi: Is it okay to call it like `GetLastRelease(state.Filesystem, applicationName)`? It feels wrong to access `.Filesystem` directly while in other cases we simply call a function on `*State`. It seems inconsistent. How should we proceed?

services/frontend-service/pkg/cmd/server.go
obreitwi: Discussed in person: We keep the return.

charts/kuberpult/values.yaml
```diff
@@ -265,6 +265,13 @@ auth:
     baseURL: ""
     # List of scopes to validate the token. Please add them as comma separated values.
     scopes: ""
+  api:
+    # New api endpoints (starting with `/api/`), are by default only turned on when IAP is enabled.
+    # As of now this applies only to the manifest endpoint <insert rest pattern here>
```
sven-urbanski-freiheit-com, line 270: insert pattern here :)

charts/kuberpult/values.yaml
```diff
@@ -265,6 +265,13 @@ auth:
...
+    # If you do not have Googles IAP enabled, but still want to use the API, be aware that it is publicly available,
+    # so you would need a protection outside of kuberpult (e.g. a VPN).
+    # If `enableDespiteNoAuth=true`, then the API will only respond, if IAP is enabled in this helm chart (`ingress.iap.enabled=true`).
```
sven-urbanski-freiheit-com, line 273:
```suggestion
    # If `enableDespiteNoAuth=true`, then the API will respond, even if IAP is disabled in this helm chart (`ingress.iap.enabled=true`).
```

services/frontend-service/pkg/interceptors/interceptors.go
```diff
@@ -95,14 +96,40 @@ func StreamAuthInterceptor(
 	return handler(srv, stream)
 }
 
+// DexLoginInterceptor intercepts HTTP calls to the frontend service.
```
sven-urbanski-freiheit-com, line 99:
```suggestion
// GoogleIAPInterceptor intercepts HTTP calls to the frontend service.
```

services/frontend-service/pkg/interceptors/interceptors.go
```diff
@@ -95,14 +96,40 @@ func StreamAuthInterceptor(
...
+// DexLoginInterceptor must only be used if dex is enabled.
```
sven-urbanski-freiheit-com, line 100: why? IAP should work without dex.

sven-urbanski-freiheit-com (approved):

PR #1454, opened 2024-03-20
fix: fix creating a release (error messages and sh)

miguel-crespo-fdc:


services/frontend-service/pkg/handler/release.go
```diff
@@ -200,7 +200,11 @@ func (s Server) HandleRelease(w http.ResponseWriter, r *http.Request, tail strin
 			tf.PreviousCommitId = previousCommitId[0]
 		} else {
 			w.WriteHeader(400)
-			fmt.Fprintf(w, "Too many previous commits provided.")
+			if len(previousCommitId) > 1 {
```
sven-urbanski-freiheit-com, line 203: this is a little nit-picky maybe, but I'd like the if-else chain to be smaller. Something like
```
if len != 1 {
  w.WriteHeader(400)
  fmt.Printf(msg1)
  return
}
if !isCommit {
  w.WriteHeader(400)
  fmt.Printf(msg2)
  return
}
tf.previousCommit = previousCommitId[0]

```

sven-urbanski-freiheit-com (approved):

PR #1458, opened 2024-03-21
fix: Resolve pkg lint errors

ahmed-nour-fdc:


pkg/auth/azure.go
```diff
@@ -110,7 +111,7 @@ func HttpAuthMiddleWare(resp http.ResponseWriter, req *http.Request, jwks *keyfu
 	}
 	if err != nil {
 		resp.WriteHeader(http.StatusUnauthorized)
-		resp.Write([]byte("Invalid authorization header provided"))
+		resp.Write([]byte("Invalid authorization header provided")) //nolint: errcheck
```
hannesg, line 114: You can  make the lint pass by writing it like this:

```
_, _ = resp.Write(...)
```

pkg/setup/setup.go
```diff
@@ -114,7 +113,7 @@ func Run(ctx context.Context, config ServerConfig) {
 	pv, handler, _ := metrics.Init()
 	ctx = metrics.WithProvider(ctx, pv)
 
-	pv.Meter("setup").Int64ObservableGauge("background_job_ready", metric.WithInt64Callback(func(_ context.Context, o metric.Int64Observer) error {
```
hannesg, line 117: What is the lint actually complaining about?

pkg/setup/setup.go
```diff
@@ -167,10 +166,10 @@ func (s *setup) listenToShutdownSignal(ctx context.Context, cancelFunc context.C
...
 
 	// call shutdown hooks
-	gracefulShutdown(ctx, s, 30*time.Second)
+	gracefulShutdown(s, 30*time.Second)
 }
 
-func gracefulShutdown(ctx context.Context, s *setup, timeout time.Duration) {
+func gracefulShutdown(s *setup, timeout time.Duration) {
 	// Instantiate background context
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
hannesg, line 174: Uh, that is a good catch. I would think that the ctx paramer actually makes sense and we should not use `context.Background()` but `ctx` as first pararmeter here to pass the logger and tracer from the context.

ahmed-nour-fdc: What I understood from tracing the code is that we don't actually want to use ctx argument as it is probably already cancelled previously, but I might be wrong here.

pkg/setup/setup.go
ahmed-nour-fdc: Return value not checked. I couldn't figure out exactly how to do so.

pkg/setup/setup.go
sven-urbanski-freiheit-com: I haven't tested it, but I would assume that still use a context for logging, even when it's cancelled. In which case we should do what hannes suggested.

sven-urbanski-freiheit-com (approved):

PR #1460, opened 2024-03-21
fix: Resolve cd-service lint errors

ahmed-nour-fdc:


services/cd-service/pkg/mapper/environments_config.go
```diff
@@ -338,9 +338,6 @@ func TransformSyncWindows(syncWindows []config.ArgoCdSyncWindow, appName string)
 }
 
 func TransformArgocd(config config.EnvironmentConfigArgoCd) *api.EnvironmentConfig_ArgoCD {
-	if &config == nil {
```
ahmed-nour-fdc, line 341: Address of a variable cannot be nil

services/cd-service/pkg/fs/fs.go
```diff
@@ -136,7 +136,7 @@ func (b *treeBuilderBlob) Write(p []byte) (int, error) {
 }
 
 func (b *treeBuilderBlob) Read(p []byte) (int, error) {
-	b.load()
+	b.load() //nolint: errcheck
```
sven-urbanski-freiheit-com, line 139: why not handle the error here?

services/cd-service/pkg/repository/repository.go
```diff
@@ -323,7 +325,7 @@ func New(ctx context.Context, cfg RepositoryConfig) (Repository, error) {
 	if err != nil {
 		return nil, err
 	}
-	go bg(ctx, nil)
+	go bg(ctx, nil) //nolint: errcheck
```
sven-urbanski-freiheit-com, line 328: we should print the error. It might be easier to do that when `bg` does not return an error, but just prints the error itself.
Or created a wrapper that handles the logging of the error.

services/cd-service/pkg/cmd/server.go
```diff
@@ -165,7 +167,7 @@ func RunServer() {
 			if err != nil {
 				logger.FromContext(ctx).Fatal("datadog.metrics.error", zap.Error(err))
 			}
-			ctx = context.WithValue(ctx, "ddMetrics", ddMetrics)
+			ctx = context.WithValue(ctx, "ddMetrics", ddMetrics) //nolint: staticcheck
```
sven-urbanski-freiheit-com, line 170: what is the issue here? Could you link to the documentation with all the different linters?

ahmed-nour-fdc: This was the error
`SA1029: should not use built-in type string as key for value; define your own type to avoid collisions (staticcheck)`

I defined a `contextKey` type and used it to resolve the error

services/cd-service/pkg/repository/repository.go
ahmed-nour-fdc: This function is only used in tests, so I just chose to ignore it.

services/cd-service/pkg/fs/fs.go
ahmed-nour-fdc: I handled it now.

sven-urbanski-freiheit-com (approved):

PR #1462, opened 2024-03-21
fix: warning when can't find prev or next commit ID

miguel-crespo-fdc:
When previous or next commit id is provided but kuberpult does not find the folder for those commits, previously we were throwing an error. Now we simply log a warning and move on.

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -559,12 +559,12 @@ func writeCommitData(ctx context.Context, sourceCommitId string, sourceMessage s
 	}
 
 	if previousCommitId != "" {
-		if err := writeNextPrevInfo(sourceCommitId, strings.ToLower(previousCommitId), fieldPreviousCommitId, fs); err != nil {
+		if err := writeNextPrevInfo(ctx, sourceCommitId, strings.ToLower(previousCommitId), fieldPreviousCommitId, app, fs); err != nil {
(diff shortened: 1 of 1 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -596,15 +596,17 @@ func writeCommitData(ctx context.Context, sourceCommitId string, sourceMessage s
...
 	otherCommitId = strings.ToLower(otherCommitId)
 	sourceCommitDir := commitDirectory(fs, sourceCommitId)
 
 	otherCommitDir := commitDirectory(fs, otherCommitId)
 
 	if _, err := fs.Stat(otherCommitDir); err != nil {
-		return err
+		logger.FromContext(ctx).Sugar().Warnf(
+			"%s application does not contain any commit with the provided id (%s). Ignoring provided information.", application, otherCommitId)
+		return nil
```
sven-urbanski-freiheit-com, line 609: The app is not part of the `othercommitDir` path. So I don't see why it's relevant, apart from contextual information in the warning.
I'd also like to see the directory name printed.

```suggestion
		logger.FromContext(ctx).Sugar().Warnf(
			"Could not find the previous commit while trying to create a new release for commit %s and application %s. This is expected when `git.enableWritingCommitData` was just turned on, however it should not happen multiple times.", otherCommitId, application, otherCommitDir)
		return nil
```

sven-urbanski-freiheit-com (approved):

PR #1475, opened 2024-03-25
fix: verify validity of next and prev commit IDs

miguel-crespo-fdc:
If next and prev commits were invalid, they could end up crashing cd-service (i.e id = 1)

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -101,8 +101,8 @@ func manifestDirectoryWithReleasesVersion(fs billy.Filesystem, application strin
 }
 
 func commitDirectory(fs billy.Filesystem, commit string) string {
-	if !valid.SHA1CommitID(commit) { //Verification should be
-		return fs.Join("commits", "") //Throw error?
+	if !valid.SHA1CommitID(commit) {
+		return fs.Join("commits", "")
 	}
 	return fs.Join("commits", commit[:2], commit[2:])
 }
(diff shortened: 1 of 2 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -562,12 +562,12 @@ func writeCommitData(ctx context.Context, sourceCommitId string, sourceMessage s
 		return GetCreateReleaseGeneralFailure(err)
 	}
 
-	if previousCommitId != "" {
+	if previousCommitId != "" && valid.SHA1CommitID(previousCommitId) {
```
sven-urbanski-freiheit-com, line 565: could we have a test for this change?

services/cd-service/pkg/repository/transformer.go
```diff
@@ -101,6 +101,9 @@ func manifestDirectoryWithReleasesVersion(fs billy.Filesystem, application strin
 }
 
 func commitDirectory(fs billy.Filesystem, commit string) string {
+	if !valid.SHA1CommitID(commit) {
+		return fs.Join("commits", "")
```
miguel-crespo-fdc, line 105: Should we return something different? Verification seemed missing here as invalid use of this function could lead to backend crashing

sven-urbanski-freiheit-com: I think at this point we need to rely that the caller already checked this.
Either that or we need to return an error here, but that complicates things.

sven-urbanski-freiheit-com, line 105:
```suggestion

```

sven-urbanski-freiheit-com (approved):

PR #1489, opened 2024-03-27
feat: Add allowLongAppNames to support longer application names

### LINKED ISSUE #1182: Allow App names with length > 39

sven-urbanski-freiheit-com:
This introduces the helm parameter `cd.allowLongAppNames`. When true, kuberpult
accepts app names longer than 39, but maximum 70 characters.
This requires some configuration in Argo CD to work.
See https://github.com/freiheit-com/kuberpult/pull/1183#issuecomment-1840790228
for how to enable it in Argo CD.

This was implemented by @jdvgh
In order to get this through our CI, I had to create this new PR.

Old PR: https://github.com/freiheit-com/kuberpult/pull/1183
Closes: #1182


jdvgh:
@sven-urbanski-freiheit-com if I should re-commit with another message just give me a heads-up. I see the commit lint is failing.

sven-urbanski-freiheit-com:
> @sven-urbanski-freiheit-com if I should re-commit with another message just give me a heads-up. I see the commit lint is failing.

that was a bug in the commit linter, should be fixed in a minute: https://github.com/freiheit-com/kuberpult/pull/1492

PR #1491, opened 2024-03-28
feat: queue size datadog tracking

miguel-crespo-fdc:
Kuberpults queue size is now tracked on datadog.
New metric: kuberpult.request_queue_size

### DIFF
--- a/services/cd-service/pkg/repository/transformer_test.go
+++ b/services/cd-service/pkg/repository/transformer_test.go
@@ -6440,7 +6440,8 @@ func TestDatadogQueueMetric(t *testing.T) {
 			}
 
 			if tc.expectedGauges != len(mockClient.gauges) {
-				// Don't compare the value of the gauge, only the number of gauges (concurrency, values may change!, creating a flakey test)
+				// Don't compare the value of the gauge, only the number of gauges,
+				// because we cannot be sure at this point what the size of the queue was during measurement
 				msg := fmt.Sprintf("expected %d gauges but got %d\n",
 					tc.expectedGauges, len(mockClient.gauges))
 				t.Fatalf(msg)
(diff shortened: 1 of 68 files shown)


services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -6386,6 +6387,64 @@ func TestUpdateDatadogMetricsInternal(t *testing.T) {
...
+			repo := setupRepositoryTest(t)
+
+			err := repo.Apply(ctx, tc.transformers...)
+
+			if err != nil {
+				t.Fatalf("Expected no error: %v", err)
+			}
+
+			if tc.expectedGauges != len(mockClient.gauges) {
+				// Don't compare the value of the gauge, only the number of gauges (concurrency, values may change!, creating a flakey test)
```
sven-urbanski-freiheit-com, line 6439: Why is this a problem?
The values should not change at this point in the test.
It seems to work fine for TestUpdateDatadogMetricsInternal.
I would prefer if we do the test the same way here.

miguel-crespo-fdc: The problem is the value of the gauge. There is no way of knowing what the exact value of the queue size will be. It might be that when we gauge the size of the queue after inserting, the concurrent process extracting the elements from the queue already removed the element, thus making the value of the gauged queue size 0 when we insert a new element. It might also happend that an element has not yet been extracted, making the value 1.

The point is: The value of the first gauge cannot safely be assumed to be 1.

miguel-crespo-fdc: The other tests do not use the queue (they use the ApplyTransformersInternal method) so they can safely assume the values of the gauges.

sven-urbanski-freiheit-com: ok, then let's just rephrase it a little, eg.: 
```suggestion
				// Don't compare the value of the gauge, only the number of gauges,
				// because we cannot be sure at this point what the size of the queue was during measurement
```

sven-urbanski-freiheit-com (approved):

PR #1501, opened 2024-04-03
fix: add lock-prevented-deployment event for env locks in trains

sven-urbanski-freiheit-com:
We already had the lock-prevented-deployment event for 3 cases:
* single deployment with app lock
* single deployment with env lock
* release train deployment with app lock

Now we also have it for:
* release train deployment with env lock

This also adds a `FirstLockMessage` field to the prognosis (but not the API) that contains the message from one of the locks that prevented a deployment. We'll extend the lock message in SRX-7CBX2O


services/cd-service/pkg/repository/transformer.go
```diff
@@ -2229,19 +2227,47 @@ func (c *envReleaseTrain) prognosis(
...
 			Error:         err,
+			LockMessage:   "",
 			AppsPrognoses: nil,
 		}
 	}
 	sort.Strings(apps)
 
 	appsPrognoses := make(map[string]ReleaseTrainApplicationPrognosis)
+	if len(envLocks) > 0 {
+		// we don't really care about multiple locks, since they all have the same effect, so we just pick one:
```
khaled-ismaeel, line 2238: Choosing the message from the first lock only feels hacky to me. It would be better in my opinion to have the `LockMessage` field in the application and environment prognoses as the collection of all individual lock messages (as a list). That way we have a _complete_ source of truth in the manifest repo as opposed to an arbitrary _partial_ truth.

If this the motivation behind this is having a small table entry in the commits page, we can handle that in the front end.

sven-urbanski-freiheit-com: The motivation behind it was that in practice it does not matter so much *why* something was not deployed. But I see your point, it is arbitrary to pick just one.
I created a story (SRX-7CBX2O) to handle all locks - and also not just the message but author, etc.

sven-urbanski-freiheit-com: And I will rename the field here.

khaled-ismaeel (approved):

PR #1503, opened 2024-04-04
feat: helm tests

miguel-crespo-fdc:


.github/workflows/release.yml
```diff
@@ -27,6 +27,9 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
```
sven-urbanski-freiheit-com, line 28: please change commit prefix to "fix"

charts/kuberpult/Earthfile
```diff
@@ -88,12 +101,20 @@ test-helm:
...
     
+
     RUN helm lint kuberpult-*.tgz --set git.url=test --set ingress.domainName=kuberpult.example.com
+    RUN pwd
+    COPY tests ./tests
+    COPY values.yaml .
+    COPY templates templates
+    COPY Chart.yaml .
+
+    RUN helm dependency build && cd tests && go mod init example.com/test-helm && go mod tidy && go test
```
sven-urbanski-freiheit-com, line 117: why do we need `example.com/test-helm` here?

charts/kuberpult/tests/charts_test.go
```diff
@@ -0,0 +1,539 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+MIT License for more details.
+
+You should have received a copy of the MIT License
+along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>.
+
+Copyright 2023 freiheit.com*/
+/*
+This file is part of kuberpult.
```
sven-urbanski-freiheit-com, line 17: header is duplicated, please remove

charts/kuberpult/tests/charts_test.go
```diff
@@ -0,0 +1,539 @@
...
+	}{
+		{
+			Name: "Basic Parsing works",
+			Setup: func(t *testing.T, values *ValuesDoc) {
+				values.Git.URL = "testURL"
+				values.Pgp.KeyRing = ""
+				values.Ingress.DomainName = "kuberpult-example.com"
+				values.EnvironmentConfigs.BootstrapMode = true //makes values.EnvironmentConfigs.EnvironmentConfigsJSON needed
+				values.EnvironmentConfigs.EnvironmentConfigsJSON = "{}"
+			},
```
sven-urbanski-freiheit-com, line 343:
```suggestion
			InitialValues: &ValuesDoc{
			/// pseudo code, this needs some more sub-structs to be instantiated.
				values.Git.URL = "testURL"
				values.Pgp.KeyRing = ""
				values.Ingress.DomainName = "kuberpult-example.com"
				values.EnvironmentConfigs.BootstrapMode = true //makes values.EnvironmentConfigs.EnvironmentConfigsJSON needed
				values.EnvironmentConfigs.EnvironmentConfigsJSON = "{}"
			},
```

charts/kuberpult/tests/charts_test.go
```diff
@@ -0,0 +1,539 @@
...
+					Name:  "KUBERPULT_ENABLE_TRACING",
+					Value: "",
+				},
+				{
+					Name:  "DD_TRACE_DEBUG",
+					Value: "",
+				},
+			},
+			ExpectedMissing: []core.EnvVar{},
+			checkValues:     false,
```
sven-urbanski-freiheit-com, line 487: I am not sure why we don't want to check the values here. I would prefer to remove this flag and always check the values.

charts/kuberpult/tests/charts_test.go
```diff
@@ -0,0 +1,539 @@
...
+	"math/rand"
+	"os"
+	"os/exec"
+	k8sYaml "sigs.k8s.io/yaml" //Needed to properly parse the yaml generated by helm, "gopkg.in/yaml.v3" cannot do it properly
+	"strconv"
+	"strings"
+	"testing"
+)
+
+type ValuesDoc struct { //Auto generated using https://zhwt.github.io/yaml-to-go/. In case you made any changes, post the content of values.yaml there and replace this struct
```
sven-urbanski-freiheit-com, line 50: Interesting approach, but I don't want to rely on external tools.
If you have generated code, then it needs to be part of the pipeline and generated automatically. But actually I don't think we need generated code here.

I would suggest to replace this struct and instead just have string that is the `content` of the yaml file. Then you can still easily only set a couple values, don't need to generate code, and the `setup` function is also not required. And we are then also able (in the future) to test for wrong types (string vs bool).

charts/kuberpult/tests/charts_test.go
```diff
@@ -0,0 +1,539 @@
...
+		}
+	}
+	return output, nil
+}
+
+func TestHelmChartsKuberpultCdEnvVariables(t *testing.T) {
+
+	tcs := []struct {
+		Name            string
+		Setup           func(t *testing.T, values *ValuesDoc) //Runs before each test case. Edit the values you want to test here.
```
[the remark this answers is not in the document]
miguel-crespo-fdc: This makes it much harder do initialise de values structure with the values that we want, right? This function makes sense to me as each test case wants to change the values file to have some values and does so through this setup function

charts/kuberpult/Earthfile
miguel-crespo-fdc: @ahmed-nour-fdc took a look at the earthly file and didn't see any problem with it, but maybe we missed something?

sven-urbanski-freiheit-com: let's leave it as it is for now.

.github/workflows/release.yml
```diff
@@ -27,6 +27,9 @@ jobs:
...
           password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run helm chart tests
```
sven-urbanski-freiheit-com, line 30: this should also run on PRs

charts/kuberpult/Earthfile
```diff
@@ -88,12 +101,21 @@ test-helm:
...
     WORKDIR /kp
 
-    COPY (+chart-tarball/kuberpult-*.tgz --VERSION=$VERSION) . 
-    
+    COPY (+chart-tarball/kuberpult-*.tgz --VERSION=$VERSION) .
+
+    COPY (+chart.yaml/Chart.yaml --VERSION=$VERSION) .
+
     RUN helm lint kuberpult-*.tgz --set git.url=test --set ingress.domainName=kuberpult.example.com
+    RUN pwd
```
ahmed-nour-fdc, line 112: We don't need this `pwd` here

PR #1518, opened 2024-04-16
chore: upgrade tracing libraries

sven-urbanski-freiheit-com:


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/freiheit-com/kuberpult
 
-go 1.19
+go 1.21
```
hannesg, line 3: Why not directly go to 1.22?

hannesg (approved):

go.mod
sven-urbanski-freiheit-com: that was actually an accident.
I'm fine with upgrading to 1.22, but I would do this in a separate PR.

PR #1528, opened 2024-04-18
fix: Allow RBAC policy to contain empty Lines

gsandok:
REV: DSN-PKCOP2


sven-urbanski-freiheit-com (approved):
approved and also changed the title from feat to fix

PR #1529, opened 2024-04-19
fix(ci): Optimize CI performance

ahmed-nour-fdc:
This PR includes the following:
  * Do not save built images locally when running in the CI.
  * Cancel stale commits' runs when a new commit is pushed.
  * Install docker with apt in integration tests instead of earthly installing it every time.

### DIFF
--- a/.github/workflows/execution-plan-pr.yml
+++ b/.github/workflows/execution-plan-pr.yml
@@ -3,6 +3,9 @@ name: "Execution Plan on Pull Request"
   pull_request:
     branches:
       - main
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 jobs:
   execution-plan:
     uses: ./.github/workflows/execution-plan-snippet-earthly.yml
--- a/services/cd-service/Makefile
+++ b/services/cd-service/Makefile
@@ -64,11 +64,11 @@ build: bin/main
 
 build-pr:
 	echo "build on pull request"
-	$(EARTHLY) --push +build-pr --registry=$(IMAGE_REGISTRY) --tag=$(VERSION) --mirror=$(ARTIFACT_REGISTRY_MIRROR)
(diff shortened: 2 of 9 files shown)


.github/workflows/execution-plan-pr.yml
```diff
@@ -3,6 +3,9 @@ name: "Execution Plan on Pull Request"
   pull_request:
     branches:
       - main
+concurrency: 
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
sven-urbanski-freiheit-com, line 8: huh, I thought that is already how it worked. I guess I was wrong.

services/cd-service/Makefile
```diff
@@ -64,11 +64,11 @@ build: bin/main
 
 build-pr:
 	echo "build on pull request"
-	$(EARTHLY) --push +build-pr --registry=$(IMAGE_REGISTRY) --tag=$(VERSION) --mirror=$(ARTIFACT_REGISTRY_MIRROR)
+	$(EARTHLY) --no-output --push +build-pr --registry=$(IMAGE_REGISTRY) --tag=$(VERSION) --mirror=$(ARTIFACT_REGISTRY_MIRROR)
```
sven-urbanski-freiheit-com, line 67: does this mean that the image is not pushed to the registry anymore?
That would disrupt @miguel-crespo-fdc's current workflow.

.github/workflows/execution-plan-pr.yml
ahmed-nour-fdc: No it never cancelled runs for stale commits.

services/cd-service/Makefile
ahmed-nour-fdc: It gets pushed, but it does not get saved locally. Check these [logs](https://github.com/freiheit-com/kuberpult/actions/runs/8754669556/job/24027084007#step:12:358)
Images are pushed, but not saved locally.

sven-urbanski-freiheit-com (approved):

PR #1530, opened 2024-04-19
feat: dex on local kind setup

miguel-crespo-fdc:


charts/kuberpult/run-kind.sh
```diff
@@ -315,13 +326,59 @@ datadogProfiling:
...
+$(gsed -e "s/^/    /" <./kuberpult-keyring.gpg)
+auth:
+  dexAuth:
+    enabled: true
+    installDex: true
+    policy_csv: |
+$(for action in CreateLock DeleteLock CreateRelease DeployRelease CreateUndeploy DeployUndeploy CreateEnvironment DeployReleaseTrain; do
+        echo "      Developer, $action, development:*, *, allow"
+      done)
+    clientId: "${iap_clientId}"
```
jdvgh, line 338: the clientId which is set here is not the iap client id, but rather the staticClient which has to be defined in e.g. `.dex.staticClients`

Kuberpult does not know about which oidc provider dex uses, it just gets the claim back.

PR #1550, opened 2024-04-25
feat(frontend): Remove plaintext logging of dexClientSecret

### LINKED ISSUE #1549: [frontend][dex] The clientSecret to connect from kuberpult to dex is exposed in the logs
The secret and id to connect from Kuberpult to DEX are passed to the frontend service as environment variables and there also to the config.
The whole config is logged inside [frontend-service/pkg/cmd/server.go](https://github.com/freiheit-com/kuberpult/blob/main/services/frontend-service/pkg/cmd/server.go#L135). 

Kuberpult version - since dex was added, but also on latest (2.27.0).

How to reconstruct it:
1. Clone the repo
2. Edit `docker-compose-earthly.yml`
3. Add dex environments to the frontend-services env variables:
```
      - KUBERPULT_DEX_CLIENT_ID=1234
      - KUBERPULT_DEX_CLIENT_SECRET=reallysecretdexconnectionsecret
``` 
4. Run `make kuberpult-earthly`
5. Check logs: 
```
kuberpult-frontend-service  | {"level":"warn","ts":1714054532.8153377,"caller":"cmd/server.go:144","msg":"config: \n{cd-service:8443 false  http://kuberpult-cd-service:80    false https://cd.dev.freiheit.systems/ tools  false https://login.microsoftonline.com/    false 1234 reallysecretdexconnectionsecret    https://github.com/freiheit-com/kuberpult/commit/{commit} https://github.com/freiheit-com/fdc-standard-setup-dev-env-manifest/tree/{branch}/{dir} main localhost:* user-local-dev-docker-compose user-local-dev@example.com 2m0s 10m0s false false}"}
```
In #1550 I removed the logging of the secret.
Revolution ID: DSN-R3VBYG

jdvgh:
Closes #1549 
This should now print e.g.:
```log
kuberpult-frontend-service  | {"level":"warn","ts":1714058149.364507,"caller":"cmd/server.go:142","msg":"config: \n{cd-service:8443 false  http://kuberpult-cd-service:80    false https://cd.dev.freiheit.systems/ tools  false https://login.microsoftonline.com/    false 1234 **********    https://github.com/freiheit-com/kuberpult/commit/{commit} https://github.com/freiheit-com/fdc-standard-setup-dev-env-manifest/tree/{branch}/{dir} main localhost:* user-local-dev-docker-compose user-local-dev@example.com 2m0s 10m0s false false}"}

```

I'm open to moving the stripSecrets function to another place - just give me a heads-up on where you want it :) The same holds if you'd like this done in another way.

Revolution ID: DSN-R3VBYG

### DIFF
--- a/services/frontend-service/pkg/cmd/server.go
+++ b/services/frontend-service/pkg/cmd/server.go
@@ -132,7 +132,6 @@ func runServer(ctx context.Context) error {
 		logger.FromContext(ctx).Error("config.parse", zap.Error(err))
 		return err
 	}
-	logger.FromContext(ctx).Warn(fmt.Sprintf("config: \n%v", c))
 	if c.GitAuthorEmail == "" {
 		msg := "DefaultGitAuthorEmail must not be empty"
 		logger.FromContext(ctx).Error(msg)


services/frontend-service/pkg/cmd/server.go
```diff
@@ -125,14 +125,21 @@ func RunServer() {
 	}
 }
 
+func stripSecrets(config config.ServerConfig) config.ServerConfig {
```
jdvgh, line 128: I do not pass by reference so I do not change the `config.ServerConfig` object itself, as we may need the `real` value of `config.ServerConfig.DexClientSecret` later on and only want to hide it in the logs

services/frontend-service/pkg/cmd/server.go
```diff
@@ -125,14 +125,21 @@ func RunServer() {
...
+
 func runServer(ctx context.Context) error {
 	err := envconfig.Process("kuberpult", &c)
 
 	if err != nil {
 		logger.FromContext(ctx).Error("config.parse", zap.Error(err))
 		return err
 	}
-	logger.FromContext(ctx).Warn(fmt.Sprintf("config: \n%v", c))
+	logger.FromContext(ctx).Warn(fmt.Sprintf("config: \n%v", stripSecrets(c)))
```
sven-urbanski-freiheit-com, line 142: I think we can just remove this log line entirely.

sven-urbanski-freiheit-com: alternatively: Just log what it's interesting.
I would say that all boolean values are kinda interesting and the rest can be skipped most likely.

jdvgh: So you propose setting up e.g. a custom `func (config config.ServerConfig) String() string` function where we define what should be logged?

From my perspective the log helped in the past to verify the setup is correct, so I would not drop it. However, manually adding new fields to it in the future is also a little cumbersome. 

So please decide, and I'll implement accordingly :laughing:

jdvgh:
@sven-urbanski-freiheit-com I just removed the line completely. Please re-check.

sven-urbanski-freiheit-com:
replaced by https://github.com/freiheit-com/kuberpult/pull/1572

PR #1552, opened 2024-04-25
chore: refactor db connection setup

sven-urbanski-freiheit-com:
This separates the DBHandler form the DBConfig.
The DBConfig has data related to authentication, so we do not want to pass it around longer than necessary. Also added a Connect function that creates the DBHandler

Ref: SRX-T8YG6D

### DIFF
--- a/services/cd-service/pkg/cmd/server.go
+++ b/services/cd-service/pkg/cmd/server.go
@@ -201,9 +201,9 @@ func RunServer() {
 		}
 
 		if c.DbOption != "NO_DB" {
-			var handler repository.DBHandler
+			var dbCfg repository.DBConfig
 			if c.DbOption == "cloudsql" {
-				handler = repository.DBHandler{
+				dbCfg = repository.DBConfig{
 					DbHost:         c.DbLocation,
 					DbPort:         c.DbAuthProxyPort,
 					DriverName:     "postgres",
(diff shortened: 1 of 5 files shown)


services/cd-service/pkg/cmd/server.go
```diff
@@ -213,43 +213,44 @@ func RunServer() {
...
 					MigrationsPath: c.DbMigrationsLocation,
 				}
 			} else {
 				logger.FromContext(ctx).Fatal("Database was enabled but no valid DB option was provided.")
 			}
-			db, err := handler.GetDBConnection()
+			db, err := repository.Connect(dbCfg)
 			if err != nil {
 				logger.FromContext(ctx).Fatal("Error establishing DB connection: ", zap.Error(err))
 			}
```
miguel-crespo-fdc, line 231: We need to close the connection if the ping fails for whatever reason, the previous version did not do this either, but maybe a defer db.DB.close()

services/cd-service/pkg/cmd/server.go
```diff
@@ -213,43 +213,44 @@ func RunServer() {
...
-			pErr := db.Ping()
+			pErr := db.DB.Ping()
 			if pErr != nil {
 				logger.FromContext(ctx).Fatal("Error pinging DB: ", zap.Error(pErr))
 			}
```
miguel-crespo-fdc, line 235: We need to close the connection if the ping fails for whatever reason, the previous version did not do this either, but maybe a defer db.DB.close()

services/cd-service/pkg/cmd/server.go
```diff
@@ -213,43 +213,44 @@ func RunServer() {
...
-			db.Close()
```
miguel-crespo-fdc, line 236: We need to close the connection if the ping fails for whatever reason, the previous version did not do this either, but maybe a defer db.DB.close()

miguel-crespo-fdc (approved):

PR #1555, opened 2024-04-26
Create cloudrun-service

ahmed-nour-fdc:
* The service currently deploys a hard-coded service.
* The service waits for the deployment to complete and reports back the status of the deployed service.


services/cloudrun-service/pkg/cloudrun/cloudrun.go
```diff
@@ -0,0 +1,177 @@
...
+	serviceRoutesReady         = "RoutesReady"
+)
+
+var (
+	runService *run.APIService
+)
+
+// Possible values are True, False or Unknown
+type ServiceConditions struct {
+	Ready              string
```
sven-urbanski-freiheit-com, line 40: if they can't be `bool`, they should use a "enum":
https://builtin.com/software-engineering-perspectives/golang-enum
something like
```
type ReadyState int

const (// Declare related constants for each weekday starting with index 1
	Ready    ReadyState = iota + 1
	NotReady
        Unknown
// ...
```

go.mod
```diff
@@ -31,18 +31,18 @@ require (
 	go.opentelemetry.io/otel/sdk/metric v1.25.0
 	go.uber.org/zap v1.27.0
 	golang.org/x/crypto v0.22.0
-	golang.org/x/oauth2 v0.16.0
```
sven-urbanski-freiheit-com, line 34: I would prefer library upgrades to be in a separate commit -  or are these upgrades mandatory for the cloud run code in this PR?

services/cloudrun-service/Earthfile
```diff
@@ -0,0 +1,79 @@
+VERSION --use-function-keyword 0.7
+IMPORT ../../infrastructure/earthly/go AS go-build
+
+LOCALLY
+ARG --global service=$(basename $PWD)
+ARG --global src_files=$(find pkg -type f ! -name "*_test.go")
+ARG --global cgo_enabled=0
```
sven-urbanski-freiheit-com, line 7: I would prefer if this was hardcoded, so nobody accidentally adds a cgo library.

services/cloudrun-service/Makefile
```diff
@@ -0,0 +1,95 @@
...
+GARBAGE := *.sentinel
+proto:
+	$(MAKE) -C../../pkg/api all
+
+bin/main: $(ALL_GO_FILES) | proto
+	mkdir -p bin
+	$(GO) build -o bin/main ./cmd/server/
+GARBAGE += bin/main
+
+export KUBERPULT_CDSERVER?=localhost:8443
```
sven-urbanski-freiheit-com, line 42: this looks like it would be long in a docker-compose file, not really in a Makefile

ahmed-nour-fdc: This is actually a copy-paste error. I'll remove this.

services/cloudrun-service/pkg/cloudrun/cloudrun.go
```diff
@@ -0,0 +1,186 @@
...
+
+func GetServiceReadyCondition(serviceCallResponse *run.Service) (ServiceReadyCondition, error) {
+	//exhaustruct:ignore
+	serviceReadyCondition := ServiceReadyCondition{
+		Status:   "",
+		Name:     serviceCallResponse.Metadata.Name,
+		Revision: serviceCallResponse.Status.ObservedGeneration,
+	}
+	conditions := serviceCallResponse.Status.Conditions
+	for _, condition := range conditions {
```
sven-urbanski-freiheit-com, line 126: should we really check all conditions here, or should we just check the last one? Are they not sorted?

services/cloudrun-service/Earthfile
ahmed-nour-fdc: It cannot be hard-coded. All earthfiles use a common [Earthfile](https://github.com/freiheit-com/kuberpult/blob/main/infrastructure/earthly/go/Earthfile) that is configurable. So the use of a global variable here is the only way to disable cgo.

services/cloudrun-service/pkg/cloudrun/cloudrun.go
```diff
@@ -0,0 +1,186 @@
...
+
+func getOperationId(parent string, serviceCallResp *run.Service) (string, error) {
+	operationId, exists := serviceCallResp.Metadata.Annotations[operationIdAnnotation]
+	if !exists {
+		return "", fmt.Errorf("failed to get operation-id for service %s", serviceCallResp.Metadata.Name)
+	}
+	return fmt.Sprintf("%s/operations/%s", parent, operationId), nil
+}
+
+func waitForOperation(parent string, serviceCallResp *run.Service, timeoutSeconds uint16) error {
```
sven-urbanski-freiheit-com, line 147: please use duration instead of an int
```suggestion
func waitForOperation(parent string, serviceCallResp *run.Service, timeout time.Duration) error {
```
https://pkg.go.dev/time#pkg-types

services/cloudrun-service/pkg/cloudrun/cloudrun.go
```diff
@@ -0,0 +1,186 @@
...
+func isPresent(services []*run.Service, serviceName string) bool {
+	for _, service := range services {
+		if service.Metadata.Name == serviceName {
+			return true
+		}
+	}
+	return false
+}
+
+func getParent(svc *run.Service) (string, error) {
```
sven-urbanski-freiheit-com, line 176: is `getParent` the right name? Maybe `getParentLocation`?

services/cloudrun-service/pkg/cloudrun/cloudrun.go
ahmed-nour-fdc: I didn't want to count on the conditions being sorted. I think it is better to make sure that we are ready the correct condition.
Also there are a maximum of 6 conditions, so this does not have any impact on performance.

services/cloudrun-service/pkg/cloudrun/cloudrun_test.go
```diff
@@ -0,0 +1,305 @@
...
+		})
+	}
+}
+
+func TestGetParent(t *testing.T) {
+	for _, test := range []struct {
+		testName       string
+		service        *run.Service
+		expectedParent string
+		expectedError  string
```
sven-urbanski-freiheit-com, line 77: please use `error` here 
```suggestion
		expectedError  error
```
See https://github.com/freiheit-com/kuberpult/blob/main/README_DEVELOPERS.md#always-use-cmpdiff
for how to compare errors.

services/cloudrun-service/pkg/cloudrun/cloudrun.go
ahmed-nour-fdc: It is defined in the [API](https://pkg.go.dev/google.golang.org/api/run/v1#ProjectsLocationsServicesService.Create) as parent.

services/cloudrun-service/pkg/cloudrun/cloudrun_test.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+}
+
+func TestGetOperationId(t *testing.T) {
+	parent := "projects/proj/locations/loc"
+	for _, test := range []struct {
+		testName      string
+		service       *run.Service
+		expectedId    string
+		expectedError string
```
sven-urbanski-freiheit-com, line 264: same here, please use `error`

go.mod
ahmed-nour-fdc: These upgrades came after importing the run api.

services/cloudrun-service/pkg/cloudrun/cloudrun_test.go
```diff
@@ -0,0 +1,307 @@
...
+			expectedParent: "projects/proj_id/locations/europe-west1",
+			expectedError:  nil,
+		},
+	} {
+		testCase := test
+		t.Run(testCase.testName, func(t *testing.T) {
+			t.Parallel()
+			parent, err := getParent(testCase.service)
+			if err != nil {
+				if diff := cmp.Diff(testCase.expectedError, err, cmpopts.EquateErrors()); diff != "" {
```
sven-urbanski-freiheit-com, line 126:
```suggestion
				if diff := cmp.Diff(testCase.expectedError, err, cmpopts.EquateErrors()); diff != "" {
```
I think the first diff is redundant, because cmp.Diff will handle it.

sven-urbanski-freiheit-com (approved):

PR #1559, opened 2024-04-29
feat: default role for dex support

miguel-crespo-fdc:
Kuberpult changes to support Dex. We created a configurable default user that is selected if it is enabled and dex does not return any roles associated with a user after authentication. This is done as we haven't retrieved the groups from google thorugh dex yet.


charts/kuberpult/values.yaml
```diff
@@ -278,6 +278,7 @@ auth:
     enabled: false
     # Indicates if dex is to be installed. If you want to use your own Dex instance do not enable this flag.
     installDex: false
+    defaultUserEnabled: false
```
sven-urbanski-freiheit-com, line 281: Please explain what this new option does in a comment, e.g.
"If kuberpult cannot find a role in the dex response, it will use the role "default".
This is only recommended for when you want the simplest possible setup, or for testing purposes."

Also it should be "defaultRoleEnabled".

pkg/auth/auth.go
```diff
@@ -143,7 +144,15 @@ func (x *DexGrpcContextReader) ReadUserFromGrpcContext(ctx context.Context) (*Us
 	// RBAC Role of the user. only mandatory if DEX is enabled.
 	if x.DexEnabled {
 		rolesInHeader := md.Get(HeaderUserRole)
+
 		if len(rolesInHeader) == 0 {
+			if x.DexDefaultUserEnabled {
+				u.DexAuthContext = &DexAuthContext{
+					Role: "default",
+				}
+				logger.FromContext(ctx).Warn("role undefined but dex is enabled. Default user enabled. Proceeding with default user.")
```
sven-urbanski-freiheit-com, line 153:
```suggestion
				logger.FromContext(ctx).Warn("role undefined but dex is enabled. Default user enabled. Proceeding with default role.")
```

sven-urbanski-freiheit-com (changes requested):
.

charts/kuberpult/values.yaml
```diff
@@ -278,6 +278,9 @@ auth:
     enabled: false
     # Indicates if dex is to be installed. If you want to use your own Dex instance do not enable this flag.
     installDex: false
+    #If kuberpult cannot find a role in the dex response, it will use the role "default".
+    #This is only recommended for when you want the simplest possible setup, or for testing purposes."
```
sven-urbanski-freiheit-com, line 282:
```suggestion
    # If kuberpult cannot find a role in the dex response, it will use the role "default".
    # This is only recommended for when you want the simplest possible setup, or for testing purposes.
```

go.mod
```diff
@@ -31,19 +31,18 @@ require (
 	go.opentelemetry.io/otel/metric v1.26.0
 	go.opentelemetry.io/otel/sdk/metric v1.26.0
 	go.uber.org/zap v1.27.0
-	golang.org/x/crypto v0.22.0
-	golang.org/x/oauth2 v0.19.0
-	google.golang.org/api v0.170.0
-	google.golang.org/grpc v1.62.1
-	google.golang.org/protobuf v1.33.0
+	golang.org/x/crypto v0.23.0
```
sven-urbanski-freiheit-com, line 34: It seems these go.mod upgrades are not really required for this PR to work, therefore we should remove those changes (go.mod and go.sum)

PR #1560, opened 2024-04-29
feat(dex): Add baseUrl for verification

jdvgh:
Hi, 
the dex setup was using the clientSecret instead of the baseURL for the loginInterceptor.
The database did not contain a seperator in the cd-service template.
Furthermore, added a debug level log message for the dex token verification as we do not have any information about what failed otherwise.


charts/kuberpult/templates/cd-service.yaml
```diff
@@ -421,6 +421,7 @@ data:
   policy.csv: {{ .Values.auth.dexAuth.policy_csv | quote}}
 {{- end }}
 {{- if or (eq .Values.cd.db.dbOption "cloudsql") (eq .Values.cd.db.dbOption "sqlite") }}
```
sven-urbanski-freiheit-com, line 423: looks good overall, I will create a PR to mirror this on Thursday.

sven-urbanski-freiheit-com:
Replaced by https://github.com/freiheit-com/kuberpult/pull/1570

PR #1565, opened 2024-04-30
chore: refactor transaction handling for DB

sven-urbanski-freiheit-com:
and one renaming.

### DIFF
--- a/services/cd-service/pkg/repository/db.go
+++ b/services/cd-service/pkg/repository/db.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/freiheit-com/kuberpult/pkg/logger"
+	"go.uber.org/zap"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"path"
 	"reflect"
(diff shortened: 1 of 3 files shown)


services/cd-service/pkg/repository/db.go
```diff
@@ -165,8 +166,34 @@ func PostgresToSqliteQuery(query string) string {
...
+	}
+	defer func(tx *sql.Tx) {
+		err := tx.Rollback()
+		if err != nil {
+			logger.FromContext(ctx).Warn("db.rollback.failed", zap.Error(err))
+		}
+	}(tx)
+
+	err = f(ctx)
+	if err != nil {
```
bjoern-michaelsen-freiheit, line 185: So there are two errors that can happen in one call of WithTransaction:
* one from either f() or Commit()
* one from Rollback()
The latter we mostly ignore. I wonder if we should wrap both in a custom error type containing both errors -- but discarded this as likely overengineered, because even the calling code will unlikely be able to recover from a failed Rollback. However, best it might do is decide to panic, if this was a critical op as the DB will be inconsistent, which isnt much help. => fine with this.

sven-urbanski-freiheit-com: yeah,  I think there's not much we can do to handle a failing Rollback, hence the log. We should consider adding a metric on this, I'm just not quite convinced yet that it's helpful.

PR #1589, opened 2024-05-10
feat: write every transformer request to the DB

sven-urbanski-freiheit-com:
This adds a new "ESL" (event sourcing light) table. This tables stores the input data of every transformer (which is basically every grpc call) in a table.
This is handled outside the transformer. Additionally this database operation is part of a transaction. The transaction is passed to every transformer (which is why this PR is big).

Usually we group parallel requests, so that we need to `git push` only once - however, that does not work with the concept of "committing the transaction once the `git push` succeeded.
Therefore this PR also disables the queuing for requests, if the DB is enabled.

Another smaller fix in here is providing the "Author" because the linter complains (it seems the linter was added after the "Author" change itself).


cd_database/migrations/1714995407000000_event_sourcing_light.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS event_sourcing_light -- aka ESL
+(
+    created TIMESTAMP,
+    event_type VARCHAR(255),
+    json VARCHAR,
+    PRIMARY KEY(created)
```
bjoern-michaelsen-freiheit, line 6: `PRIMARY KEY(created)` Hmm, this means we can only have one event per tick and never a series of events in a batch. Are we sure we'll never need that? I'd prefer a simple increment key, because then we could also rep. ordered events on the same timestamp.

bjoern-michaelsen-freiheit (approved):
see comment

cd_database/migrations/1714995407000000_event_sourcing_light.up.sql
sven-urbanski-freiheit-com: very good point! I added a proper primary key. Decided to go for UUID.

PR #1591, opened 2024-05-13
fix: increase size of json event on database

miguel-crespo-fdc:
While testing on dev-env, we got an error: 'Error inserting information into DB. Error: pq: value too long for type character varying(255)' when creating a new release.

### DIFF
--- a/cd_database/migrations/1714726277027648_events.up.sql
+++ b/cd_database/migrations/1714726277027648_events.up.sql
@@ -4,7 +4,7 @@ CREATE TABLE IF NOT EXISTS events
     timestamp TIMESTAMP,
     commitHash VARCHAR(64),
     eventType VARCHAR(32),
-    json VARCHAR(255),
+    json VARCHAR(1024),
     PRIMARY KEY(uuid)
 );
 
--- a/charts/kuberpult/templates/cd-service.yaml
+++ b/charts/kuberpult/templates/cd-service.yaml
@@ -428,5 +428,5 @@ metadata:
   name: kuberpult-migrations
 data:
   1713956693027648_all_apps.up.sql: "CREATE TABLE IF NOT EXISTS all_apps(version BIGINT, created TIMESTAMP, json VARCHAR(255), PRIMARY KEY(version));"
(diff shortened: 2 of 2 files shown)


cd_database/migrations/1714726277027648_events.up.sql
```diff
@@ -4,7 +4,7 @@ CREATE TABLE IF NOT EXISTS events
     timestamp TIMESTAMP,
     commitHash VARCHAR(64),
     eventType VARCHAR(32),
-    json VARCHAR(255),
+    json VARCHAR(1024),
```
sven-urbanski-freiheit-com, line 7: Do we really know it can't ever be bigger than 1024?
I would prefer to make it unlimited:
```suggestion
    json VARCHAR,
```
```suggestion
    json VARCHAR(1024),
```

charts/kuberpult/templates/cd-service.yaml
```diff
@@ -428,5 +428,5 @@ metadata:
   name: kuberpult-migrations
 data:
   1713956693027648_all_apps.up.sql: "CREATE TABLE IF NOT EXISTS all_apps(version BIGINT, created TIMESTAMP, json VARCHAR(255), PRIMARY KEY(version));"
-  1714726277027648_events.up.sql: "CREATE TABLE IF NOT EXISTS events(uuid VARCHAR(64), timestamp TIMESTAMP, commitHash VARCHAR(64), eventType VARCHAR(32), json VARCHAR(255), PRIMARY KEY(uuid)); CREATE INDEX commitHashIdx on events (commitHash);"
+  1714726277027648_events.up.sql: "CREATE TABLE IF NOT EXISTS events(uuid VARCHAR(64), timestamp TIMESTAMP, commitHash VARCHAR(64), eventType VARCHAR(32), json VARCHAR(1024), PRIMARY KEY(uuid)); CREATE INDEX commitHashIdx on events (commitHash);"
```
sven-urbanski-freiheit-com, line 431: same here

cd_database/migrations/1714726277027648_events.up.sql
```diff
@@ -4,7 +4,7 @@ CREATE TABLE IF NOT EXISTS events
     timestamp TIMESTAMP,
```
sven-urbanski-freiheit-com, line 4: > This isn't done through a migration because sqlite does not support altering table schemas
That's fine for now, but next time we might have already rolled this version out somewhere. So we would need a different approach next time.

sven-urbanski-freiheit-com (approved):

PR #1592, opened 2024-05-13
fix: return proper error code when creating a group lock

sven-urbanski-freiheit-com:


services/cd-service/pkg/service/batch.go
```diff
@@ -370,6 +370,11 @@ func (d *BatchServer) ProcessBatch(
 		case *repository.TeamNotFoundErr:
 			return nil, status.Error(codes.FailedPrecondition, fmt.Sprintf("Could not process ProcessBatch request. Err: %s", applyErr.TransformerError.Error()))
 		default:
+			tmp, ok := status.FromError(applyErr.TransformerError)
+			if tmp != nil && ok {
```
ahmed-nour-fdc, line 374: Why do we check if this `ok` variable is true? What happens if it is false? Doesn't that mean that there is an error and we should return it anyway?

ahmed-nour-fdc (approved):

services/cd-service/pkg/service/batch.go
sven-urbanski-freiheit-com: `ok==false` means that `FromError` could not deduce the status from the given error.
But also `FromError` has an edge-case where it returns (nil, true), which is why we check both.
If `ok` is false, we return the outer error (as before this PR).

PR #1601, opened 2024-05-15
feat: allow writing to only the esl table

sven-urbanski-freiheit-com:
While we are still implementing the database, this feature toggle allows operators to already write the "ESL" (event sourcing light) table.

This is useful for storing the historical data already, and it can be used as a proof that the database setup works, without degrading performance by writing lots of data to lots of tables and also writing to the manifest repo at the same time.

Writing to the manifest repo will later be separated into its own service.


miguel-crespo-fdc (commented):
Seems good , but shouldn't we write a simple unit test that tests this logic?

sven-urbanski-freiheit-com:
> Seems good , but shouldn't we write a simple unit test that tests this logic?

I added helm chart tests and a simple unit test for the new helper functions.

miguel-crespo-fdc (approved):

Issue #1599, opened 2024-05-15
[dex] When running kuberpult behind GCP IAP the requests to `https://kuberpult.com/dex/.well-known/openid-configuration` fail


jdvgh:
Hi,
we are running kuberpult behind an GCP IAP.
When enabling dex, it tries to query the endpoints by accessing the URL `https://kuberpult.com/dex/.well-known/openid-configuration` via GET.
However, this will return the GCP login and not a JSON, so Dex cannot extract the role.

Even though we are able to change the issuer in the call to `oidc.NewProvider` to e.g. `dexServiceURL + issuerPATH`, the page will still return the urls to the different endpoints with a public URL e.g.:
```JSON
{
	"issuer": "https://kuberpult.com/dex",
	"authorization_endpoint": "https://kuberpult.com/dex/auth",
	"token_endpoint": "https://kuberpult.com/dex/token",
	"jwks_uri": "https://kuberpult.com/dex/keys",
	"userinfo_endpoint": "https://kuberpult.com/dex/userinfo",
	"device_authorization_endpoint": "https://kuberpult.com/dex/device/code",
	"introspection_endpoint": "https://kuberpult.com/dex/token/introspect",
	"grant_types_supported": [
		"authorization_code",
		"refresh_token",
		"urn:ietf:params:oauth:grant-type:device_code",
		"urn:ietf:params:oauth:grant-type:token-exchange"
	],
	"response_types_supported": [
		"code"
	],
	"subject_types_supported": [
		"public"
	],
	"id_token_signing_alg_values_supported": [
		"RS256"
	],
	"code_challenge_methods_supported": [
		"S256",
		"plain"
	],
	"scopes_supported": [
		"openid",
		"email",
		"groups",
		"profile",
		"offline_access"
	],
	"token_endpoint_auth_methods_supported": [
		"client_secret_basic",
		"client_secret_post"
	],
	"claims_supported": [
		"iss",
		"sub",
		"aud",
		"iat",
		"exp",
		"email",
		"email_verified",
		"locale",
		"name",
		"preferred_username",
		"at_hash"
	]
}
```

I found a solution to this by creating an own providerConfig and provider which uses the aforementioned cluster-internal paths.

See #1600

PR #1603, opened 2024-05-16
feat(dex): Add cluster-internal dex communication

gsandok:
closes https://github.com/freiheit-com/kuberpult/issues/1599
Recreation of https://github.com/freiheit-com/kuberpult/pull/1600

### DIFF
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -279,6 +279,7 @@ auth:
     # Indicates if dex is to be installed. If you want to use your own Dex instance do not enable this flag.
     installDex: false
     # If using e.g. GCP IAP cluster internal communication to Dex is necessary as otherwise its endpoints cannot be accessed by the frontend service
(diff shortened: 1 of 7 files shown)


charts/kuberpult/values.yaml
```diff
@@ -278,6 +278,8 @@ auth:
     enabled: false
     # Indicates if dex is to be installed. If you want to use your own Dex instance do not enable this flag.
     installDex: false
+    # If using e.g. GCP IAP cluster internal communication to Dex is necessary as otherwise its endpoints cannot be accessed by the frontend service
```
sven-urbanski-freiheit-com, line 281:
```suggestion
    # If using e.g. GCP IAP cluster internal communication to Dex is necessary as otherwise its endpoints cannot be accessed by the frontend service
    # If enabled, kuberpult communicates with dex over http, not https
```

sven-urbanski-freiheit-com (approved):

PR #1608, opened 2024-05-16
chore: introduce manifest-repo-export-service

sven-urbanski-freiheit-com:
At this point the new service only prints hello world.

In the future it will read from the DB and write to the manifest repo.


charts/kuberpult/templates/manifest-repo-export-service.yaml
```diff
@@ -0,0 +1,213 @@
...
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+{{- if eq .Values.cd.db.dbOption "cloudsql" }}
+      serviceAccountName: {{ .Values.cd.db.k8sName }}
+{{- end }}
+      containers:
+{{- if eq .Values.cd.db.dbOption "cloudsql" }}
```
ahmed-nour-fdc, line 65: Is this required for the service?
Also there are a lot of `cd` references in this file, is this correct?

services/manifest-repo-export-service/policy.csv
ahmed-nour-fdc, line 1: Is this file going to be needed in the future?

ahmed-nour-fdc (approved):

charts/kuberpult/templates/manifest-repo-export-service.yaml
sven-urbanski-freiheit-com: It's not technically required right now, but it will be in the next PR.
I do think we should rename `cd.db` to `cd`, but I would prefer to do this later.

services/manifest-repo-export-service/policy.csv
sven-urbanski-freiheit-com: probably not, I will remove it for now.

PR #1612, opened 2024-05-17
fix: do not reset application sync policy on update

leandro-freiheit:
When we do an update using the rollout-service, we always enable the auto-sync. If there is an application that has its auto-sync disabled manually for hotfix purposes, kuberpult will override this almost instantaneously on the next consume of the overview. For this purpose, we should make the diff between applications on the update call more granular, excluding the sync policy.

### DIFF
--- a/services/rollout-service/pkg/argo/argo.go
+++ b/services/rollout-service/pkg/argo/argo.go
@@ -176,7 +176,7 @@ func (a ArgoAppProcessor) CreateOrUpdateApp(ctx context.Context, overview *api.G
 
 			//exhaustruct:ignore
 			emptyAppSpec := v1alpha1.ApplicationSpec{}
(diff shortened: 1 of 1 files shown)


services/rollout-service/pkg/argo/argo.go
```diff
@@ -172,11 +172,18 @@ func (a ArgoAppProcessor) CreateOrUpdateApp(ctx context.Context, overview *api.G
 				Application:          appToUpdate,
 				Project:              ptr.FromString(appToUpdate.Spec.Project),
 			}
-			//We have to exclude the unexported type isServerInferred. It is managed by Argo.
 
 			//exhaustruct:ignore
```
sven-urbanski-freiheit-com, line 176: please remove the word "should" from the PR.

sven-urbanski-freiheit-com (approved):

PR #1615, opened 2024-05-17
fix: introduce generated sequence for esl table

sven-urbanski-freiheit-com:
This is necessary because otherwise there is not total order on the rows of the esl table.

Unfortunately, it is note possible to write an SQL statement that creates a sequence and works for both sqlite and postgres.

Therefore, we split the migrations into sqlite and postgres and work with symlinks for cases where the SQL statements are the same - which is most of the time.


miguel-crespo-fdc (approved):
~~~ Looks good. I've had some problems regarding symbolic links and Earthly, but I guess the pipeline will warn us if my concerns are justified. ~~~ Just checked out the branch and configs seem to be generated without issues

PR #1627, opened 2024-05-23
feat: Allow configuring release versions history limit

ahmed-nour-fdc:


services/cd-service/pkg/cmd/server.go
```diff
@@ -260,7 +270,7 @@ func RunServer() {
 				KnownHostsFile: c.GitSshKnownHosts,
 			},
 			Branch:                 c.GitBranch,
-			GcFrequency:            20,
+			GcFrequency:            c.ReleaseVersionsLimit,
```
sven-urbanski-freiheit-com, line 273: this is the wrong field.
See `keptVersionsOnCleanup` in transformer.go

sven-urbanski-freiheit-com (changes requested):
.

services/cd-service/pkg/cmd/server.go
ahmed-nour-fdc: Done

services/cd-service/pkg/repository/transformer.go
```diff
@@ -1216,10 +1217,10 @@ func findOldApplicationVersions(state *State, name string) ([]uint64, error) {
 		return versions[i] >= oldestDeployedVersion
 	})
 
-	if positionOfOldestVersion < (keptVersionsOnCleanup - 1) {
+	if positionOfOldestVersion < (int(state.ReleaseVersionsLimit) - 1) {
 		return nil, nil
 	}
-	return versions[0 : positionOfOldestVersion-(keptVersionsOnCleanup-1)], err
+	return versions[0 : positionOfOldestVersion-(int(state.ReleaseVersionsLimit)-1)], err
```
khaled-ismaeel, line 1223: Let's add some unit tests for this? There are unit tests which check for the default limit of 20, but not with a custom value for `.ReleaseVersionsLimit`

services/cd-service/pkg/repository/repository.go
```diff
@@ -1443,6 +1452,7 @@ type State struct {
 	Commit                 *git.Commit
 	BootstrapMode          bool
 	EnvironmentConfigsPath string
+	ReleaseVersionsLimit   uint
```
khaled-ismaeel, line 1455: Non-blocking comment:

I understand why this field was added here; it is the only scope to pass parameters from `Repository` down to `Transform`. Still, I think we shouldn't treat a "State" as a catch-all bucket of parameters, but that's for a different PR.

services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: agree, there should be at least one test that calls the transformer and checks that files or releases were deleted

ahmed-nour-fdc: I modified some tests to test for different values of kept versions.

services/cd-service/pkg/repository/repository.go
ahmed-nour-fdc: Unfortunately, like you mentioned, `State` is the only way to pass the configuration to the transformers. It might not be the best solution here, but I don't see any other way without completely refactoring the transformers' parameters.

khaled-ismaeel (approved):

sven-urbanski-freiheit-com (approved):

PR #1638, opened 2024-05-27
fix(db): handle NULL in deployments query and deployments migration

sven-urbanski-freiheit-com:
There were three issues:
1) Querying a deployment did not handle NULL for "releaseVersion" field
2) Custom migration called "GetEnvironmentApplicationVersion" which queried the DB, not the repository
3) The migration script for postgres had a syntax error ("NULLABLE" instead of "NULL")

### DIFF
--- a/pkg/db/db_test.go
+++ b/pkg/db/db_test.go
@@ -339,7 +339,7 @@ func TestReadWriteDeployment(t *testing.T) {
 		ExpectedDeployment *Deployment
 	}{
 		{
-			Name:            "1",
+			Name:            "with eslVersion != nil",
 			App:             "app-a",
 			Env:             "dev",
 			VersionToDeploy: version(7),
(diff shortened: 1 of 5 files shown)


pkg/db/db_test.go
```diff
@@ -321,3 +324,115 @@ func TestHelperFunctions(t *testing.T) {
...
+			VersionToDeploy: version(7),
+			ExpectedDeployment: &Deployment{
+				App:        "app-a",
+				Env:        "dev",
+				EslVersion: 2,
+				Version:    version(7),
+			},
+		},
+		{
+			Name:            "2",
```
miguel-crespo-fdc, line 354: Maybe add some more comprehensive test names? Would be a big help in the review

miguel-crespo-fdc (changes requested):
Had some issues with the migrations when testing other unrelated changes, went back and figured out that [the migrations script that introduces the deployments table](https://github.com/freiheit-com/kuberpult/blob/6f8a3fa6ef391612255c3767be0f5c1c530191d0/database/migrations/sqlite/1716384616708554_deployments.up.sql) is NOT valid psql. The NULLABLE keyword apparently does not exist in psql. Maybe we should address this in another PR.

miguel-crespo-fdc (approved):

PR #1647, opened 2024-05-29
feat: add release train prognoses to the UI

khaled-ismaeel:
This PR adds a UI page for displaying the release train prognosis on a certain environment at `/ui/environments/<env>/releaseTrain`. It is accessible from a link on the environments group called "Release train details", as shown.

| image|
| :---: |
| _figure 1: the environments page with the new link to the release train prognosis page_|

The release train prognosis page contains a table with an entry for every application stating what happens to the application in case a release train triggered on this environment. It looks like this.

| image |
| :---: |
| _figure 2: a sample of the release train prognosis page_ |

There is a special case when the release train on an environment is skipped (for example, because it is locked). In which case no table is rendered and a descriptive message is shown instead.

| image |
| :---: |
| _figure 3: release train prognosis page when an environment is locked_ |


services/frontend-service/src/ui/components/ReleaseTrainPrognosis/ReleaseTrainPrognosis.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+        case ReleaseTrainEnvSkipCause.UPSTREAM_ENV_CONFIG_NOT_FOUND:
+            return (
+                <p>
+                    Release train on this environment is skipped because no configuration was found for it in the
+                    manifest repository.
+                </p>
+            );
+
+        case ReleaseTrainEnvSkipCause.UNRECOGNIZED:
```
miguel-crespo-fdc, line 122: Redundant case

miguel-crespo-fdc (approved):

services/frontend-service/src/ui/components/ReleaseTrainPrognosis/ReleaseTrainPrognosis.tsx
khaled-ismaeel: Indeed, but still I'd rather be explicit than implicit about it.

PR #1648, opened 2024-05-30
fix(db): close rows

sven-urbanski-freiheit-com:
And handle nil in GetDeploymentMetaData

### DIFF
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -1659,7 +1659,9 @@ func (s *State) GetDeploymentMetaData(ctx context.Context, environment, applicat
 		if err != nil {
 			return "", time.Time{}, err
 		}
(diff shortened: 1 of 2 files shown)


services/cd-service/pkg/repository/repository.go
```diff
@@ -1659,7 +1659,9 @@ func (s *State) GetDeploymentMetaData(ctx context.Context, environment, applicat
 		if err != nil {
 			return "", time.Time{}, err
 		}
-		return result.Metadata.DeployedByEmail, result.Created, nil
+		if result != nil {
```
ahmed-nour-fdc, line 1662: What happens if the result is nil? What is returned in that case?

sven-urbanski-freiheit-com: fixed

ahmed-nour-fdc (approved):

PR #1649, opened 2024-05-30
fix(db): add missing env vars for export-service

sven-urbanski-freiheit-com:
this fixes missing env vars,
adds chart tests
and handles the logging initialization better.
The logging with `fmt.Printf` did not work on kubernetes, so we now wrap the whole "Run" function into the logger.

### DIFF
--- a/services/manifest-repo-export-service/pkg/cmd/server.go
+++ b/services/manifest-repo-export-service/pkg/cmd/server.go
@@ -41,6 +41,8 @@ func storageBackend(enableSqlite bool) repository.StorageBackend {
 
 func RunServer() {
 	err := logger.Wrap(context.Background(), func(ctx context.Context) error {
+		logger.FromContext(ctx).Warn("DEBUG: RunServer startup")
+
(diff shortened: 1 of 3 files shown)


khaled-ismaeel (approved):

services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -40,168 +40,175 @@ func storageBackend(enableSqlite bool) repository.StorageBackend {
...
-					if err != nil {
-						return fmt.Errorf("error in DBWriteCutoff %v", err)
-					}
+		repo, err := repository.New(ctx, cfg)
+		if err != nil {
+			return fmt.Errorf("repository.new failed %v", err)
+		}
 
+		log := logger.FromContext(ctx).Sugar()
+		for {
```
khaled-ismaeel, line 168: Non-blocking comment:

Should be worried of some log flood here?

sven-urbanski-freiheit-com: a little bit, yes. We'll take care of that later.

PR #1651, opened 2024-05-31
feat: esl environment lock support 

miguel-crespo-fdc:
Environment lock creation and deletion is now stored on the database by the cd-service. The export manifest service now also also deals with deletion and creation of theses locks


pkg/db/db.go
```diff
@@ -757,6 +805,216 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+	rows, err := tx.QueryContext(
+		ctx,
+		selectQuery,
+		environment,
+		lockID,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("could not query esl table from DB. Error: %w\n", err)
+	}
+	if rows.Next() {
```
sven-urbanski-freiheit-com, line 825: we also need to call `rows.Close` (in a defer) and `rows.Err` after using the row.

pkg/db/db.go
```diff
@@ -757,6 +805,216 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+	}
+	span, _ := tracer.StartSpanFromContext(ctx, "DBSelectEnvLocks")
+	defer span.Finish()
+
+	selectQuery := h.AdaptQuery(
+		fmt.Sprintf(
+			"SELECT eslVersion, created, lockID, envName, metadata" +
+				" FROM environment_locks " +
+				" WHERE envName=? " +
+				" ORDER BY eslVersion DESC " +
```
sven-urbanski-freiheit-com, line 947: this will not get you the last 25 locks, it will give you the last 25 changes in locks.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -120,14 +120,12 @@ func commitEventDir(fs billy.Filesystem, commit, eventId string) string {
...
-func GetEnvironmentLocksCount(fs billy.Filesystem, env string) float64 {
-	envLocksCount := 0
-	envDir := environmentDirectory(fs, env)
-	locksDir := fs.Join(envDir, "locks")
-	if entries, _ := fs.ReadDir(locksDir); entries != nil {
-		envLocksCount += len(entries)
+func (s *State) GetEnvironmentLocksCount(ctx context.Context, env string) float64 {
+	locks, err := s.GetEnvironmentLocks(ctx, env)
+	if err != nil {
+		return 0
```
sven-urbanski-freiheit-com, line 126: please do not ignore errors.
Ideally, return the error from this function and handle it in `GaugeEnvLockMetric` by logging

sven-urbanski-freiheit-com (changes requested):
.

pkg/db/db.go
miguel-crespo-fdc: Isn't this the same thing? Recorded changes  to the environment_locks table are just insertions. Deletions remove the locks from the table altogether.  In my eyes this query returns the last 25 locks that were created on a certain environment (and still exist)

pkg/db/db.go
```diff
@@ -757,6 +805,228 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+	if h == nil {
+		return nil
+	}
+	if tx == nil {
+		return fmt.Errorf("DBDeleteEnvironmentLock: no transaction provided")
+	}
+	span, _ := tracer.StartSpanFromContext(ctx, "DBDeleteEnvironmentLock")
+	defer span.Finish()
+
+	deleteStatement := h.AdaptQuery("DELETE FROM environment_locks WHERE lockID=(?) AND envName=(?);")
```
sven-urbanski-freiheit-com, line 1016: we do not want to delete anything - we store everything forever.
Instead, we need to insert a new row here that has a flag "deleted=true".

pkg/db/db.go
```diff
@@ -793,26 +905,494 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+	}
+	err = rows.Err()
+	if err != nil {
+		return nil, fmt.Errorf("environment locks: row has error: %v\n", err)
+	}
+	return nil, nil // no rows, but also no error
+
 }
 
-type AllApplicationsRow struct {
```
miguel-crespo-fdc, line 800: Re-arranged code for methods related to some structs are closer together

services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -212,7 +212,7 @@ func readEslEvent(ctx context.Context, transaction *sql.Tx, eslId *db.EslId, log
 	if eslId == nil {
 		log.Warnf("no cutoff found, starting at the beginning of time.")
 		// no read cutoff yet, we have to start from the beginning
-		esl, err := dbHandler.DBReadEslEventInternal(ctx, transaction, false)
+		esl, err := dbHandler.DBReadEslEventInternal(ctx, transaction, true)
```
miguel-crespo-fdc, line 215: first row needs to be true for when we start at the beggining of time

pkg/db/db.go
```diff
@@ -240,6 +240,31 @@ func WithTransactionT[T any](h *DBHandler, ctx context.Context, f DBFunctionT[T]
 	return result, nil
 }
 
+type DBFunctionMultipleEntriesT[T any] func(ctx context.Context, transaction *sql.Tx) ([]T, error)
+
+// WithTransactionMultipleEntriesT is the same as WithTransaction, but you can also return and array of data, not just the error.
+func WithTransactionMultipleEntriesT[T any](h *DBHandler, ctx context.Context, f DBFunctionMultipleEntriesT[T]) ([]T, error) {
```
sven-urbanski-freiheit-com, line 246: this may be a little nitpick-y:
I would prefer if we had only 1 function to handle transactions e.g. calling `BegintTx`.

Idea:
We could implement `WithTransactionT` like this:
```
	res, err := WithTransactionMultipleEntriesT(h, ctx, func(ctx context.Context, transaction *sql.Tx) ([]T, error) {
		fRes, err2 := f(ctx, transaction)
		if err2 != nil {
			return nil, err2
		}
		return []T{*fRes}, nil
	})
	return &res[0], err
```
(It compiles, but I haven't run it)

pkg/db/db.go
```diff
@@ -793,26 +905,495 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+	if rows.Next() {
+
+		err := rows.Scan(&row.EslVersion, &row.Created, &row.LockID, &row.Env, &row.Metadata, &row.Deleted)
+		if err != nil {
+			if errors.Is(err, sql.ErrNoRows) {
+				return nil, nil
+			}
+			return nil, fmt.Errorf("Error scanning environment lock row from DB. Error: %w\n", err)
+		}
+		return &row, nil
```
sven-urbanski-freiheit-com, line 938: we should also call `close` and `Err` in this case.

pkg/db/db.go
```diff
@@ -793,26 +905,495 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+		if err != nil {
+			return nil, fmt.Errorf("Error during json unmarshal. Error: %w. Data: %s\n", err, row.Data)
+		}
+		var resultGo = AllEnvLocksGo{
+			Version:         row.Version,
+			Created:         row.Created,
+			Environment:     row.Environment,
+			AllEnvLocksJson: AllEnvLocksJson{EnvLocks: resultJson.EnvLocks},
+		}
+		return &resultGo, nil
```
sven-urbanski-freiheit-com, line 1226: here we should also run `Close` and `Err`

pkg/db/db.go
```diff
@@ -793,26 +905,495 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+	span, _ := tracer.StartSpanFromContext(ctx, "DBSelectEnvironmentLockSet")
+	defer span.Finish()
+	//This is needed as lock ids with '-' might confuse sql and you can pass a string array in query
+	for i, l := range lockIDs {
+		lockIDs[i] = "'" + l + "'" //Assuming no
+	}
+	selectQuery := h.AdaptQuery(
+		"SELECT eslVersion, created, lockID, envName, metadata, deleted" +
+			" FROM environment_locks " +
+			" WHERE envName=? AND lockID in (" + strings.Join(lockIDs, ",") + ")" +
```
sven-urbanski-freiheit-com, line 1255: Why do we need the lockId in the `where` clause?
Just `where envName=?` should be enough to get a unique row.

The limit should be `1` .

pkg/db/db.go
```diff
@@ -793,26 +905,495 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+
+	if existingEnvLock == nil {
+		previousVersion = 0
+	} else {
+		previousVersion = existingEnvLock.EslVersion
+	}
+
+	//WE simply want to increment the esl version and set the deleted field to 1
+	insertQuery := h.AdaptQuery(
+		"INSERT INTO environment_locks (eslVersion, created, lockID, envName, deleted, metadata) SELECT ?, ?, lockID, envName, ?, metadata FROM environment_locks WHERE eslVersion=? AND envName=? AND lockID=?;")
```
sven-urbanski-freiheit-com, line 1365: I think this can be simplified.
We already obtained the last lock in `existingEnvLock`.
This should give us everything we need.
So this statement could then be a simple `INSERT` without the `SELECT`.

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -21,24 +21,27 @@ import (
...
-	billy "github.com/go-git/go-billy/v5"
-	"github.com/go-git/go-billy/v5/util"
+	"time"
 )
 
 const (
-	queueFileName = "queued_version"
+	queueFileName       = "queued_version"
+	fieldCreatedAt      = "created_at"
+	fieldcreatedByName  = "created_by_name"
```
sven-urbanski-freiheit-com, line 42: consistent naming please:
```suggestion
	fieldCreatedByName  = "created_by_name"
```

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -21,24 +21,27 @@ import (
...
+	fieldcreatedByEmail = "created_by_email"
```
sven-urbanski-freiheit-com, line 43:
```suggestion
	fieldCreatedByEmail = "created_by_email"
```

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -351,3 +354,124 @@ func (c *DeployApplicationVersion) Transform(
...
+	return db.EvtDeleteEnvironmentLock
+}
+
+func (c *DeleteEnvironmentLock) Transform(
+	ctx context.Context,
+	state *State,
+	_ TransformerContext,
+	_ *sql.Tx,
+) (string, error) {
+	//read event from DB
```
sven-urbanski-freiheit-com, line 451:
```suggestion
```

Outdated comment?

sven-urbanski-freiheit-com (changes requested):
.

pkg/db/db.go
miguel-crespo-fdc: This function is used to return a set of locks that correspond to the lockIds provided as input. The use case is to filter the table by deleted locks so: we query the all locks table, get the 'active' locks and then filter the environment locks table for these active locks, so we can get the whole lock information. 
And I think the limit should be the length of lockIDS

pkg/db/db.go
miguel-crespo-fdc: Thanks, was trying to do something similar but after a good hour fighting the compiler I decided to go with this solution. After some tweaking, got a similar solution working :)

pkg/db/db.go
```diff
@@ -793,26 +913,475 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+	//This is needed as lock ids with '-' might confuse sql and you can pass a string array in query
+	for i, l := range lockIDs {
+		lockIDs[i] = "'" + l + "'" //Assuming no
+	}
+	selectQuery := h.AdaptQuery(
+		"SELECT eslVersion, created, lockID, envName, metadata, deleted" +
+			" FROM environment_locks " +
+			" WHERE envName=? AND lockID in (" + strings.Join(lockIDs, ",") + ")" +
+			" ORDER BY eslVersion DESC " +
+			" LIMIT " + strconv.Itoa(len(lockIDs)) + ";")
```
sven-urbanski-freiheit-com, line 1260: instead of using `limit`, we should log an error/warning if the number is not equal the expected number.

pkg/db/db.go
```diff
@@ -793,26 +913,475 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+
+	rows, err := tx.QueryContext(ctx, selectQuery, environment)
```
sven-urbanski-freiheit-com, line 1262: the query needs to be getting the latests eslVersion per lock. I think this needs to be done by one SELECT per lockId

pkg/db/db.go
```diff
@@ -793,26 +913,475 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+		envLock.Deleted,
+		jsonToInsert)
+
+	if err != nil {
+		return fmt.Errorf("could not write environment lock into DB. Error: %w\n", err)
+	}
+	return nil
+}
+
+func (h *DBHandler) DBSelectEnvLocks(ctx context.Context, tx *sql.Tx, environmentName string) ([]EnvironmentLock, error) {
```
sven-urbanski-freiheit-com, line 1109:
```suggestion
// DBSelectEnvLocks is only used for testing
func (h *DBHandler) DBSelectEnvLocks(ctx context.Context, tx *sql.Tx, environmentName string) ([]EnvironmentLock, error) {
```

pkg/db/db.go
```diff
@@ -760,6 +816,69 @@ func (h *DBHandler) RunCustomMigrationDeployments(ctx context.Context, getAllDep
...
+func (h *DBHandler) RunCustomMigrationEnvLocks(ctx context.Context, getAllEnvLocksFun GetAllEnvLocksFun) error {
+	return h.WithTransaction(ctx, func(ctx context.Context, transaction *sql.Tx) error {
+		l := logger.FromContext(ctx).Sugar()
+		allEnvLocksDb, err := h.DBSelectAnyEnvLock(ctx, transaction)
+		if err != nil {
+			l.Warnf("could not get environment locks from database - assuming the manifest repo is correct: %v", err)
+			allEnvLocksDb = nil
+		}
+		if allEnvLocksDb != nil {
+			l.Warnf("There are already environment locks in the DB - skipping migrations")
```
sven-urbanski-freiheit-com, line 852:
```suggestion
			l.Infof("There are already environment locks in the DB - skipping migrations")
```

pkg/db/db.go
```diff
@@ -760,6 +816,69 @@ func (h *DBHandler) RunCustomMigrationDeployments(ctx context.Context, getAllDep
...
+	EventType  event.EventType
+	EventJson  string
+}
+
+func (h *DBHandler) RunCustomMigrationEnvLocks(ctx context.Context, getAllEnvLocksFun GetAllEnvLocksFun) error {
+	return h.WithTransaction(ctx, func(ctx context.Context, transaction *sql.Tx) error {
+		l := logger.FromContext(ctx).Sugar()
+		allEnvLocksDb, err := h.DBSelectAnyEnvLock(ctx, transaction)
+		if err != nil {
+			l.Warnf("could not get environment locks from database - assuming the manifest repo is correct: %v", err)
```
sven-urbanski-freiheit-com, line 848:
```suggestion
			l.Infof("could not get environment locks from database - assuming the manifest repo is correct: %v", err)
```

pkg/db/db.go
```diff
@@ -793,26 +912,486 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+	}
+	if tx == nil {
+		return fmt.Errorf("DBWriteEnvironmentLock: no transaction provided")
+	}
+	span, _ := tracer.StartSpanFromContext(ctx, "DBWriteEnvironmentLock")
+	defer span.Finish()
+
+	var previousVersion EslId
+
+	//See if there is an existing lock with the same lock id in this environment. If it exists, just add a +1 to the eslversion
```
sven-urbanski-freiheit-com, line 1040: this comment seems outdated

pkg/db/db.go
```diff
@@ -793,26 +912,486 @@ func (h *DBHandler) RunCustomMigrationAllTables(ctx context.Context, getAllAppsF
...
+				Created:    row.Created,
+				LockID:     row.LockID,
+				Env:        row.Env,
+				Deleted:    row.Deleted,
+				Metadata:   resultJson,
+			})
+		}
+
+	}
+	err := closeRows(rows)
```
sven-urbanski-freiheit-com, line 1317: this just closes the last row, but we need to close all rows.

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -351,3 +354,123 @@ func (c *DeployApplicationVersion) Transform(
...
+	return db.EvtCreateEnvironmentLock
+}
+
+func (c *CreateEnvironmentLock) Transform(
+	ctx context.Context,
+	state *State,
+	_ TransformerContext,
+	transaction *sql.Tx,
+) (string, error) {
+	//read event from DB
```
sven-urbanski-freiheit-com, line 375:
```suggestion
```
this comment seems outdated

PR #1652, opened 2024-06-04
fix(db): close rows properly

sven-urbanski-freiheit-com:
and check for errors.
This also removes some unnecessary logging.

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -310,7 +310,7 @@ func (h *DBHandler) DBReadEslEventInternal(ctx context.Context, tx *sql.Tx, firs
 		selectQuery,
 	)
 	if err != nil {
-		return nil, fmt.Errorf("could not query esl table from DB. Error: %w\n", err)
+		return nil, fmt.Errorf("could not query event_sourcing_light table from DB. Error: %w\n", err)
 	}
 	defer func(rows *sql.Rows) {
 		err := rows.Close()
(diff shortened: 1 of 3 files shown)


pkg/db/db.go
```diff
@@ -310,71 +310,84 @@ func (h *DBHandler) DBReadEslEventInternal(ctx context.Context, tx *sql.Tx, firs
...
 				return nil, nil
 			}
 			return nil, fmt.Errorf("Error scanning event_sourcing_light row from DB. Error: %w\n", err)
 		}
-		logger.FromContext(ctx).Sugar().Warnf("read row: %v", row)
-		return &row, nil
 	}
-	return nil, nil // no rows, but also no error
+	err = rows.Close()
+	if err != nil {
```
miguel-crespo-fdc, line 337: Isn't this rows.close (and dealing with the error) treated on the defer function?

pkg/db/db.go
```diff
@@ -571,53 +596,58 @@ func (h *DBHandler) DBSelectDeployment(ctx context.Context, tx *sql.Tx, appSelec
 		envSelector,
 	)
 	if err != nil {
-		return nil, fmt.Errorf("could not query esl table from DB. Error: %w\n", err)
+		return nil, fmt.Errorf("could not deployments esl table from DB. Error: %w\n", err)
```
khaled-ismaeel, line 599: typo?

pkg/db/db.go
sven-urbanski-freiheit-com: Yes, but in a `defer`, we don't have a good way to return the error to the caller.
Therefore we use `defer` here only as a "catch all", and the proper error handling happens outside of `defer`.

PR #1659, opened 2024-06-07
chore(db): migrate releases table

sven-urbanski-freiheit-com:
This only takes care of the custom migrations for the releases tables. It does not use the table yet, except during startup.


pkg/db/db.go
```diff
@@ -638,6 +783,7 @@ func (h *DBHandler) RunCustomMigrations(
 	ctx context.Context,
 	getAllAppsFun GetAllAppsFun,
 	getAllDeploymentsFun GetAllDeploymentsFun,
+	getAllReleasesFun GetAllReleasesFun,
```
khaled-ismaeel, line 786: Some of these functions read from the manifest repo and some read from the database. Let's put an explanatory comment for this as it's a bit unintuitive (and perhaps something we need to fix in the future).

khaled-ismaeel (approved):

PR #1663, opened 2024-06-10
chore: set up integration tests with export service

khaled-ismaeel:


services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -97,6 +101,18 @@ func Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	var eslProcessingBackoff uint64
+	if val, exists := os.LookupEnv("KUBERPULT_ESL_PROCESSING_BACKOFF"); !exists {
```
miguel-crespo-fdc, line 106: Why are we doing this in a different manner than previously? Not saying any individual one is better, but maybe we should strive for consistency

miguel-crespo-fdc (approved):

PR #1666, opened 2024-06-11
respect browser setting and add dark mode scss

malteehrlen:
I actually didnt manage to run this project, so this is an untested draft until I can get the frontend running locally

Example:
image


malteehrlen:
@sven-urbanski-freiheit-com can you tell me if this is something I should pursue or not?

sven-urbanski-freiheit-com:
@malte-ehrlen dark mode is already merged. Do we still need need this PR?

PR #1668, opened 2024-06-13
chore(db): Migration releases to Database 

sven-urbanski-freiheit-com:
Ref: SRX-L070MM


database/migrations/sqlite/1717749007994908_releases.up.sql
```diff
@@ -4,8 +4,17 @@ CREATE TABLE IF NOT EXISTS releases
     releaseVersion INTEGER, -- release ID given by the client that calls CreateApplicationVersion
```
khaled-ismaeel, line 4: Let's also add for PostgreSQL?

database/migrations/sqlite/1717749007994908_releases.up.sql
```diff
@@ -4,8 +4,17 @@ CREATE TABLE IF NOT EXISTS releases
...
-    envName VARCHAR,
-    manifest VARCHAR,
+    manifests VARCHAR,
     metadata VARCHAR,
-    PRIMARY KEY(eslVersion, releaseVersion, appName, envName)
+    PRIMARY KEY(eslVersion, releaseVersion, appName)
+);
+
+
+CREATE TABLE IF NOT EXISTS all_releases
```
khaled-ismaeel, line 13: Do we document these tables and their columns anywhere? If not, maybe this is the right place for it. Let's add an explanatory comment to these fields, at least the problematic JSON blobs.

pkg/ptr/ptr.go
```diff
@@ -37,3 +37,12 @@ func ToUint64(u *uint64) uint64 {
 	}
 	return *u
 }
+
```
khaled-ismaeel, line 40: Maybe we should rename the package to "util", or "conv", since this function clearly doesn't involve any "ptr"s.

services/cd-service/pkg/cmd/server.go
```diff
@@ -336,39 +336,47 @@ func RunServer() {
...
+						DisplayVersion:  repoRelease.DisplayVersion,
+						Manifests:       manifestsMap,
 					}
 				}
 				return result, nil
 			}
-			migErr := dbHandler.RunCustomMigrations(ctx, getAppsAndTeams, repo.State().GetCurrentlyDeployed, getAllReleases, repo.State().GetCurrentEnvironmentLocks, repo.State().GetCurrentApplicationLocks)
+			migErr := dbHandler.RunCustomMigrations(
+				ctx,
+				getAppsAndTeams,
```
khaled-ismaeel, line 374: I don't see why `getAppsAndTeams` and `getAllReleases` are different from other callbacks passed to `RunCustomMigrations`, in my understanding, they all have the same "nature". If there is indeed nothing particular about them, I suggest to reflect that in this function call here.

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -4538,7 +4571,7 @@ func TestTransformer(t *testing.T) {
 						t.Errorf("unexpected version: expected 3, actual %d", *prodVersion)
 					}
 					checkReleaseExists := func(v uint64) {
-						_, err := s.GetApplicationRelease("test", v)
+						_, err := s.GetApplicationReleaseFromManifest("test", v)
```
khaled-ismaeel, line 4574: Should we add tests that test for reading from the database? I would also suggest they don't be separate tests, but we use both in the same test and compare them.

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -242,29 +235,26 @@ func (c *DeployApplicationVersion) Transform(
 	t TransformerContext,
 	transaction *sql.Tx,
 ) (string, error) {
-	fs := state.Filesystem
+	fsys := state.Filesystem
```
khaled-ismaeel, line 238: Just curious: why are we using `fsys` instead of `fs` all of the sudden? 😄

pkg/db/db.go
```diff
@@ -431,6 +431,9 @@ func (h *DBHandler) DBReadEslEventLaterThan(ctx context.Context, tx *sql.Tx, esl
 }
 
 // RELEASES
+// Releases work a bit different, because they are already immutable.
+// We still store the eslId for consistency with other tables,
+// but technically it's not required here.
 
 type DBRelease struct {
```
khaled-ismaeel, line 438: This type is unused (and kind of misleading). Let's delete it.

pkg/db/db.go
```diff
@@ -445,18 +448,32 @@ type DBReleaseMetaData struct {
...
 	ReleaseNumber uint64
+	Created       time.Time
 	App           string
-	Env           string
-	Manifest      string
+	Manifests     DBReleaseManifests
 	Metadata      DBReleaseMetaData
 }
 
+type DBAllReleaseMetaData struct {
```
khaled-ismaeel, line 464: I find it hard to understand this notion of "metadata". Maybe we should rename this? If not, let's add an explanatory comment for why "metadata" isn't used traditionally here.

pkg/db/db.go
```diff
@@ -766,25 +892,20 @@ type ReleaseWithManifest struct {
...
 }
 
-// During the "CustomMigrations" we read from the manifest repo, and write to the database.
-// The functions here are there to retrieve data, so they should not need to access the DB.
-// Therefore, they should not need a "transaction" parameter.
-// There are currently some exceptions, like GetAllDeploymentsFun. This will be changed in SRX-PA568W.
-
 type AllDeployments []Deployment
 type AllEnvLocks map[string][]EnvironmentLock
+type AllReleases map[uint64]ReleaseWithManifest
```
khaled-ismaeel, line 900: What's key and value here?

services/cd-service/pkg/repository/repository.go
```diff
@@ -2188,7 +2187,23 @@ func (s *State) GetCurrentApplicationLocks(ctx context.Context) (db.AllAppLocks,
...
+			return nil, fmt.Errorf("could not get all releases of app %s (nil)", application)
+		}
+		res := ptr.ToUint64Slice(app.Metadata.Releases)
+		return res, nil
+	} else {
+		return s.GetAllApplicationReleasesFromFile(application)
+	}
+}
+
+func (s *State) GetAllApplicationReleasesFromFile(application string) ([]uint64, error) {
```
khaled-ismaeel, line 2206: We call such function `getSomethingFromManifest` elsewhere. Is there any difference between `FromManifest` and `FromFile`? If not, let's rename this.

services/cd-service/pkg/repository/repository.go
```diff
@@ -2324,14 +2362,58 @@ func (s *State) GetApplicationRelease(application string, version uint64) (*Rele
...
+			manifests[entry.Name()] = &api.Manifest{
+				Environment: entry.Name(),
+				Content:     string(content),
+			}
+		}
+		return manifests, nil
+	}
+}
+
+func (s *State) GetApplicationReleaseManifestsFromFile(application string, version uint64) (map[string]*api.Manifest, error) {
```
khaled-ismaeel, line 2409: Same comment as above.

services/cd-service/pkg/repository/repository.go
```diff
@@ -2324,14 +2362,58 @@ func (s *State) GetApplicationRelease(application string, version uint64) (*Rele
...
+		}
+		for index, mani := range release.Manifests.Manifests {
+			manifests[index] = &api.Manifest{
+				Environment: index,
+				Content:     mani,
+			}
+		}
+		return manifests, nil
+	} else {
+		dir := manifestDirectoryWithReleasesVersion(s.Filesystem, application, version)
```
khaled-ismaeel, line 2380: This `else` looks identical to `GetApplicationReleaseManifestsFromFile`, is this intentional? If not, let's consolidate them.

database/migrations/sqlite/1717749007994908_releases.up.sql
sven-urbanski-freiheit-com: it's just a symlink, so it didn't need any changes. But it exists already.

services/cd-service/pkg/repository/transformer_test.go
sven-urbanski-freiheit-com: I wouldn't want to do this here. The tests here are not uniform at all (they're not really table driven tests).
I will add more tests to the transformer_db_test.go though.

services/manifest-repo-export-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: because fs is the same as the name of the import, so goland marks it as a warning.

database/migrations/sqlite/1717749007994908_releases.up.sql
sven-urbanski-freiheit-com: I added some comments

pkg/ptr/ptr.go
sven-urbanski-freiheit-com: done

services/cd-service/pkg/cmd/server.go
sven-urbanski-freiheit-com: I made them all the same

services/cd-service/pkg/repository/transformer_test.go
sven-urbanski-freiheit-com: done, see transfomer_db_test.go, I added "expectedDbReleases"

pkg/db/db.go
sven-urbanski-freiheit-com: I didn't come up with a good alternative name.
I would say let's do this in another PR.

pkg/db/db.go
sven-urbanski-freiheit-com: done

services/cd-service/pkg/repository/repository.go

services/cd-service/pkg/repository/repository.go

services/cd-service/pkg/repository/repository.go

pkg/ptr/ptr_test.go
```diff
@@ -0,0 +1,57 @@
...
+Copyright freiheit.com*/
+
+package ptr
+
+import (
+	"github.com/google/go-cmp/cmp"
+	"testing"
+)
+
+func TestSlice(t *testing.T) {
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: That would be really hard to achieve, because the only way to add a release is via createApplicationVersion, and that expects a uint64.
Tests are never bad, but I think these would be lower priority.

pkg/db/db.go
sven-urbanski-freiheit-com: done

database/migrations/sqlite/1717749007994908_releases.up.sql
khaled-ismaeel: oh I see, the symlink already existed before so I didn't see it in the PR diff.

khaled-ismaeel (approved):

PR #1681, opened 2024-06-20
fix(db): render argo manifests correctly

sven-urbanski-freiheit-com:
The manifest-export service accidentally used the manifest repo to gather data. Now it uses the database.

### DIFF
--- a/services/manifest-repo-export-service/pkg/repository/repository.go
+++ b/services/manifest-repo-export-service/pkg/repository/repository.go
@@ -730,7 +730,7 @@ func (r *repository) updateArgoCdApps(ctx context.Context, transaction *sql.Tx, 
 	span, ctx := tracer.StartSpanFromContext(ctx, "updateArgoCdApps")
 	defer span.Finish()
 	fs := state.Filesystem
-	if apps, err := state.GetEnvironmentApplications(env); err != nil {
+	if apps, err := state.GetEnvironmentApplications(ctx, transaction, env); err != nil {
 		return err
 	} else {
 		spanCollectData, _ := tracer.StartSpanFromContext(ctx, "collectData")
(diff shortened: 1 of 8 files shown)


services/manifest-repo-export-service/pkg/repository/repository.go
```diff
@@ -1148,8 +1148,16 @@ func (s *State) DeleteQueuedVersionIfExists(environment string, application stri
 	return s.DeleteQueuedVersion(environment, application)
 }
 
-func (s *State) GetEnvironmentApplicationVersion(environment, application string) (*uint64, error) {
-	return s.readSymlink(environment, application, "version")
+func (s *State) GetEnvironmentApplicationVersion(ctx context.Context, transaction *sql.Tx, environment, application string) (*uint64, error) {
```
miguel-crespo-fdc, line 1151: We should keep this functionality for reading from the manifest for kuberpult deployments that don't use the database

services/manifest-repo-export-service/pkg/repository/repository.go
```diff
@@ -1288,9 +1296,15 @@ func (s *State) GetEnvironmentConfigsForGroup(envGroup string) ([]string, error)
 	return groupEnvNames, nil
 }
 
-func (s *State) GetEnvironmentApplications(environment string) ([]string, error) {
```
miguel-crespo-fdc, line 1291: Same as before

services/manifest-repo-export-service/pkg/repository/repository.go
sven-urbanski-freiheit-com: Why would the manifest-service need to read from the manifest-repo? What is the use-case?

miguel-crespo-fdc: Ah, wrong service, my bad

miguel-crespo-fdc (approved):

PR #1683, opened 2024-06-20
chore(manifest-export-service): expose a metric for esl processing delays

AminSlk:
manifest-repo-export-service creates a metric that outputs the delay between "last processed event" and "last event that should be processed". Its measured in seconds and is exposed on datadog

### DIFF
--- a/services/manifest-repo-export-service/pkg/cmd/server.go
+++ b/services/manifest-repo-export-service/pkg/cmd/server.go
@@ -26,8 +26,8 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/freiheit-com/kuberpult/pkg/db"
 	"github.com/DataDog/datadog-go/v5/statsd"
+	"github.com/freiheit-com/kuberpult/pkg/db"
 	"github.com/freiheit-com/kuberpult/pkg/logger"
 	cutoff "github.com/freiheit-com/kuberpult/services/manifest-repo-export-service/pkg/db"
 	"github.com/freiheit-com/kuberpult/services/manifest-repo-export-service/pkg/repository"
@@ -161,7 +161,7 @@ func Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	var ddMetrics statsd.ClientInterface;
+	var ddMetrics statsd.ClientInterface
 	if enableMetrics {
 		ddMetrics, err = statsd.New(DatatDogStatsAddr, statsd.WithNamespace("Kuberpult"))
 		if err != nil {
@@ -215,6 +215,15 @@ func Run(ctx context.Context) error {
 			if err != nil {
 				return fmt.Errorf("error in readEslEvent %v", err)
 			}
+			if ddMetrics != nil {
+				processDelay, err := calculateProcessDelay(ctx, esl, time.Now())
+				if err != nil {
+					log.Warn("error in calculateProcessDelay %v", err)
+				}
+				if err := ddMetrics.Gauge("process_delay_seconds", processDelay, []string{}, 1); err != nil {
+					log.Warn("error in ddMetrics.Gauge %v", err)
+				}
+			}
 			if esl == nil {
 				log.Warn("event processing skipped: no esl event found")
 				eslTableEmpty = true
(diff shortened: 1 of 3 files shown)


services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -101,6 +102,15 @@ func Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	enableMetricsString, err := readEnvVar("KUBERPULT_ENABLE_METRICS")
+	if err != nil {
+		enableMetricsString = "false"
```
sven-urbanski-freiheit-com, line 107: please return an error (and exit) if an env var is not set.
Reason:
We have 2 places with default values: the helm chart (values.yaml) and the code here.
There should only be one place.

services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -101,6 +102,15 @@ func Run(ctx context.Context) error {
...
+	}
+	enableMetrics := enableMetricsString == "true"
+	DatatDogStatsAddr, err := readEnvVar("KUBERPULT_DOGSTATSD_ADDR")
+	if err != nil {
+		DatatDogStatsAddr = "127.0.0.1:8125"
```
sven-urbanski-freiheit-com, line 112: same here, do not use defaults

services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -197,6 +215,15 @@ func Run(ctx context.Context) error {
 			if err != nil {
 				return fmt.Errorf("error in readEslEvent %v", err)
 			}
+			if ddMetrics != nil {
+				processDelay, err := calculateProcessDelay(ctx, esl, time.Now())
+				if err != nil {
+					log.Warn("error in calculateProcessDelay %v", err)
```
sven-urbanski-freiheit-com, line 221:
```suggestion
					log.Error("error in calculateProcessDelay, cannot calculate process delay metric %v", err)
```

services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -151,6 +161,14 @@ func Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	var ddMetrics statsd.ClientInterface
+	if enableMetrics {
+		ddMetrics, err = statsd.New(DatatDogStatsAddr, statsd.WithNamespace("Kuberpult"))
+		if err != nil {
+			logger.FromContext(ctx).Fatal("datadog.metrics.error", zap.Error(err))
+		}
+		ctx = context.WithValue(ctx, repository.DdMetricsKey, ddMetrics)
```
sven-urbanski-freiheit-com, line 170: why do we need to change the context here?
I think this is not required
```suggestion
```

services/manifest-repo-export-service/pkg/repository/repository.go
```diff
@@ -54,6 +54,10 @@ import (
 	git "github.com/libgit2/git2go/v34"
 )
 
+type contextKey string
+
+const DdMetricsKey contextKey = "ddMetrics"
```
sven-urbanski-freiheit-com, line 59: this seems unnecessary

sven-urbanski-freiheit-com (changes requested):
see other comments

services/manifest-repo-export-service/pkg/cmd/server.go
AminSlk: I thought maybe in the future we may need the ddmetrics somewhere else for example in repository.go, exactly like what happens in cd-service.

sven-urbanski-freiheit-com: That may or may not happen. But we don't want to write code that is unused right now, so please remove it.

AminSlk: Done!

services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -216,7 +242,9 @@ func Run(ctx context.Context) error {
 			if err != nil {
 				return fmt.Errorf("error in DBWriteCutoff %v", err)
 			}
-
+			if err != nil {
```
sven-urbanski-freiheit-com, line 245: this code will never be reached

sven-urbanski-freiheit-com (approved):

PR #1687, opened 2024-06-21
feat(dex): Allow multiple roles for one user

gsandok:
- Currently we are treating the role of a user as one value but in the future user's might be a part of multiple roles. We need to add both of these roles to the context and then check the permissions for both of the roles


miguel-crespo-fdc:
Overall implementation looks good. Could you add a couple unit tests where we actually have multiple roles?

miguel-crespo-fdc (approved):
Overall implementation looks good. Could you add a couple unit tests where we actually have multiple roles?

PR #1691, opened 2024-06-24
fix: added information to create release dir on export service

miguel-crespo-fdc:


services/cd-service/pkg/repository/transformer.go
```diff
@@ -431,6 +431,7 @@ func (c *CreateApplicationVersion) Transform(
 	transaction *sql.Tx,
 ) (string, error) {
 	version, err := c.calculateVersion(ctx, transaction, state)
+	fmt.Println(version)
```
khaled-ismaeel, line 434: Forgotten debug print? If not, please wrap it in a descriptive message and use a proper logger.

khaled-ismaeel (approved):

PR #1701, opened 2024-06-25
fix(db): use UTC times in database

AminSlk:
Use Utc time in all records in database. Also, log warning when processDelay metric is negative.

### DIFF
--- a/services/manifest-repo-export-service/pkg/cmd/server.go
+++ b/services/manifest-repo-export-service/pkg/cmd/server.go
@@ -219,6 +219,9 @@ func Run(ctx context.Context) error {
 				if err != nil {
 					log.Error("Error in calculateProcessDelay %v", err)
 				}
+				if processDelay < 0 {
+					log.Warn("process delay is negative: esl-time: %v, now: %v, delay: %v", esl.Created, time.Now().UTC(), processDelay)
+				}
 				if err := ddMetrics.Gauge("process_delay_seconds", processDelay, []string{}, 1); err != nil {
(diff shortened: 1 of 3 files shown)


services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -215,10 +215,13 @@ func Run(ctx context.Context) error {
 				return fmt.Errorf("error in readEslEvent %v", err)
 			}
 			if ddMetrics != nil {
-				processDelay, err := calculateProcessDelay(ctx, esl, time.Now())
+				processDelay, err := calculateProcessDelay(ctx, esl, time.Now().UTC())
 				if err != nil {
 					log.Error("Error in calculateProcessDelay %v", err)
 				}
+				if processDelay < 0 {
+					log.Warn("process delay is negative: esl-time: %v, now: %v, delay: %v", esl.Created, time.Now().UTC(), processDelay)
```
sven-urbanski-freiheit-com, line 223:
```suggestion
                                now :=time.Now().UTC()
				processDelay, err := calculateProcessDelay(ctx, esl, now)
				if err != nil {
					log.Error("Error in calculateProcessDelay %v", err)
				}
				if processDelay < 0 {
					log.Warn("process delay is negative: esl-time: %v, now: %v, delay: %v", esl.Created, now, processDelay)
```

You need to use one variable in both functions, otherwise the numbers don't match.

sven-urbanski-freiheit-com (changes requested):
.

sven-urbanski-freiheit-com (approved):

PR #1710, opened 2024-06-28
feat(ESL): Esl for DeleteEnvFromApp to Database

AminSlk:
This Esl is created in the database. Also, in each release in the database the environment will be deleted from the manifest(actually, a new release with new eslID will be created).
Also, The manifest-export-service will read this esl, and delete environment directory in the git repo.

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -560,6 +560,7 @@ func (h *DBHandler) processReleaseRows(ctx context.Context, err error, rows *sql
 	var result []*DBReleaseWithMetaData
 	var maxEslId EslId = -1
 	for rows.Next() {
+		//exhaustruct:ignore
 		var row = &DBReleaseWithMetaData{}
 		var metadataStr string
 		var manifestStr string
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -483,7 +483,7 @@ func (h *DBHandler) DBSelectReleasesByApp(ctx context.Context, tx *sql.Tx, app s
 		"SELECT eslVersion, created, appName, metadata, manifests, releaseVersion, deleted " +
 			" FROM releases " +
 			" WHERE appName=?" +
-			" ORDER BY created DESC;"))
+			" ORDER BY eslVersion, created DESC;"))
 	rows, err := tx.QueryContext(
 		ctx,
 		selectQuery,
@@ -558,6 +558,7 @@ func (h *DBHandler) processReleaseRows(ctx context.Context, err error, rows *sql
 	}(rows)
 	//exhaustruct:ignore
 	var result []*DBReleaseWithMetaData
+	var maxEslId EslId = -1
 	for rows.Next() {
 		var row = &DBReleaseWithMetaData{}
 		var metadataStr string
@@ -568,6 +569,9 @@ func (h *DBHandler) processReleaseRows(ctx context.Context, err error, rows *sql
 				return nil, nil
 			}
 			return nil, fmt.Errorf("Error scanning releases row from DB. Error: %w\n", err)
+		}
+		if row.EslId > maxEslId {
+			maxEslId = row.EslId
 		}
 		// handle meta data
 		var metaData = DBReleaseMetaData{
@@ -592,6 +596,12 @@ func (h *DBHandler) processReleaseRows(ctx context.Context, err error, rows *sql
 		}
 		row.Manifests = manifestData
 		result = append(result, row)
+	}
+	for i := 0; i < len(result); i++ {
+		if result[i].EslId != maxEslId {
+			result = append(result[:i], result[i+1:]...)
+			i--
+		}
 	}
 	err = closeRows(rows)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/db/db.go
```diff
@@ -444,7 +444,14 @@ func (h *DBHandler) DBSelectAnyRelease(ctx context.Context, tx *sql.Tx) (*DBRele
 		ctx,
 		selectQuery,
 	)
-	return h.processReleaseRow(ctx, err, rows)
+	processedRows, err := h.processReleaseRows(ctx, err, rows)
+	if err != nil {
+		return nil, err
+	}
```
miguel-crespo-fdc, line 450: Why are we using the function to process multiple rows? Your query has a 'LIMIT 1 ', so the number of rows should either be 0 (processedRows = nil) or 1

pkg/db/db.go
```diff
@@ -460,7 +467,30 @@ func (h *DBHandler) DBSelectReleaseByVersion(ctx context.Context, tx *sql.Tx, ap
...
-	return h.processReleaseRow(ctx, err, rows)
+
+	processedRows, err := h.processReleaseRows(ctx, err, rows)
+	if err != nil {
+		return nil, err
+	}
+	if len(processedRows) == 0 {
+		return nil, nil
+	}
+	return processedRows[0], nil
```
miguel-crespo-fdc, line 478: Same as above

services/cd-service/pkg/repository/transformer.go
```diff
@@ -1343,36 +1343,66 @@ func (u *DeleteEnvFromApp) Transform(
...
-		return "", fmt.Errorf(thisSprintf("Need to provide the application"))
-	}
+		for _, dbReleaseWithMetadata := range releases {
+			newManifests := make(map[string]string)
+			for envName, manifest := range dbReleaseWithMetadata.Manifests.Manifests {
+				if envName != u.Environment {
+					newManifests[envName] = manifest
+				}
+			}
+			newRelese := db.DBReleaseWithMetaData{
```
miguel-crespo-fdc, line 1359:
```suggestion
			newRelease := db.DBReleaseWithMetaData{
```

services/cd-service/pkg/repository/transformer.go
```diff
@@ -1343,36 +1343,66 @@ func (u *DeleteEnvFromApp) Transform(
 		return "", err
 	}
 
-	fs := state.Filesystem
-	thisSprintf := func(format string, a ...any) string {
-		return fmt.Sprintf("DeleteEnvFromApp app '%s' on env '%s': %s", u.Application, u.Environment, fmt.Sprintf(format, a...))
-	}
+	if state.DBHandler.ShouldUseOtherTables() {
+		releases, err := state.DBHandler.DBSelectReleasesByApp(ctx, transaction, u.Application)
+		if err != nil {
```
miguel-crespo-fdc, line 1348: How do we filter by deleted releases? We shouldn't remove envs from deleted releases

miguel-crespo-fdc: We keep track of what releases are active by application on an all_releases table

miguel-crespo-fdc (changes requested):

services/cd-service/pkg/repository/transformer.go
AminSlk: Fixed

services/cd-service/pkg/repository/transformer.go
AminSlk: Fixed

pkg/db/db.go
AminSlk: Further in the code, I've introduced a new function `DBSelectReleasesByApp` , which should return a list of `DBReleaseWithMetaData` So I needed a function to process multiple rows.
Still we can have a function to process one row, but that would create a lot of repetitive code.

pkg/db/db.go
```diff
@@ -561,15 +597,19 @@ func (h *DBHandler) processReleaseRow(ctx context.Context, err error, rows *sql.
...
 		}
 		row.Manifests = manifestData
-
-	} else {
-		row = nil
+		result = append(result, row)
+	}
+	for i := 0; i < len(result); i++ {
+		if result[i].EslId != maxEslId {
+			result = append(result[:i], result[i+1:]...)
```
miguel-crespo-fdc, line 604: This would have to mean that all of the most recent changes to all releases from some application are in sync regarding the eslID. For example, the query returns:

_row 1: eslID = 1, appName = app, releaseVersion = 1
row 2: eslID = 2, appName = app, releaseVersion = 2
row 3: eslID = 1, appName = app, releaseVersion = 3

Following this implementation, we would only return row2, while we would want to return all 3 rows, as they are the most recent changes to each version, right? 
I think this scenario is possible

miguel-crespo-fdc: And there might be 100 releases that need to be changed, not just the latest one added to the table. You need to change every single active release that is stored in the all_releases table

miguel-crespo-fdc (approved):

PR #1711, opened 2024-06-28
feat: read DexEnabled variable frontend service

bernardo-bastos:
If the user is not logged in, it needs to be redirected to the `/login` page. 
This is only true if Dex is enabled, thus, we need to check this condition on the frontend service.

### DIFF
--- a/services/frontend-service/src/ui/App/index.tsx
+++ b/services/frontend-service/src/ui/App/index.tsx
@@ -74,8 +74,7 @@ export const App: React.FC = () => {
                 (result: GetFrontendConfigResponse) => {
                     UpdateFrontendConfig.set({ configs: result, configReady: true });
                     if (result.authConfig?.dexAuth?.enabled) {
-                        // eslint-disable-next-line no-console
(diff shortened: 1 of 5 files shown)


services/frontend-service/src/ui/App/index.tsx
```diff
@@ -73,6 +73,10 @@ export const App: React.FC = () => {
             .then(
                 (result: GetFrontendConfigResponse) => {
                     UpdateFrontendConfig.set({ configs: result, configReady: true });
+                    if (result.authConfig?.dexAuth?.enabled) {
+                        // eslint-disable-next-line no-console
+                        console.log('Dex is enabled lets redirect ');
```
miguel-crespo-fdc, line 78: Won't this flood console.log?

bernardo-bastos: Yes, the next step is to remove this log and actually make the redirect. I can remove it

miguel-crespo-fdc (approved):

PR #1713, opened 2024-07-01
feat(charts): Add an option to enable cloudSqlProxy in helm charts

AminSlk:
Refactored helm charts.
All options under cd.db. are moved to db. in helm charts.
The dbOption cloudsql is renamed to "postgreSQL".
The sqlite db option is removed.
A helm parameter is added cloudSqlProxyEnabled: true that just adds the proxy

BREAKING CHANGE: helm chart values related to db change. dbOption cloudsql should be renamed to postgreSQL. Options under cd.db should move to db.

### DIFF
--- a/charts/kuberpult/templates/manifest-repo-export-service.yaml
+++ b/charts/kuberpult/templates/manifest-repo-export-service.yaml
@@ -62,7 +62,7 @@ spec:
       serviceAccountName: {{ .Values.db.k8sName }}
 {{- end }}
       containers:
-{{- if eq .Values.db.cloudSqlProxyEnabled }}
+{{- if .Values.db.cloudSqlProxyEnabled }}
       - name: cloud-sql-proxy
         # It is recommended to use the latest version of the Cloud SQL Auth Proxy
         # Make sure to update on a regular schedule!
--- a/charts/kuberpult/templates/manifest-repo-export-service.yaml
+++ b/charts/kuberpult/templates/manifest-repo-export-service.yaml
@@ -14,14 +14,14 @@
 
 # Copyright freiheit.com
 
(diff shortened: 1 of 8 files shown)


charts/kuberpult/values.yaml
```diff
@@ -132,6 +108,32 @@ cd:
...
+  authProxyPort: 5432
+# As the Database feature is still work in progress, for now we recommend to set:
+# dbOption: "NO_DB"
+# Other valid values are "postgreSQL".
+# If you do want to try out the database, for now use
+# writeEslTableOnly: true
+  dbOption: "NO_DB"
+# k8sName is required if `dbOption = "postgreSQL"`, otherwise it's ignored.
+# k8sName is the name of the kubernetes service account.
+  k8sName: "k8sname"
```
sven-urbanski-freiheit-com, line 123: please rename this to `k8sServiceAccountName`

charts/kuberpult/templates/manifest-repo-export-service.yaml
```diff
@@ -58,11 +58,11 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-{{- if eq .Values.cd.db.dbOption "cloudsql" }}
-      serviceAccountName: {{ .Values.cd.db.k8sName }}
+{{- if eq .Values.db.dbOption "postgreSQL" }}
```
sven-urbanski-freiheit-com, line 61: same as above: We should only set this when the k8sname is supplied.

sven-urbanski-freiheit-com (changes requested):
.

sven-urbanski-freiheit-com (approved):

PR #1715, opened 2024-07-01
feat: migration of replaced-by event to database and export service

miguel-crespo-fdc:
Migration of replaced-by event to database and export service.
Fixed bug where events were being fetched from both the database and the manifest


services/cd-service/pkg/service/git.go
```diff
@@ -226,28 +226,6 @@ func (s *GitServer) GetEvents(ctx context.Context, transaction *sql.Tx, fs billy
...
-	allEventsPath := fs.Join(commitPath, "events")
-	potentialEventDirs, err := fs.ReadDir(allEventsPath)
-	if err != nil {
-		return nil, fmt.Errorf("could not read events directory '%s': %v", allEventsPath, err)
-	}
-	for i := range potentialEventDirs {
-		oneEventDir := potentialEventDirs[i]
-		if oneEventDir.IsDir() {
-			fileName := oneEventDir.Name()
-			rawUUID, err := timeuuid.ParseUUID(fileName)
```
miguel-crespo-fdc, line 238: getEvents was fetching both from the manifest and the database. We now only fetch from the database

services/cd-service/pkg/repository/transformer_db_test.go
```diff
@@ -1289,6 +1289,49 @@ func TestEvents(t *testing.T) {
 				},
 			},
 		},
+		{
+			Name: "Replaced By test",
+			// no need to bother with environments here
```
sven-urbanski-freiheit-com, line 1294: not sure what this comments means. Is the `CreateEnvironment` transformer still needed here?

sven-urbanski-freiheit-com (approved):

services/cd-service/pkg/repository/transformer_db_test.go
miguel-crespo-fdc: We do need to bother, so that the CreateApplicationVersion actually deploys it. Copy-paste mistake from previous tests

PR #1724, opened 2024-07-04
feat(export-service): Configure garbage collection frequency and network timeout

ahmed-nour-fdc:


services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -141,7 +141,26 @@ func Run(ctx context.Context) error {
 			return fmt.Errorf("error converting KUBERPULT_ESL_PROCESSING_BACKOFF, error: %w", err)
 		}
 	}
-
+	var networkTimeoutSeconds uint64
+	if val, exists := os.LookupEnv("KUBERPULT_NETWORK_TIMEOUT_SECONDS"); !exists {
+		log.Infof("environment variable KUBERPULT_NETWORK_TIMEOUT_SECONDS is not set, using default timeout of 120 seconds")
+		networkTimeoutSeconds = 120
```
sven-urbanski-freiheit-com, line 147: We currently have 2 default values, 1 in the helm chart, one here.
We should only have 1 in the helm chart.
```suggestion
		return fmt.errorf...
```

services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -141,7 +141,26 @@ func Run(ctx context.Context) error {
...
+	} else {
+		networkTimeoutSeconds, err = strconv.ParseUint(val, 10, 64)
+		if err != nil {
+			return fmt.Errorf("error converting KUBERPULT_NETWORK_TIMEOUT_SECONDS, error: %w", err)
+		}
+	}
+	var gcFrequency uint64
+	if val, exists := os.LookupEnv("KUBERPULT_GARBAGE_COLLECTION_FREQUENCY"); !exists {
+		log.Infof("environment variable KUBERPULT_GARBAGE_COLLECTION_FREQUENCY is not set, using default value of 20")
+		gcFrequency = 20
```
sven-urbanski-freiheit-com, line 157:
```suggestion
		return fmt.Errorf(...)
```

sven-urbanski-freiheit-com (approved):

services/manifest-repo-export-service/pkg/cmd/server.go
ahmed-nour-fdc: The helm chart is not the only way to start the service. It can also be started locally, either directly or with docker-compose, and I don't think this is an important environment variable that needs to be explicitly set for the service to start.

ahmed-nour-fdc: As agreed, I set the variables to required

sven-urbanski-freiheit-com (approved):

PR #1740, opened 2024-07-05
fix(ui): display limit for events and rephrase text

sven-urbanski-freiheit-com:


AminSlk (approved):
Pipeline is failed. Seems like the dates in Commitinfo test shouldn't have been changed.

PR #1741, opened 2024-07-05
fix: fix foreign keys

miguel-crespo-fdc:


database/migrations/sqlite/1714995407000000_event_sourcing_light.up.sql
```diff
@@ -1,6 +1,6 @@
 CREATE TABLE IF NOT EXISTS event_sourcing_light -- aka ESL
 (
-    eslId INTEGER PRIMARY KEY,
+    eslId INTEGER PRIMARY KEY autoincrement,
     created TIMESTAMP,
```
miguel-crespo-fdc, line 4: This allows us to implement the DBDiscoverCurrentEsldID function for sqlite. SQLITE_SEQUENCE table does not exist if there is no autoincrement index

pkg/db/db.go
```diff
@@ -315,6 +315,58 @@ type EslEventRow struct {
...
+// Next value should be the returned on + 1
+func (h *DBHandler) DBDiscoverCurrentEsldID(ctx context.Context, tx *sql.Tx) (*int, error) {
+	if h == nil {
+		return nil, nil
+	}
+	if tx == nil {
+		return nil, fmt.Errorf("DBDiscoverCurrentEsldID: no transaction provided")
+	}
+	var selectQuery string
+	if h.DriverName == "postgres" {
```
sven-urbanski-freiheit-com, line 328: It's not great that we have to look at such low level as the sequence number.
For now, we'll merge it, as it seems to work fine, but we may need to reconsider this.

PR #1743, opened 2024-07-08
feat(CI): Set up export-service with postgres container

ahmed-nour-fdc:
* Create and run postgres container in integrations tests and docker-compose.
* Run export-service in integration tests.
* Removed the duplicate docker compose files, and the default one now is earthly.
* `kuberpult` make target now maps to the target `kuberpult-earthly`.


database/migrations/postgres/1720455590920137_resize_environments_json.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE environments ALTER COLUMN json TYPE varchar(1000000);
```
AminSlk, line 1: Why should we put a limit on the number of characters? we can use `VARCHAR` without any limits, like similar columns in other tables.
```suggestion
ALTER TABLE environments ALTER COLUMN json TYPE VARCHAR;
```

database/migrations/sqlite/1718277940006100_environments.up.sql
```diff
@@ -3,6 +3,6 @@ CREATE TABLE IF NOT EXISTS environments
     created TIMESTAMP,
     version BIGINT,
     name VARCHAR(255),
-    json VARCHAR(255),
+    json VARCHAR(1000000),
```
AminSlk, line 6: Same here

charts/kuberpult/setup-postgres.sh
```diff
@@ -0,0 +1,74 @@
...
+    echo "logs:"
+    kubectl -n "$ns" logs -l "$label" || echo "could not get logs for $label"
+    echo "describe pod:"
+    kubectl -n "$ns" describe pod -l "$label"
+    echo ...
+  done
+}
+
+earthly ../../infrastructure/docker/postgres+docker
+kind load docker-image postgres:local
```
sven-urbanski-freiheit-com, line 22: don't we need a `docker pull` here too?

database/migrations/postgres/1718277940006100_environments.up.sql
```diff
@@ -1 +0,0 @@
-../sqlite/1718277940006100_environments.up.sql
```
sven-urbanski-freiheit-com, line 1: why was this deleted?

charts/kuberpult/setup-postgres.sh
ahmed-nour-fdc: The image is generated from the above earthly command `earthly ../../infrastructure/docker/postgres+docker`

database/migrations/postgres/1720428454051313_fk_deployments.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE deployments
+    ADD COLUMN transformerEslId INTEGER
```
sven-urbanski-freiheit-com, line 2: there should not be any changes in existing migrations. Please create new migrations instead.

sven-urbanski-freiheit-com (changes requested):
.

database/migrations/postgres/1718277940006100_environments.up.sql
ahmed-nour-fdc: This was a symbolic link to the environments migrations file for sqlite. To solve the json size limit problem, we agreed to create new migrations file for postgres, but edit the migrations file for sqlite directly. Since this file is a symbolic link, it also changes by changing the original file.
The solution to this was to remove the symbolic link, and create a file with the same name and content.

tests/integration-tests/cluster-setup/setup-postgres.sh
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bash
+
+set -eu
+set -o pipefail
+
```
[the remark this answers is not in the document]
ahmed-nour-fdc: Will be taken care of in a follow-up story.

PR #1744, opened 2024-07-09
fix(dex): restore default value for fullNameOverride

sven-urbanski-freiheit-com:
This change restores the original behavior of "fullNameOverride", so that there is no breaking change in the helm chart values.

The field was introduced in #1734 where the default value was changed accidentally.

### DIFF
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -307,7 +307,7 @@ auth:
     # Indicates if dex is to be installed. If you want to use your own Dex instance do not enable this flag.
     installDex: false
     # Overrides the name of the dex resources. Allows creating multiple Dex instances on the same cluster. 
-    fullNameOverride: ""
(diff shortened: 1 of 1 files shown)


bernardo-bastos (approved):
Approved. But doesn’t this here accomplish the same result? https://github.com/freiheit-com/kuberpult/blob/main/services/frontend-service/pkg/config/config.go#L43

sven-urbanski-freiheit-com:
> Approved. But doesn’t this here accomplish the same result? https://github.com/freiheit-com/kuberpult/blob/main/services/frontend-service/pkg/config/config.go#L43

Almost, but the helm values.yaml overrides the defaults there because it is defined here as "": https://github.com/freiheit-com/kuberpult/blob/f44d25cb45e630690085c9bb5d3926a46a41eae2/charts/kuberpult/values.yaml#L310

ahmed-nour-fdc (approved):

PR #1745, opened 2024-07-09
chore: add precommit hook to check for secrets

bjoern-michaelsen-freiheit:


Makefile
```diff
@@ -116,3 +116,16 @@ commitlint:
 	@echo
```
sven-urbanski-freiheit-com, line 116: I renamed the PR to "chore", because it's not really a "feature" for kuberpult operators.

sven-urbanski-freiheit-com (approved):

PR #1746, opened 2024-07-09
fix(db): manifest export must not fail on empty commit hash

sven-urbanski-freiheit-com:


sven-urbanski-freiheit-com:
replaced by #1747

PR #1751, opened 2024-07-10
feat(cd-service): Cache overview response

AminSlk:
A new table called `overview_cache` is created. It has a `json` column which contains the result of the overview service.
Whenever the overview service is called it will check the `overview_cache` table. It gets the latests row's json and sends it as a response. If there isn't any rows in the table, it will create one.
After any change in the database, the `overview_cache` will be updated.


pkg/db/db.go
```diff
@@ -4183,3 +4179,87 @@ func (h *DBHandler) RunCustomMigrationEnvironments(ctx context.Context, getAllEn
...
+
+func (h *DBHandler) ReadLatestOverviewCache(ctx context.Context, transaction *sql.Tx) (*OverviewCacheRow, error) {
+	if h == nil {
+		return nil, fmt.Errorf("readLatestOverviewCache: DBHandler is nil")
+	}
+	if transaction == nil {
+		return nil, fmt.Errorf("readLatestOverviewCache: no transaction provided")
+	}
+
+	span, _ := tracer.StartSpanFromContext(ctx, "readLatestOverviewCache")
```
sven-urbanski-freiheit-com, line 4197:
```suggestion
	span, ctx := tracer.StartSpanFromContext(ctx, "readLatestOverviewCache")
```
the returned context should be used.
Also please move all spans to the beginning of the function.

services/cd-service/pkg/repository/overview_cache.go
```diff
@@ -0,0 +1,420 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+MIT License for more details.
+
+You should have received a copy of the MIT License
+along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>.
+
+Copyright freiheit.com*/
+/*
+This file is part of kuberpult.
```
sven-urbanski-freiheit-com, line 17: remove the duplicate header

services/cd-service/pkg/repository/overview_cache.go
```diff
@@ -0,0 +1,420 @@
...
+		EnvironmentGroups: []*api.EnvironmentGroup{},
+		GitRevision:       rev,
+	}
+	latestOverviewRow, err := state.DBHandler.ReadLatestOverviewCache(ctx, transaction)
+	if err != nil {
+		return err
+	}
+
+	if latestOverviewRow != nil {
+		err = json.Unmarshal([]byte(latestOverviewRow.Blob), result)
```
sven-urbanski-freiheit-com, line 71: `Unmarshal` (and `Marshal`) should be part of the function in db.go for consistency.

services/cd-service/pkg/repository/overview_cache.go
```diff
@@ -0,0 +1,420 @@
...
+			var groupName = mapper.DeriveGroupName(config, envName)
+			var envInGroup = getEnvironmentInGroup(previousOverview.EnvironmentGroups, groupName, envName)
+			//exhaustruct:ignore
+			argocd := &api.EnvironmentConfig_ArgoCD{}
+			if config.ArgoCd != nil {
+				argocd = mapper.TransformArgocd(*config.ArgoCd)
+			}
+			env := api.Environment{
+				DistanceToUpstream: 0,
+				Priority:           api.Priority_PROD,
```
sven-urbanski-freiheit-com, line 170: we is this set to `PROD`?

services/cd-service/pkg/repository/overview_cache.go
```diff
@@ -0,0 +1,420 @@
...
+							}
+						}
+					} // Err != nil means no team name was found so no need to parse team locks
+
+					var version *uint64
+					version, err = state.GetEnvironmentApplicationVersion(ctx, transaction, envName, appName)
+					if err != nil && !errors.Is(err, os.ErrNotExist) {
+						return nil, err
+					} else {
+
```
sven-urbanski-freiheit-com, line 241: nitpic
```suggestion
```

services/cd-service/pkg/repository/overview_cache_test.go
```diff
@@ -0,0 +1,438 @@
...
+				t.Fatal("Output mismatch (-want +got):\n", cmp.Diff(tc.ExpectedResult, actualResult))
+			}
+		})
+	}
+}
+
+func TestUpdateOverviewCacheDB(t *testing.T) {
+	tcs := []struct {
+		Name           string
+		CreateTestData func(context.Context, *sql.Tx, *db.DBHandler) error
```
sven-urbanski-freiheit-com, line 341: having a function like this in a test array indicates that these should be either
A) Separate Tests
B) the data the function produces should instead be stored in the test array directly.
Either way, there should not be function pointers in a test array.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -343,6 +343,11 @@ func (r *transformerRunner) Execute(t Transformer, transaction *sql.Tx) error {
 		r.Stack[idx-1] = append(r.Stack[idx-1], msg)
 	}
 	r.Stack = r.Stack[:idx]
+	if r.State.DBHandler.ShouldUseOtherTables() {
```
sven-urbanski-freiheit-com, line 346: There is already a mechanism to reload the overview, so these lines don't seem necessary. See `r.notify.Notify()` in repository.go

services/cd-service/pkg/service/overview_test.go
```diff
@@ -45,170 +48,14 @@ func (m *mockOverviewService_StreamOverviewServer) Context() context.Context {
...
-
 func TestOverviewService(t *testing.T) {
 	var dev = "dev"
 	tcs := []struct {
-		Name  string
-		Setup []repository.Transformer
-		Test  func(t *testing.T, svc *OverviewServiceServer)
+		Name         string
+		Setup        []repository.Transformer
+		Test         func(t *testing.T, svc *OverviewServiceServer)
```
sven-urbanski-freiheit-com, line 56: same comment about test functions: Please make this its own (top-level) test. There should be no need for functions in a test array.

database/migrations/postgres/1720170031470270_overview_cache.up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS overview_cache
+(
+    eslId SERIAL PRIMARY KEY,
+    timestamp TIMESTAMP,
+    blob VARCHAR
```
sven-urbanski-freiheit-com, line 5: we called this `json` in other tables, so we hsould call it the same here.

pkg/db/db.go
```diff
@@ -4309,3 +4344,443 @@ func (h *DBHandler) RunCustomMigrationEnvironments(ctx context.Context, getAllEn
...
+	}
+
+	insertQuery := h.AdaptQuery(
+		"INSERT INTO overview_cache (timestamp, Json) VALUES (?, ?);",
+	)
+	jsonResponse, err := json.Marshal(overviewResponse)
+	if err != nil {
+		return fmt.Errorf("could not marshal overview json data: %w", err)
+	}
+	span.SetTag("query", insertQuery)
```
sven-urbanski-freiheit-com, line 4436:
```suggestion
	span.SetTag("query", insertQuery)
	jsonResponse, err := json.Marshal(overviewResponse)
	if err != nil {
		return fmt.Errorf("could not marshal overview json data: %w", err)
	}
```
we should set the query tag as soon as we have it.

pkg/db/db.go
```diff
@@ -4309,3 +4344,443 @@ func (h *DBHandler) RunCustomMigrationEnvironments(ctx context.Context, getAllEn
...
+		ManifestRepoUrl:   "",
+	}
+	err := h.WriteOverviewCache(ctx, transaction, emptyOverview)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (h *DBHandler) CheckOverviewIsEmpty(overviewResp *api.GetOverviewResponse) bool {
```
sven-urbanski-freiheit-com, line 4670:
```suggestion
func (h *DBHandler) IsOverviewEmpty(overviewResp *api.GetOverviewResponse) bool {
```

pkg/db/db.go
```diff
@@ -4309,3 +4344,443 @@ func (h *DBHandler) RunCustomMigrationEnvironments(ctx context.Context, getAllEn
...
+	return nil
+}
+
+func (h *DBHandler) UpdateOverviewTeamLock(ctx context.Context, transaction *sql.Tx, teamLock TeamLock) error {
+	latestOverview, err := h.ReadLatestOverviewCache(ctx, transaction)
+	if err != nil {
+		return err
+	}
+	if h.CheckOverviewIsEmpty(latestOverview) {
+		return nil
```
sven-urbanski-freiheit-com, line 4454: is it correct that we return here, even though the overview is empty?
I think we need to recalculate the whole overview in this case.

pkg/db/db.go
```diff
@@ -4309,3 +4344,443 @@ func (h *DBHandler) RunCustomMigrationEnvironments(ctx context.Context, getAllEn
...
+	return nil
+}
+
+func (h *DBHandler) UpdateOverviewEnvironmentLock(ctx context.Context, transaction *sql.Tx, environmentLock EnvironmentLock) error {
+	latestOverview, err := h.ReadLatestOverviewCache(ctx, transaction)
+	if err != nil {
+		return err
+	}
+	if h.CheckOverviewIsEmpty(latestOverview) {
+		return nil
```
sven-urbanski-freiheit-com, line 4491: same here (and in other places)

pkg/db/db.go
```diff
@@ -4309,3 +4344,443 @@ func (h *DBHandler) RunCustomMigrationEnvironments(ctx context.Context, getAllEn
...
+	if overviewResp == nil {
+		return true
+	}
+	if len(overviewResp.Applications) == 0 && len(overviewResp.EnvironmentGroups) == 0 && overviewResp.GitRevision == "" {
+		return true
+	}
+	return false
+}
+
+func getEnvironmentByName(groups []*api.EnvironmentGroup, envNameToReturn string) *api.Environment {
```
sven-urbanski-freiheit-com, line 4680: It's a bit surprising to find these functions here.
Maybe if we just move everything here into db/overview.go it would be a bit more logical.

pkg/db/db_test.go
```diff
@@ -1823,6 +1837,1151 @@ func TestReadReleasesByApp(t *testing.T) {
...
+						return fmt.Errorf("mismatch between errors (-want +got):\n%s", diff)
+					}
+					return nil
+				}
+				latestOverview, err := dbHandler.ReadLatestOverviewCache(ctx, transaction)
+				if err != nil {
+					return err
+				}
+				opts := cmpopts.IgnoreUnexported(api.GetOverviewResponse{}, api.EnvironmentGroup{}, api.Environment{}, api.Application{}, api.Release{}, timestamppb.Timestamp{}, api.EnvironmentConfig{}, api.EnvironmentConfig_Upstream{}, api.Environment_Application{}, api.Environment_Application_DeploymentMetaData{}, api.EnvironmentConfig_ArgoCD{}, api.Lock{}, api.Actor{})
+				if diff := cmp.Diff(tc.ExcpectedOverview, latestOverview, opts); diff != "" {
```
sven-urbanski-freiheit-com, line 2376: maybe extract this into a function like `getOverviewIgnoredTypes`, as it's used at least twice.

pkg/db/db_test.go
```diff
@@ -1823,6 +1837,1151 @@ func TestReadReleasesByApp(t *testing.T) {
...
+
+			if err != nil {
+				t.Fatal(err)
+			}
+		})
+	}
+}
+
+func TestForceOverviewRecalculation(t *testing.T) {
+	t.Run("ForceOverviewRecalculation", func(t *testing.T) {
```
sven-urbanski-freiheit-com, line 2961: even if you don't need the table-driven approach right now, please make it easier for the next person who wants to add another test.
=> Add the test array with one element and iterate over it, like in other tests.

services/cd-service/pkg/service/overview.go
```diff
@@ -83,8 +83,20 @@ func (o *OverviewServiceServer) getOverviewDB(
 	if s.DBHandler.ShouldUseOtherTables() {
 		err := s.DBHandler.WithTransaction(ctx, false, func(ctx context.Context, transaction *sql.Tx) error {
 			var err2 error
+			cached_result, err2 := s.DBHandler.ReadLatestOverviewCache(ctx, transaction)
+			if err2 != nil {
+				return err2
+			}
+			if !s.DBHandler.CheckOverviewIsEmpty(cached_result) {
+				response = cached_result
```
sven-urbanski-freiheit-com, line 91: I think this function would be more easy to follow, if we used `WithTransactionT` here and then `return cached_result`.

sven-urbanski-freiheit-com (approved):
.

PR #1765, opened 2024-07-12
fix(dex): do not format the serviceUrl twice

sven-urbanski-freiheit-com:
The formatting for the dex service url was applied twice, thus returning an invalid url.

### DIFF
--- a/pkg/auth/dex.go
+++ b/pkg/auth/dex.go
@@ -296,7 +296,7 @@ func (a *DexAppClient) oauth2Config(scopes []string) (c *oauth2.Config, err erro
 }
 
 // Verifies if the user is authenticated.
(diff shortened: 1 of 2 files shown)


pkg/auth/dex.go
```diff
@@ -296,7 +296,7 @@ func (a *DexAppClient) oauth2Config(scopes []string) (c *oauth2.Config, err erro
 }
 
 // Verifies if the user is authenticated.
-func VerifyToken(ctx context.Context, r *http.Request, clientID, baseURL, dexFullNameOverride string, useClusterInternalCommunication bool) (jwt.MapClaims, error) {
+func VerifyToken(ctx context.Context, r *http.Request, clientID, baseURL, dexServiceURL string, useClusterInternalCommunication bool) (jwt.MapClaims, error) {
```
ahmed-nour-fdc, line 299: Is the `dexServiceURL` parameter used in this function?

ahmed-nour-fdc (approved):

PR #1771, opened 2024-07-16
feat(cd-service): store Failed Esl Events

AminSlk:
Whenever the manifest-export service fails to process and esl, it will save it in the failed esls table and skip it.
There's a new Endpoint in cd-service/frontend-service which Gives the list of last 20 failed esls.

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -4453,7 +4453,6 @@ func (h *DBHandler) DBWriteFailedEslEvent(ctx context.Context,  tx *sql.Tx, eslE
 	span, _ := tracer.StartSpanFromContext(ctx, "DBWriteEslEventInternal")
 	defer span.Finish()
 
-
 	insertQuery := h.AdaptQuery("INSERT INTO event_sourcing_light_failed (created, event_type , json)  VALUES (?, ?, ?);")
 
 	span.SetTag("query", insertQuery)
@@ -4479,7 +4478,7 @@ func (h *DBHandler) DBReadLastFailedEslEvents(ctx context.Context, tx *sql.Tx, l
 		return nil, fmt.Errorf("DBReadlastFailedEslEvents: no transaction provided")
 	}
 
-	query := h.AdaptQuery("SELECT created, event_type, json FROM event_sourcing_light_failed ORDER BY created DESC LIMIT ?;")
+	query := h.AdaptQuery("SELECT eslId, created, event_type, json FROM event_sourcing_light_failed ORDER BY created DESC LIMIT ?;")
 	span.SetTag("query", query)
 	rows, err := tx.QueryContext(ctx, query, limit)
 	if err != nil {
@@ -4497,7 +4496,7 @@ func (h *DBHandler) DBReadLastFailedEslEvents(ctx context.Context, tx *sql.Tx, l
 
 	for rows.Next() {
 		row := &EslEventRow{}
-		err := rows.Scan(&row.Created, &row.EventType, &row.EventJson)
+		err := rows.Scan(&row.EslId, &row.Created, &row.EventType, &row.EventJson)
 		if err != nil {
 			return nil, fmt.Errorf("could not read failed events from DB. Error: %w\n", err)
 		}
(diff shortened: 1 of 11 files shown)


pkg/db/db.go
```diff
@@ -4309,3 +4309,71 @@ func (h *DBHandler) RunCustomMigrationEnvironments(ctx context.Context, getAllEn
...
+
+func (h *DBHandler) DBWriteFailedEslEvent(ctx context.Context, tx *sql.Tx, eslEvent *EslEventRow) error {
+	if h == nil {
+		return nil
+	}
+	if tx == nil {
+		return fmt.Errorf("DBWriteFailedEslEvent: no transaction provided")
+	}
+	span, _ := tracer.StartSpanFromContext(ctx, "DBWriteEslEventInternal")
+	defer span.Finish()
```
sven-urbanski-freiheit-com, line 4321: nitpick: trace should be the first thing in any function:
```suggestion
	span, _ := tracer.StartSpanFromContext(ctx, "DBWriteEslEventInternal")
	defer span.Finish()
	if h == nil {
		return nil
	}
	if tx == nil {
		return fmt.Errorf("DBWriteFailedEslEvent: no transaction provided")
	}
```

pkg/db/db.go
```diff
@@ -4309,3 +4309,71 @@ func (h *DBHandler) RunCustomMigrationEnvironments(ctx context.Context, getAllEn
...
+			EventType: "",
+			EventJson: "",
+		}
+		err := rows.Scan(&row.EslId, &row.Created, &row.EventType, &row.EventJson)
+		if err != nil {
+			return nil, fmt.Errorf("could not read failed events from DB. Error: %w\n", err)
+		}
+		failedEsls = append(failedEsls, row)
+	}
+
```
sven-urbanski-freiheit-com, line 4377: closeRows is missing

pkg/db/db_test.go
```diff
@@ -1592,6 +1592,137 @@ func TestReadWriteEslEvent(t *testing.T) {
...
+				if err != nil {
+					return err
+				}
+
+				if len(actualEvents) > tc.Limit {
+					t.Fatalf("expected %d events, got %d", tc.Limit, len(actualEvents))
+				}
+
+				for i, actualEvent := range actualEvents {
+					if diff := cmp.Diff(tc.Events[len(tc.Events)-1-i].EslId, actualEvent.EslId); diff != "" {
```
sven-urbanski-freiheit-com, line 1706: why do we need this logic to reverse the array index?
We should just expect the events to be in the correct order.
```suggestion
					if diff := cmp.Diff(tc.Events[i].EslId, actualEvent.EslId); diff != "" {
```

services/cd-service/pkg/service/esl.go
```diff
@@ -0,0 +1,62 @@
...
+			response = &api.GetFailedEslsResponse{
+				FailedEsls: failedEslItems,
+			}
+
+			return nil
+		})
+		if err != nil {
+			return nil, err
+		}
+	}
```
sven-urbanski-freiheit-com, line 60:
```suggestion
	} else {
	    return nil, fmt.Errorf("GetFailedEsl is only implemented for the database") 
```

services/cd-service/pkg/service/esl_test.go
```diff
@@ -0,0 +1,144 @@
...
+		Name             string
+		FailedEsls       []*db.EslEventRow
+		ExpectedResponse *api.GetFailedEslsResponse
+	}{
+		{
+			Name: "One failed Esl",
+			FailedEsls: []*db.EslEventRow{
+				{
+					EslId:     1,
+					EventJson: string(`{"env":"dev","app":"my-app","lockId":"ui-v2-ke1up","message":"test","metadata":{"authorEmail":"testemail@example.com","authorName":"testauthor"}}`),
```
sven-urbanski-freiheit-com, line 44: the string conversion is redundant when using the backticks
```suggestion
					EventJson: `{"env":"dev","app":"my-app","lockId":"ui-v2-ke1up","message":"test","metadata":{"authorEmail":"testemail@example.com","authorName":"testauthor"}}`,
```

sven-urbanski-freiheit-com (approved):

PR #1772, opened 2024-07-16
fix(db): Fix json marshaling and deploy time in overview cache

AminSlk:
- Use protojson for marshaling api objects
- Fix Deploy time in overview cache
- ForceRecalculate should not write anything when overview cache is already empty

### DIFF
--- a/pkg/db/overview.go
+++ b/pkg/db/overview.go
@@ -20,9 +20,11 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"regexp"
+	"time"
+
 	"github.com/freiheit-com/kuberpult/pkg/api/v1"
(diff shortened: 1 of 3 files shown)


pkg/db/overview.go
```diff
@@ -20,9 +20,11 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"regexp"
+	"time"
```
sven-urbanski-freiheit-com, line 24:
```suggestion
	"time"
```
Unused import?

sven-urbanski-freiheit-com: Ah, it's just re-ordered. Nevermind.

sven-urbanski-freiheit-com (approved):

PR #1783, opened 2024-07-19
fix(db): handle retries in the manifest-repo-export-service

sven-urbanski-freiheit-com:
This introduces the "RetryError" which indicates that an error should be retried,
and handles the retries in the export-service.

This also splits up the one big transaction in the export into 2: One to read from the DB and one to write the cutoff table.
This was done to avoid long running transactions.

Ref: SRX-4CJW8O


pkg/errors/retry.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (e *RetryError) IsTransaction() bool {
+	return e.errorType == errorTypeTransaction
+}
+
+func (e *RetryError) IsGitRepo() bool {
+	return e.errorType == errorTypeGitRepo
+}
+
+func RetryGitRepo(originalError error) *RetryError {
```
ahmed-nour-fdc, line 64: The name of this function is misleading. Reading it implies that it retries the action, when in reality it only returns the error type.
It could be renamed to something like GitRepoError

pkg/errors/retry.go
```diff
@@ -0,0 +1,76 @@
...
+	return &RetryError{
+		originalError: originalError,
+		errorType:     errorTypeGitRepo,
+	}
+}
+
+func RetryTransaction(originalError error) *RetryError {
```
ahmed-nour-fdc, line 71: Same as above

pkg/db/transactions.go
```diff
@@ -85,25 +90,37 @@ func WithTransactionMultipleEntriesT[T any](h *DBHandler, ctx context.Context, r
 	return WithTransactionMultipleEntriesRetryT(h, ctx, DefaultNumRetries, readonly, f)
 }
 
-// WithTransactionMultipleEntriesRetryT also supports retries
+// WithTransactionMultipleEntriesRetryT is the same as WithTransaction, but you can also return and array of data and retries.
 func WithTransactionMultipleEntriesRetryT[T any](h *DBHandler, ctx context.Context, maxRetries uint8, readonly bool, f DBFunctionMultipleEntriesT[T]) ([]T, error) {
```
ahmed-nour-fdc, line 94: Make this function private.

sven-urbanski-freiheit-com: done

services/manifest-repo-export-service/pkg/repository/repository.go
```diff
@@ -354,19 +355,39 @@ func DefaultPushActionCallback(pushOptions git.PushOptions, r *repository) PushA
...
 
+	err, _ := apply()
+	if err != nil {
+		return fmt.Errorf("first apply failed, aborting: %v", err)
+	}
+	return nil
+}
+
+func (r *repository) PushRepo(ctx context.Context) error {
+	//if true {
```
ahmed-nour-fdc, line 382: Remove these commented lines if they're not needed.

ahmed-nour-fdc (approved):

services/manifest-repo-export-service/pkg/repository/repository.go
sven-urbanski-freiheit-com: done

PR #1793, opened 2024-07-23
fix(CI): Add migrations integration tests

ahmed-nour-fdc:


database/migrations/postgres/1721378995777505_rename_application_locks.up.sql
```diff
@@ -1,2 +1,9 @@
-ALTER TABLE IF EXISTS application_locks
```
AminSlk, line 1: Doesn't the `IF EXISTS` part here work?

database/migrations/postgres/1720012750850077_rename_commit_events.up.sql
```diff
@@ -1,2 +1,6 @@
-ALTER TABLE IF EXISTS events
-    RENAME TO commit_events;
+DO $$
+BEGIN
+    IF NOT EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = 'commit_events') THEN
+        ALTER TABLE events RENAME TO commit_events;
```
AminSlk, line 4: I think we should still keep the `IF EXISTS` part after the ALTER TABLE

database/migrations/postgres/1721378995777505_rename_application_locks.up.sql
```diff
@@ -1,2 +1,9 @@
...
+DO $$ BEGIN IF NOT EXISTS (
+    SELECT 1
+    FROM information_schema.tables
+    WHERE table_name = 'app_locks'
+) THEN
+ALTER TABLE application_locks
```
AminSlk, line 6: Same as above

tests/integration-tests/migrations_test.go
```diff
@@ -0,0 +1,86 @@
...
+		t.Fatalf("Error pinging database: %v", err)
+	}
+	if err := deleteSchemaMigrationsTable(dbConfig); err != nil {
+		t.Fatalf("Failed to delete schema migrations table: %v", err)
+	}
+	testCases := []struct {
+		name        string
+		postRunFunc func(db.DBConfig) error
+	}{
+		{
```
sven-urbanski-freiheit-com, line 61: these 2 elements don't seem to be independent test cases.
So I would suggest to have just 1 element here, and in the actual test (line ~73), to run `RunDBMigrations` twice.

sven-urbanski-freiheit-com (approved):

database/migrations/postgres/1720012750850077_rename_commit_events.up.sql
ahmed-nour-fdc: Done

PR #1794, opened 2024-07-23
fix(db): retry all transactions by default

sven-urbanski-freiheit-com:
By default, all transactions are now retried 3 times.

This also removes writing the argo apps from the cd-service (if the db is enabled), because the manifest-repo-export service already takes care of that.

Another change is to wrap all errors from the database, and unwrap TransformerErrors accordingly in ProcessBatch, so that we have both detailed error messages (wrap) and correct return codes (unwrap).

Ref: SRX-4CJW8O

### DIFF
--- a/pkg/db/transactions.go
+++ b/pkg/db/transactions.go
@@ -19,6 +19,7 @@ package db
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	"github.com/freiheit-com/kuberpult/pkg/logger"
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
@@ -96,7 +97,7 @@ func WithTransactionMultipleEntriesRetryT[T any](h *DBHandler, ctx context.Conte
 
 	tx, err := h.BeginTransaction(ctx, readonly)
 	if err != nil {
-		return onError(err)
+		return onError(fmt.Errorf("error beginning transaction: %v", err))
 	}
 	defer func(tx *sql.Tx) {
 		_ = tx.Rollback()
@@ -106,12 +107,12 @@ func WithTransactionMultipleEntriesRetryT[T any](h *DBHandler, ctx context.Conte
 
 	result, err := f(ctx, tx)
 	if err != nil {
-		return onError(err)
+		return onError(fmt.Errorf("error within transaction: %v", err))
 	}
 	err = tx.Commit()
 	if err != nil {
 		if maxRetries == 0 {
-			return onError(err)
+			return onError(fmt.Errorf("error committing transaction: %v", err))
 		}
 		logger.FromContext(ctx).Sugar().Warnf("db transaction failed, will retry: %v", err)
 		span.Finish()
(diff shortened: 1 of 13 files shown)


pkg/db/transactions.go
```diff
@@ -19,17 +19,36 @@ package db
...
-	_, err := WithTransactionT(h, ctx, readonly, func(ctx context.Context, transaction *sql.Tx) (*interface{}, error) {
+	_, err := WithTransactionT(h, ctx, DefaultNumRetries, readonly, func(ctx context.Context, transaction *sql.Tx) (*interface{}, error) {
+		err2 := f(ctx, transaction)
+		if err2 != nil {
+			return nil, err2
+		}
+		return nil, nil
+	})
+	if err != nil {
+		return err
```
ahmed-nour-fdc, line 45: shouldn't this be `return nil, err`?

pkg/db/transactions.go
```diff
@@ -62,6 +81,11 @@ func WithTransactionT[T any](h *DBHandler, ctx context.Context, readonly bool, f
 
 // WithTransactionMultipleEntriesT is the same as WithTransaction, but you can also return and array of data, not just the error.
 func WithTransactionMultipleEntriesT[T any](h *DBHandler, ctx context.Context, readonly bool, f DBFunctionMultipleEntriesT[T]) ([]T, error) {
+	return WithTransactionMultipleEntriesRetryT(h, ctx, DefaultNumRetries, readonly, f)
+}
+
+// WithTransactionMultipleEntriesRetryT also supports retries
+func WithTransactionMultipleEntriesRetryT[T any](h *DBHandler, ctx context.Context, maxRetries uint, readonly bool, f DBFunctionMultipleEntriesT[T]) ([]T, error) {
 	span, ctx := tracer.StartSpanFromContext(ctx, "DBTransaction")
 	defer span.Finish()
```
ahmed-nour-fdc, line 90: We should also finish the span before the recursive call

pkg/db/transactions.go
```diff
@@ -62,6 +81,11 @@ func WithTransactionT[T any](h *DBHandler, ctx context.Context, readonly bool, f
 
 // WithTransactionMultipleEntriesT is the same as WithTransaction, but you can also return and array of data, not just the error.
 func WithTransactionMultipleEntriesT[T any](h *DBHandler, ctx context.Context, readonly bool, f DBFunctionMultipleEntriesT[T]) ([]T, error) {
+	return WithTransactionMultipleEntriesRetryT(h, ctx, DefaultNumRetries, readonly, f)
+}
+
+// WithTransactionMultipleEntriesRetryT also supports retries
+func WithTransactionMultipleEntriesRetryT[T any](h *DBHandler, ctx context.Context, maxRetries uint, readonly bool, f DBFunctionMultipleEntriesT[T]) ([]T, error) {
```
ahmed-nour-fdc, line 88: I think we should set the `maxRetries` to `uint8` so that we don't just get stuck here forever if a large retries number is passed.
Or we can can check if `maxRetries` is greater than `DefaultNumRetries` we pass the `DefaultNumRetries`.

sven-urbanski-freiheit-com: done

pkg/db/transactions.go
sven-urbanski-freiheit-com: done

ahmed-nour-fdc (approved):

PR #1799, opened 2024-07-25
feat: Allow team specific RBAC permissions

0lione:


charts/kuberpult/templates/cd-service.yaml
```diff
@@ -372,9 +374,15 @@ spec:
 {{- if .Values.auth.dexAuth.enabled }}
 ---
 apiVersion: v1
+apiVersion: v1
```
sven-urbanski-freiheit-com, line 377: this line looks like it belongs somewhere else

pkg/auth/rbac.go
```diff
@@ -185,6 +191,43 @@ func ValidateRbacPermission(line string) (p Permission, err error) {
 	}, nil
 }
 
+func ValidateTeamRbacPermission(line string, teamPermissions *RBACTeams) (team string, err error) {
+
+	p := strings.Split(line, ",")
+
+	if len(p) != 2 {
+		return "", fmt.Errorf("2 fields are expected but %d were specified", len(p))
```
sven-urbanski-freiheit-com, line 199: please return the whole line in the error, e.g.
```suggestion
		return "", fmt.Errorf("2 fields are expected but %d were specified in line %s", len(p), line)
```

pkg/auth/rbac.go
```diff
@@ -185,6 +191,43 @@ func ValidateRbacPermission(line string) (p Permission, err error) {
...
+	users := strings.Split(p[1], " ")
+
+	if p[0] != "*" && !valid.TeamName(p[0]) {
+		return "", fmt.Errorf("invalid team name %s", p[0])
+	}
+
+	for _, user := range users {
+
+		if !valid.UserEmail(user) {
+			return "", fmt.Errorf("invalid user email %s", user)
```
sven-urbanski-freiheit-com, line 211:
```suggestion
			return "", fmt.Errorf("invalid user email '%s'", user)
```

pkg/auth/rbac.go
```diff
@@ -185,6 +191,43 @@ func ValidateRbacPermission(line string) (p Permission, err error) {
...
+		}
+
+		t, ok := teamPermissions.Permissions[user]
+
+		if !ok {
+			teamPermissions.Permissions[user] = []string{p[0]}
+		} else {
+			if teamPermissions.Permissions[user][len(t)-1] == p[0] {
+				// Ignone
```
sven-urbanski-freiheit-com, line 220:
```suggestion
				// Ignore
```
please elaborate a little bit in the comment, why we ignore this line.

pkg/auth/rbac.go
```diff
@@ -185,6 +191,43 @@ func ValidateRbacPermission(line string) (p Permission, err error) {
 	}, nil
 }
 
+func ValidateTeamRbacPermission(line string, teamPermissions *RBACTeams) (team string, err error) {
```
sven-urbanski-freiheit-com, line 194: The function name surprises me. Given that it starts with "Validate" I would expect that it just validates. But it also changes the `teamPermission` parameter.
I'd prefer if the actuall adding of values to `teamPermission` is done outside this function.

pkg/auth/rbac.go
0lione: Would "Ignore if user is listed more than once in the same line/team" be more adequate?

services/cd-service/team.csv
```diff
@@ -0,0 +1,8 @@
+sre,sven.urbanski@freiheit.com ahmed.nour@freiheit.com
+ds,benedikt.jost@freiheit.com
```
sven-urbanski-freiheit-com, line 2: please use a different team name here
```suggestion
medusa,benedikt.jost@freiheit.com
```

sven-urbanski-freiheit-com (changes requested):
(see other comments)

pkg/auth/rbac.go
```diff
@@ -185,6 +191,31 @@ func ValidateRbacPermission(line string) (p Permission, err error) {
 	}, nil
 }
 
+func ValidateTeamRbacPermission(line string) (team string, users []string, err error) {
+
+	p := strings.Split(line, ",")
```
ahmed-nour-fdc, line 196: Can you rename this `p` variable to something that indicates what this variable holds?

pkg/auth/rbac.go
```diff
@@ -302,6 +409,28 @@ func CheckUserPermissions(rbacConfig RBACConfig, user *User, env, team, envGroup
 	}
 }
 
+func CheckUserTeamPermissions(rbacConfig RBACConfig, user *User, team string, action string) error {
+	if rbacConfig.Team == nil {
+		return errors.New("the desired action can not be performed because Dex is enabled without any RBAC Team permissions")
+
+	}
+	t := rbacConfig.Team.Permissions[user.Email]
```
ahmed-nour-fdc, line 417: Please also use a proper name instead of `t`

pkg/auth/rbac.go
```diff
@@ -240,6 +271,57 @@ func ReadRbacPolicy(dexEnabled bool, DexRbacPolicyPath string) (policy *RBACPoli
 	return policy, nil
 }
 
+func AddUsersToTeam(team string, users []string, teamPermissions *RBACTeams) {
+
+	for _, user := range users {
+		t, ok := teamPermissions.Permissions[user]
```
ahmed-nour-fdc, line 277: Same for the `t` variable

services/cd-service/pkg/repository/transformer.go
```diff
@@ -686,6 +686,12 @@ func (c *CreateApplicationVersion) Transform(
 		if err != nil {
 			return "", GetCreateReleaseGeneralFailure(err)
 		}
+		err = state.checkUserTeamPermissions(ctx, transaction, state, c.Application, auth.PermissionCreateRelease, c.RBACConfig)
```
sven-urbanski-freiheit-com, line 689: is it necessary that `checkUserTeamPermissions` is not part of `checkUserPermissions`?
Or the other way around:
Could you create a wrapper function that does both, so that in the transformers we only need to call one?

sven-urbanski-freiheit-com (approved):

PR #1801, opened 2024-07-25
chore: export service tests

miguel-crespo-fdc:


services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -360,7 +359,7 @@ func (c *DeployApplicationVersion) Transform(
 			envLocks, appLocks, teamLocks map[string]Lock
 			err                           error
 		)
-		envLocks, err = state.GetEnvironmentLocksFromDB(ctx, transaction, c.Environment)
+		envLocks, err = state.GetEnvironmentLocks(c.Environment)
```
miguel-crespo-fdc, line 362: We should not read information that will be used for decision making (i.e. deploying or not based on an env lock). If a user creates deploys an application version and creates a lock in quick succession, meaning that the export service has not yet processed the deploy application transformer, this deployment will be blocked by the export service, in spite of being deployed on the manifest

Fix: we read from the manifest

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -846,7 +845,7 @@ func (c *CreateApplicationVersion) Transform(
 			return "", GetCreateReleaseGeneralFailure(err)
 		}
 	}
-	isLatest, err := isLatestVersion(ctx, transaction, state, c.Application, version)
+	isLatest, err := isLatestVersion(state, c.Application, version)
```
miguel-crespo-fdc, line 848: Same as above

PR #1805, opened 2024-07-26
chore: added lock tests for export service

miguel-crespo-fdc:
Ref: SRX-S2YKE1

### DIFF
--- a/services/manifest-repo-export-service/pkg/repository/transformer.go
+++ b/services/manifest-repo-export-service/pkg/repository/transformer.go
@@ -360,7 +360,7 @@ func (c *DeployApplicationVersion) Transform(
 			envLocks, appLocks, teamLocks map[string]Lock
 			err                           error
 		)
-		envLocks, err = state.GetEnvironmentLocksFromDB(ctx, transaction, c.Environment)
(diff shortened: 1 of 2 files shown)


services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -360,7 +360,7 @@ func (c *DeployApplicationVersion) Transform(
 			envLocks, appLocks, teamLocks map[string]Lock
 			err                           error
 		)
-		envLocks, err = state.GetEnvironmentLocksFromDB(ctx, transaction, c.Environment)
+		envLocks, err = state.GetEnvironmentLocks(c.Environment)
```
miguel-crespo-fdc, line 363: We should not read information that will be used for decision making (i.e. deploying or not based on an env lock). If a user creates deploys an application version and creates a lock in quick succession, meaning that the export service has not yet processed the deploy application transformer, this deployment will be blocked by the export service, in spite of being deployed on the manifest. We should only read from the database information that will be written to the manifest

Fix: we read from the manifest

ahmed-nour-fdc (approved):

PR #1806, opened 2024-07-26
chore(manifest-service): move GetManifests from cd-service to manifest-repo-export-service

AminSlk:
Ref: SRX-BVHNX1

### DIFF
--- a/services/manifest-repo-export-service/pkg/service/version_test.go
+++ b/services/manifest-repo-export-service/pkg/service/version_test.go
@@ -368,40 +368,33 @@ func TestGetManifests(t *testing.T) {
 		}
 	}
 
-	for _, tc := range []*testCase{
-		func() *testCase {
-			release := fixtureRelease(appName, 30)
-
-			return &testCase{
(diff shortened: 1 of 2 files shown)


services/manifest-repo-export-service/pkg/service/version_test.go
```diff
@@ -276,3 +283,281 @@ func TestVersion(t *testing.T) {
...
+					fixtureRelease(appNameOther, 10),
+					fixtureRelease(appNameOther, 20),
+					fixtureRelease(appName, 10),
+					fixtureRelease(appName, 20),
+					release,
+				},
+				req:  fixtureRequest(),
+				want: fixtureReleaseToManifests(release),
+			}
+		}(),
```
sven-urbanski-freiheit-com, line 387: I think this would be a bit easier to read, if you pull up the line `release := fixtureRelease(appName, 30)` above the for loop. Then the array element does not need to be an inline function.

sven-urbanski-freiheit-com (approved):

PR #1811, opened 2024-07-29
feat(cloudrun-service): Deploy services asynchronously

ahmed-nour-fdc:
* Write the application release manifest to `queued_deployments` table.
* Configure a background task to poll the table for new deployments, deploy the service and update the database.


database/migrations/postgres/1722263909995805_deployments_cutoff.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS queued_deployments
+(
+    id SERIAL PRIMARY KEY,
```
sven-urbanski-freiheit-com, line 3: please make everything `NOT NULL`

database/migrations/sqlite/1722263909995805_deployments_cutoff.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS queued_deployments
+(
```
sven-urbanski-freiheit-com, line 2: please make everything `NOT NULL`

pkg/api/v1/api.proto
```diff
@@ -675,7 +675,7 @@ message GetManifestsResponse {
 }
 
 service CloudRunService {
-  rpc Deploy (ServiceDeployRequest) returns (ServiceDeployResponse) {}
+  rpc QueueDeployment (ServiceDeployRequest) returns (ServiceDeployResponse) {}
```
sven-urbanski-freiheit-com, line 678:
```suggestion
  rpc QueueDeployment (QueueDeploymentRequest) returns (QueueDeploymentResponse) {}
```

services/cloudrun-service/pkg/cloudrun/cloudrun.go
```diff
@@ -51,49 +53,64 @@ func Init(ctx context.Context) error {
...
+		return nil, fmt.Errorf("could not write deployment to %s table: %v", dbx.QueuedDeploymentsTable, err)
+	}
+	return &api.ServiceDeployResponse{}, nil
+}
+
+func DeployService(manifest []byte) error {
+	// We already validated the manifest before writing it to the database, so we shouldn't expect any error here
+	svc, err := validateService(manifest)
+	if err != nil {
+		return err
```
sven-urbanski-freiheit-com, line 76:
```suggestion
		return fmt.Errorf("unexpected error: invalid manifest found in db: %w", err)
```

services/cloudrun-service/pkg/cmd/server.go
```diff
@@ -46,12 +97,61 @@ func runServer(ctx context.Context) error {
...
+		time.Sleep(5 * time.Second)
+		queuedDeployments, err := dbx.GetQueuedDeployments(ctx, dbHandler)
+		if err != nil {
+			log.Errorf("failed to read queued deployment events: %v", err)
+			continue
+		}
+		if len(queuedDeployments) == 0 {
+			log.Info("no queued deployments to process")
+			continue
+		}
```
sven-urbanski-freiheit-com, line 139:
```suggestion
		queuedDeployments, err := dbx.GetQueuedDeployments(ctx, dbHandler)
		if err != nil {
			log.Errorf("failed to read queued deployment events: %v", err)
			continue
		}
		if len(queuedDeployments) == 0 {
			log.Info("no queued deployments to process")
        		time.Sleep(5 * time.Second)
			continue
		}
```

sven-urbanski-freiheit-com: We should not sleep when there is work to do

services/cloudrun-service/pkg/cmd/server.go
```diff
@@ -46,12 +97,61 @@ func runServer(ctx context.Context) error {
...
+	}
+}
+
+func processEvent(ctx context.Context, event *dbx.QueuedDeployment, dbHandler *db.DBHandler) error {
+	err := cloudrun.DeployService(event.Manifest)
+	if err != nil {
+		// We don't return because error during deploying the service means that the service was deployed but not ready to serve traffic
+		// which is expected behavior from the cloudrun api
+		logger.FromContext(ctx).Sugar().Warnf("service failed to deploy: %v", err)
```
sven-urbanski-freiheit-com, line 154: we should think about a metric here (not necessarily in this PR though)

sven-urbanski-freiheit-com (approved):

database/migrations/postgres/1722263909995805_deployments_cutoff.up.sql
ahmed-nour-fdc: Done for all except the `processed_at` column which will only be written once the service is deployed.

services/cloudrun-service/pkg/cmd/server.go
ahmed-nour-fdc: I'd rather keep the sleep at the beginning of the loop so that if there is a case where `GetQueuedDeployments` always returns an error the loop waits before making another call to the database. Otherwise, loop would continuously poll the database without any delays

sven-urbanski-freiheit-com (approved):

PR #1815, opened 2024-07-30
feat: configurable ssl mode for database

miguel-crespo-fdc:


pkg/db/db.go
```diff
@@ -113,9 +114,11 @@ func Connect(cfg DBConfig) (*DBHandler, error) {
 
 func GetDBConnection(cfg DBConfig) (*sql.DB, error) {
 	if cfg.DriverName == "postgres" {
-		dbURI := fmt.Sprintf("host=%s user=%s password=%s port=%s database=%s sslmode=disable",
+		dbURI := fmt.Sprintf("host=%s user=%s password=%s port=%s database=%s",
 			cfg.DbHost, cfg.DbUser, cfg.DbPassword, cfg.DbPort, cfg.DbName)
-
+		if !cfg.SSLRequired {
+			dbURI = dbURI + " sslmode=disable"
```
hannesg, line 120: This is sadly not the correct way to use it. `sslmode` has 6 different possible values ( https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS ) and the default is `prefer` which is not requiring ssl.

PR #1824, opened 2024-08-01
chore: optimized migrations for commit events to avoid unnecessary pulling

miguel-crespo-fdc:
REF: SRX-YBJOK9


pkg/db/db.go
```diff
@@ -1003,6 +1003,31 @@ func (h *DBHandler) DBSelectAnyEvent(ctx context.Context, transaction *sql.Tx) (
...
 }
 
+func (h *DBHandler) DBContainsMigrationCommitEvent(ctx context.Context, transaction *sql.Tx) (bool, error) {
+	if h == nil {
+		return false, nil
+	}
+	if transaction == nil {
+		return false, fmt.Errorf("DBSelectAnyEvent: no transaction provided")
+	}
+	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectAnyEvent")
```
sven-urbanski-freiheit-com, line 1013:
```suggestion
	span, ctx := tracer.StartSpanFromContext(ctx, "DBContainsMigrationCommitEvent")
```

pkg/db/db.go
```diff
@@ -2120,19 +2145,25 @@ func (h *DBHandler) RunCustomMigrationsCommitEvents(ctx context.Context, getAllE
 				}
 			}
 		}
+		//Migration event
+		err = h.writeEvent(ctx, transaction, 0, "00000000-0000-0000-0000-000000000001", event.EventTypeDBMigrationEventType, strings.Repeat("0", 40), []byte("{}"))
```
sven-urbanski-freiheit-com, line 2149: why the "1" at the end?
```suggestion
		err = h.writeEvent(ctx, transaction, 0, "00000000-0000-0000-0000-000000000000", event.EventTypeDBMigrationEventType, strings.Repeat("0", 40), []byte("{}"))
```

sven-urbanski-freiheit-com: Please extract the special UUID into a constant, or a function that returns it.

sven-urbanski-freiheit-com (changes requested):
.

PR #1828, opened 2024-08-02
feat(ui): show app locks release page

AminSlk:
Ref: SRX-TCE1GP
Now application locks are shown in releases page.
Also, There was a problem in saving application locks date in overview cache which is fixed.
Screenshot 2024-08-02 at 16 54 25
Screenshot 2024-08-02 at 16 54 58


services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -553,6 +554,47 @@ export const useFilteredApplicationLocks = (appNameParam: string | null): Displa
...
 };
 
+export interface DisplayApplicationLock {
+    lock: DisplayLock;
+    application: Environment_Application;
+    environment: Environment;
+    environmentGroup: EnvironmentGroup;
+}
+
+export const useDisplayApplicationLocks = (appName: string | null): DisplayApplicationLock[] => {
```
miguel-crespo-fdc, line 564: We do something similar here at the component level, which allows the UI to be updated when a new lock is inserted: 
https://github.com/freiheit-com/kuberpult/blob/93a322c816e976de388316bcdf93afebe6e7b2e8/services/frontend-service/src/ui/Pages/Locks/LocksPage.tsx#L63

Please use the same approach

miguel-crespo-fdc (changes requested):

PR #1831, opened 2024-08-05
feat: optimize db requests

miguel-crespo-fdc:
Reduced number of DBSelectDeployments calls done by the findOldApplicationVersions function. Created a new table that, for each app, stores the releaseVersion deployed on each environment, reducing the number of requests to the database by X, where X is the number of environments.
Reduced the number of DBSelectEnvironment calls done by the checkUserPermissions function. Instead of requesting all environment configs, looping through them and finding the other we are looking for, we simply request the one we are looking for, reducing the number of requests to the database  by X fold, where X is the number of environments.

Ref: SRX-XT0P80


database/migrations/postgres/1722591376495565_indexes_deployments_environments.up.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX IF NOT EXISTS deployments_idx  ON deployments (appName, envname);
```
sven-urbanski-freiheit-com, line 1: please fix the typos in the commit message
"looging for, educing"
and "a new table tahat"

database/migrations/postgres/1722702812877934_all_deployments.up.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS all_deployments
+(
+    eslVersion INTEGER, -- internal ID for ESL
```
sven-urbanski-freiheit-com, line 3: please mark all of these as "not nullable"

services/cd-service/pkg/repository/repository.go
```diff
@@ -1982,42 +2018,47 @@ func (s *State) WriteCurrentlyDeployed(ctx context.Context, transaction *sql.Tx,
...
+					DeployedByName:  "",
+					DeployedByEmail: "",
+				},
+			}
+			err = dbHandler.DBWriteDeployment(ctx, transaction, deployment, 0)
+			if err != nil {
+				return fmt.Errorf("error writing Deployment to DB for app %s in env %s: %w", deployment.App, deployment.Env, err)
+			}
+		}
+		err = dbHandler.DBWriteAllDeploymentsForApp(ctx, transaction, 0, appName, deploymentsForApp)
```
sven-urbanski-freiheit-com, line 2059: if this migration already ran, we won't execute it again, right? So we have the deployments, but not "all deployments".
 Is this an issue?

miguel-crespo-fdc: Yup, you are correct. I separated these migrations for now

services/manifest-repo-export-service/pkg/repository/repository.go
```diff
@@ -687,18 +687,18 @@ func (r *repository) afterTransform(ctx context.Context, transaction *sql.Tx, st
...
-	}
-	for env, config := range configs {
-		if config.ArgoCd != nil {
-			err := r.updateArgoCdApps(ctx, transaction, &state, env, config)
-			if err != nil {
-				return err
-			}
-		}
-	}
+	//configs, err := state.GetEnvironmentConfigs()
```
sven-urbanski-freiheit-com, line 690: please remove dead code

sven-urbanski-freiheit-com: but actually, I think this has to stay, otherwise the updateArgoCdApps is never called

sven-urbanski-freiheit-com (approved):

PR #1843, opened 2024-08-06
fix: Define database username/password in a secret

ahmed-nour-fdc:
Ref: SRX-FPNRJ4

### DIFF
--- a/charts/kuberpult/Earthfile
+++ b/charts/kuberpult/Earthfile
@@ -1,7 +1,5 @@
 VERSION 0.8
 
-
-
 godeps:
     FROM golang:1.22-alpine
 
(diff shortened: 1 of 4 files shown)


charts/kuberpult/Earthfile
```diff
@@ -46,7 +44,7 @@ chart.yaml:
     WORKDIR /kp
 
     COPY Chart.yaml.tpl .
-    
+
```
sven-urbanski-freiheit-com, line 47: nitpick: A linter/formatter for earthfiles would be great

ahmed-nour-fdc: There are no available formatters for Earthfiles currently. I configured my editor to treat Earthfiles as Dockerfiles and it formats them as such.

sven-urbanski-freiheit-com (approved):

PR #1845, opened 2024-08-07
fix: measure push failures also in success case

sven-urbanski-freiheit-com:
So that we always send data to datadog


services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -341,6 +339,18 @@ func processEsls(ctx context.Context, repo repository.Repository, dbHandler *db.
 	}
 }
 
+func measurePushFailures(ddMetrics statsd.ClientInterface, log *zap.SugaredLogger, failure bool) {
```
bjoern-michaelsen-freiheit, line 342: `measurePushFailures` is not what this does anymore, remove `Failures`?

bjoern-michaelsen-freiheit (approved):

services/manifest-repo-export-service/pkg/cmd/server.go
sven-urbanski-freiheit-com: done

PR #1849, opened 2024-08-07
feat: Add rule to always check for a history reference in pr's

0lione:
Ref: SRX-VC22X3


commitlint.config.js
0lione, line 1: Had to change the file format as I was not able to recreate the same plugin without using javascript

bjoern-michaelsen-freiheit (approved):

PR #1854, opened 2024-08-08
feat(frontend-service): new /api endpoint for releasetrains

AminSlk:
Ref: SRX-K9MHOE
Added two `/api/environments/releasetrain` and `/api/environment-groups/releasetrain` endpoints for creating a releasetrain.
Added a new `TargetType` field for releasetrain requests. So, the target type is either environment, environmentGroup, or Unknown. For backward compatibility, we use Unknown in previous apis.


services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -1458,13 +1458,13 @@ func getEnvironmentGroupsEnvironmentsOrEnvironment(configs map[string]config.Env
 			envGroupConfigs[env] = config
 		}
 	}
-	if len(envGroupConfigs) == 0 {
+	if isTargetEnvGroup || len(envGroupConfigs) == 0 {
```
sven-urbanski-freiheit-com, line 1461: I'm not sure if this logic is correct.
I would prefer something like this on top level of this function:
if isTargetEnvGroup {
  /// look for group
} else {
 // look for env
}

docs/release-train.md
```diff
@@ -13,7 +13,7 @@ The train should run *often enough* to not slow down development, while also giv
 A release train takes the versions that are currently deployed on one environment and deploys those version to another environment.
```
sven-urbanski-freiheit-com, line 13: from the commit message:
> However, because of backward compatibility this field is not forced
Why not? This is a backwards incompatibility only within grpc requests, so we have control over everyone who calls them.
We should update all places that call them, and be done with it.

services/cd-service/pkg/service/release_train_prognosis.go
```diff
@@ -42,6 +42,7 @@ func (s *ReleaseTrainPrognosisServer) GetReleaseTrainPrognosis(ctx context.Conte
 		WriteCommitData:       false,
 		Repo:                  s.Repository,
 		TransformerEslVersion: 0,
+		TargetType:            api.ReleaseTrainRequest_UNKNOWN.String(),
```
sven-urbanski-freiheit-com, line 45: Either add `TargetType` to the type `api.ReleaseTrainRequest` - or create a follow-up story to handle that.

services/frontend-service/pkg/handler/releasetrain.go
```diff
@@ -101,6 +131,7 @@ func (s Server) handleReleaseTrainPrognosis(w http.ResponseWriter, req *http.Req
 		Target:     target,
 		CommitHash: "",
 		Team:       teamParam,
+		TargetType: api.ReleaseTrainRequest_UNKNOWN,
```
sven-urbanski-freiheit-com, line 134: same here

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
```diff
@@ -181,7 +186,12 @@ export const ProductVersion: React.FC = () => {
                     addAction({
                         action: {
                             $case: 'releaseTrain',
-                            releaseTrain: { target: env, commitHash: selectedTag, team: '' },
+                            releaseTrain: {
+                                target: env,
+                                commitHash: selectedTag,
+                                team: '',
+                                targetType: ReleaseTrainRequest_TargetType.UNKNOWN,
```
sven-urbanski-freiheit-com, line 193: I think we're just one call of `splitCombinedGroupName` away, to know if it's an env or a group.
Either do that, or add it to the follow-up story.

services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
```diff
@@ -195,7 +205,12 @@ export const ProductVersion: React.FC = () => {
                 addAction({
                     action: {
                         $case: 'releaseTrain',
-                        releaseTrain: { target: env, commitHash: selectedTag, team: teams[0] },
+                        releaseTrain: {
+                            target: env,
+                            commitHash: selectedTag,
+                            team: teams[0],
+                            targetType: ReleaseTrainRequest_TargetType.UNKNOWN,
```
sven-urbanski-freiheit-com, line 212: same here

PR #1859, opened 2024-08-09
fix: Render long lock messages nicer

0lione:
Now every column where there is the possibility to have a very long value (Environment, Lock Message, Team and Application)  is limited to a width and height of 15vw and 5vw in case of overflow a scroll will appear. This makes so the rest of the width is redistributed by the other flex box elements.  

BEFORE:
Screenshot from 2024-08-09 11-53-06

AFTER:
Screenshot from 2024-08-12 09-22-51

Ref: SRX-7KZ8HP

### DIFF
--- a/services/frontend-service/src/ui/components/LockDisplay/LockDisplay.scss
+++ b/services/frontend-service/src/ui/components/LockDisplay/LockDisplay.scss
@@ -38,8 +38,8 @@ Copyright freiheit.com*/
 }
 
 .lock-display-info-size-limit {
-    max-width: 300px;
-    max-height: 300px;
+    max-width: 15vw;
+    max-height: 5vw;
     overflow: auto;
     @extend .lock-display-info;
 }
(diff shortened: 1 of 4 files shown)


services/frontend-service/src/ui/components/LockDisplay/LockDisplay.scss
```diff
@@ -20,7 +20,7 @@ Copyright freiheit.com*/
     margin: $locks-list-row-spacing;
     justify-content: center;
     align-items: center;
```
sven-urbanski-freiheit-com, line 22: please update the `after` screenshot

services/frontend-service/src/ui/components/LockDisplay/LockDisplay.scss
```diff
@@ -20,7 +20,7 @@ Copyright freiheit.com*/
...
-    height: $locks-list-row-height;
+    height: fit-content;
 }
 
 .lock-display-table {
```
sven-urbanski-freiheit-com, line 26: please give the commit message ~twice as much horizontal space as the other columns.

sven-urbanski-freiheit-com (approved):

PR #1862, opened 2024-08-09
feat:  cli errors and timeouts

miguel-crespo-fdc:
CLI now returns errors codes based on operation failure/success
CLI now retries failed requests to Kuberpult. User can provide number of attempts.
CLI now has a default timeout for http requests of 10 seconds.

Ref: SRX-CWT5O2


cli/pkg/cmd/parsing_test.go
```diff
@@ -126,6 +130,22 @@ func TestParseArgs(t *testing.T) {
...
+			},
+			expectedOther: []string{"potato", "--tomato"},
+		},
+		{
+			name:    "--retries is negative",
+			cmdArgs: "--url something.somewhere --retries -1 --author_email john  subcommand --arg1 val1 etc etc",
+			expectedError: errMatcher{
+				msg: "error while creating kuberpult client parameters, error: --retries arg value must be positive",
+			},
+		},
```
ahmed-nour-fdc, line 148: We should test overwriting the default value of retries.

cli/pkg/release/http.go
```diff
@@ -21,13 +21,19 @@ import (
...
 	"mime/multipart"
 	"net/http"
 	urllib "net/url"
+	"time"
 
 	kutil "github.com/freiheit-com/kuberpult/cli/pkg/kuberpult_utils"
 )
 
+const (
+	DefaultTimeout = 10
```
ahmed-nour-fdc, line 34: A hard-coded value for the HTTP requests' timeout of 10 seconds seems a bit too small. There could be a case where the manifests are large and might need more than 10 seconds to process, in which case, all three retries will fail because 10 seconds are just not enough.
Can you tests this with a decent size manifest?

miguel-crespo-fdc: I have a trace from the load tests where a call to the /release endpoint takes around 1 minute. So I bumped this to 3 minutes

cli/pkg/cmd/parsing_test.go
```diff
@@ -126,6 +130,31 @@ func TestParseArgs(t *testing.T) {
...
+			name:    "default retries",
+			cmdArgs: "--url something.somewhere potato --tomato",
+			expectedParams: &kuberpultClientParameters{
+				url:     "something.somewhere",
+				retries: DefaultRetries,
+			},
+			expectedOther: []string{"potato", "--tomato"},
+		},
+		{
+			name:    "default retries",
```
ahmed-nour-fdc, line 143:
```suggestion
			name:    "overriding default retries",
```

PR #1864, opened 2024-08-12
feat: release CLI version alongside other kuberpult services 

miguel-crespo-fdc:
Ref: SRX-6SQ4UE


Makefile.variables
```diff
@@ -19,6 +19,6 @@ KUBERPULT_BUILDER?=europe-west3-docker.pkg.dev/fdc-public-docker-registry/kuberp
 
 DOCKER_REGISTRY_URI?=europe-west3-docker.pkg.dev/fdc-public-docker-registry/kuberpult
 SERVICE_IMAGES:=kuberpult-cd-service kuberpult-rollout-service kuberpult-frontend-service kuberpult-manifest-repo-export-service kuberpult-cloudrun-service
-
+CLI_IMAGE:=kuberpult-client
```
bjoern-michaelsen-freiheit, line 22: Hmm, this seems wrong somehow. The execplaner concept has multiple selfcontained subprojects in one repo. Those subprojects should be as independant as possible (each having their own Buildfile.yaml and Makefile). Dumping cli specific stuff into the toplevel Makefile thus seems wrong to me.

But: that seems to maybe be something for a refactor by itself?

bjoern-michaelsen-freiheit (approved):

Issue #1895, opened 2024-08-19
Commit Event Data still being written despite the flag enableWritingCommitData set to false


gsandok:
We have the flag enableWritingCommitData set to false but still get the event `lock-prevented-deployment` written to our manifest repository. This is because the flag is never checked [here](https://github.com/freiheit-com/kuberpult/blob/main/services/cd-service/pkg/repository/transformer.go#L3749).  As far as we can see there is no other event written besides the event `lock-prevented-deployment`

PR #1897, opened 2024-08-20
fix: Check if WriteCommitData is enabled before writing skip causes

ahmed-nour-fdc:
Fixes https://github.com/freiheit-com/kuberpult/issues/1895

Ref: SRX-ORIWUQ

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -3754,22 +3754,23 @@ func (c *envReleaseTrain) Transform(
 			}
 			releaseDir := releasesDirectoryWithVersion(state.Filesystem, appName, release)
 			newEvent := createLockPreventedDeploymentEvent(appName, c.Env, prognosis.FirstLockMessage, "environment")
-
-			if state.DBHandler.ShouldUseOtherTables() {
-				commitID, err := getCommitID(ctx, transaction, state, state.Filesystem, release, releaseDir, appName)
-				if err != nil {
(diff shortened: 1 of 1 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -3754,22 +3754,23 @@ func (c *envReleaseTrain) Transform(
...
-				if err != nil {
-					logger.FromContext(ctx).Sugar().Warnf("could not write event data - continuing. %v", fmt.Errorf("getCommitIDFromReleaseDir %v", err))
-				} else {
-					gen := getGenerator(ctx)
-					eventUuid := gen.Generate()
-					err = state.DBHandler.DBWriteLockPreventedDeploymentEvent(ctx, transaction, c.TransformerEslVersion, eventUuid, commitID, newEvent)
+			if c.WriteCommitData {
+				if state.DBHandler.ShouldUseOtherTables() {
+					commitID, err := getCommitID(ctx, transaction, state, state.Filesystem, release, releaseDir, appName)
 					if err != nil {
```
bjoern-michaelsen-freiheit, line 3760: if we get three levels deep in conditionals, we should consider to refactor out into a separate function (also for enabling unittesting with decent coverage).

bjoern-michaelsen-freiheit (approved):

services/cd-service/pkg/repository/transformer.go
ahmed-nour-fdc: I refactored it so that we don't have three-level deep conditionals.
Now we just return before the parsing logic if `c.WriteCommitData` is not enabled.

PR #1899, opened 2024-08-21
feat: Get commit deployment status for all applications

ahmed-nour-fdc:
The `GetCommitDeploymentInfo` endpoint now accepts only a commit hash, and returns a map of all applications and their deployment status for all environments, as shown:
```
{
  "app1": {
    "dev": DEPLOYED,
    "staging": DEPLOYED,
  },
  "app2": {
    "dev": DEPLOYED,
    "staging": PENDING,
  },
}
```
Ref: SRX-H0ZDQS


services/cd-service/pkg/service/commit_deployment.go
```diff
@@ -95,13 +96,53 @@ func getCommitDeploymentInfoForApp(ctx context.Context, h *db.DBHandler, commit,
 	if err != nil {
 		return nil, fmt.Errorf("Could not get all environments from all_environments metadata: %v", err)
 	}
+
+	for _, app := range apps {
+		commitDeploymentStatusForApp, err := getCommitDeploymentInfoForApp(ctx, s.DBHandler, releaseNumber, app, allEnvironments)
```
bjoern-michaelsen-freiheit, line 101: hmm, instead of spawning a query to the db for each app, maybe we should have a `get CommitDeploymentInfoForApps` that does it for all of them.
SQL would look something like:
```
SELECT ... FROM all_environments GROUP BY app ORDER BY version
```

ahmed-nour-fdc: We would have to query the database for each app to get the release version for each environment, and I don't see a way to do this in one query.

ahmed-nour-fdc: Resolved.
Now all apps and their releases are returned in a single query.

bjoern-michaelsen-freiheit (approved):
nice ;)

PR #1909, opened 2024-08-26
fix: add option for ignoring sync policy in argo request

leandro-freiheit:
We need a unit-test for this case plus the AllowUnexported does not guarantee the exclusion. There is a package called `cmpopts` that gives exactly what we want.

Ref: SRX-87V1VH

### DIFF
--- a/services/rollout-service/pkg/argo/argo.go
+++ b/services/rollout-service/pkg/argo/argo.go
@@ -175,6 +175,7 @@ func (a ArgoAppProcessor) CreateOrUpdateApp(ctx context.Context, overview *api.G
 			}
 
 			//We have to exclude the unexported type destination and the syncPolicy
+			//exhaustruct:ignore
 			diff := cmp.Diff(appUpdateRequest.Application.Spec, existingApp.Spec,
 				cmp.AllowUnexported(v1alpha1.ApplicationDestination{}),
 				cmpopts.IgnoreTypes(v1alpha1.SyncPolicy{}))
--- a/services/rollout-service/pkg/argo/argo.go
+++ b/services/rollout-service/pkg/argo/argo.go
@@ -174,11 +174,9 @@ func (a ArgoAppProcessor) CreateOrUpdateApp(ctx context.Context, overview *api.G
 				Project:              conversion.FromString(appToUpdate.Spec.Project),
 			}
 
-			//exhaustruct:ignore
-			emptyAppSpec := v1alpha1.ApplicationSpec{}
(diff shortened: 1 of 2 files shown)


services/rollout-service/pkg/argo/argo.go
```diff
@@ -174,11 +175,14 @@ func (a ArgoAppProcessor) CreateOrUpdateApp(ctx context.Context, overview *api.G
...
 			//exhaustruct:ignore
-			emptyAppSpec := v1alpha1.ApplicationSpec{}
+			emptyAppSpec := v1alpha1.ApplicationSpec{
+				Destination: v1alpha1.ApplicationDestination{},
+				SyncPolicy:  &v1alpha1.SyncPolicy{},
+			}
 			//We have to exclude the unexported type destination and the syncPolicy
 			diff := cmp.Diff(appUpdateRequest.Application.Spec, existingApp.Spec,
 				cmp.AllowUnexported(emptyAppSpec.Destination),
-				cmp.AllowUnexported(emptyAppSpec.SyncPolicy))
```
hannesg, line 181: How about:

```go
				cmpopts.IgnoreUnexported(v1alpha1.ApplicationDestination{}),
				cmpopts.IgnoreUnexported(v1alpha1.SyncPolicy{}))
```

leandro-freiheit: I found that `IgnoreTypes` is exactly what we want. ``IgnoreUnexported` does not guarantee the diff to be excluded 100% of the times

services/rollout-service/pkg/argo/argo_test.go
```diff
@@ -771,8 +839,15 @@ func (a mockArgoProcessor) CreateOrUpdateApp(ctx context.Context, overview *api.
 			Application:          appToUpdate,
 			Project:              conversion.FromString(appToUpdate.Spec.Project),
 		}
+		//exhaustruct:ignore
+		emptyAppSpec := v1alpha1.ApplicationSpec{}
+
+		//We have to exclude the unexported type destination and the syncPolicy
+		diff := cmp.Diff(appUpdateRequest.Application.Spec, existingApp.Spec,
+			cmp.AllowUnexported(emptyAppSpec.Destination),
```
hannesg, line 847: Use

```
cmp.AllowUnexported(v1alpha1.ApplicationDestination{})
```
and then you can get rid of `emptyAppSpec`.

hannesg (approved):
The change looks okay so far. However, the issue was not caught by unit tests althought it should have. I don't see how the missing unit test coverage is addressed or is it?

PR #1915, opened 2024-08-30
feat(cd-service): Add Regexes for isMinor calculation

AminSlk:
Ref: SRX-QUEYU1
In the helm charts under `cd.minorRegexes` you can specify a list of regexs. Any line in the manifests that match with any of the regexes are ignored during the process of finding out whether the release is minor or not.


services/cd-service/pkg/repository/transformer.go
```diff
@@ -780,21 +782,45 @@ func (c *CreateApplicationVersion) checkMinorFlags(ctx context.Context, transact
...
+		filteredLines1 := filterManifestLines(ctx, manifest1, comparisonRegexes)
+		filteredLines2 := filterManifestLines(ctx, manifest2, comparisonRegexes)
+		if !reflect.DeepEqual(filteredLines1, filteredLines2) {
 			return false
 		}
 	}
 	return true
 }
 
+func filterManifestLines(ctx context.Context, str string, regexes []*regexp.Regexp) []string {
```
ahmed-nour-fdc, line 806: We should write unit tests for this function

ahmed-nour-fdc (approved):

PR #1916, opened 2024-08-30
feat: Implement /api/commit-deployments endpoint

ahmed-nour-fdc:
Ref: SRX-2YR47Y

### DIFF
--- a/services/frontend-service/pkg/handler/commit_deployments.go
+++ b/services/frontend-service/pkg/handler/commit_deployments.go
@@ -1,0 +1,58 @@
+/*This file is part of kuberpult.
+
+Kuberpult is free software: you can redistribute it and/or modify
+it under the terms of the Expat(MIT) License as published by
+the Free Software Foundation.
+
+Kuberpult is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


services/frontend-service/pkg/handler/commit_deployments.go
```diff
@@ -0,0 +1,58 @@
...
+		return
+	}
+	json, err := json.Marshal(resp.DeploymentStatus)
+	if err != nil {
+		logger.FromContext(ctx).Error("failed to get commit deployments from server: failed to marshal response", zap.Error(err))
+		http.Error(w, fmt.Sprintf("failed to marshal response: %v", err), http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write(json)
```
miguel-crespo-fdc, line 56: We probably do not want to ignore these errors

miguel-crespo-fdc (approved):

services/frontend-service/pkg/handler/commit_deployments.go
ahmed-nour-fdc: There is no real way of handling that since this function, or every handle* function we have, don't return errors. So if an error happens while writing the response, then the client wouldn't even be able to know about since we couldn't send anything to it.
This is just following what we already have in all parts of handling rest calls.

PR #1917, opened 2024-09-02
fix: rename commit hash for clarity

sven-urbanski-freiheit-com:
This commit hash references commits in the manifest repo, not in the source repo.


services/cd-service/pkg/service/git.go
```diff
@@ -45,7 +45,7 @@ type GitServer struct {
 	PageSize        uint64
 }
 
-func (s *GitServer) GetGitTags(ctx context.Context, in *api.GetGitTagsRequest) (*api.GetGitTagsResponse, error) {
+func (s *GitServer) GetGitTags(ctx context.Context, _ *api.GetGitTagsRequest) (*api.GetGitTagsResponse, error) {
```
AminSlk, line 48: We have [the same line](https://github.com/freiheit-com/kuberpult/blob/5d333b58ca5155497201649c578ce6ef7e5fc0c2/services/manifest-repo-export-service/pkg/service/git.go#L51) in the manifest export service. We can change that too.

AminSlk (approved):

services/cd-service/pkg/service/git.go
sven-urbanski-freiheit-com: done

PR #1921, opened 2024-09-03
feat: Add  as an option for the self managed apps

leandro-freiheit:
Currently all of the teams need to be specified if we want to have the self managed applications in one project. However, this is not extendable at all. Instead, we should have an all option (marked as `*`).

Ref: SRX-SC9GBF

### DIFF
--- a/services/rollout-service/pkg/argo/argo.go
+++ b/services/rollout-service/pkg/argo/argo.go
@@ -43,15 +43,6 @@ type SimplifiedApplicationServiceClient interface {
 	Watch(ctx context.Context, qry *application.ApplicationQuery, opts ...grpc.CallOption) (application.ApplicationService_WatchClient, error)
 }
 
-type Processor interface {
-	Push(ctx context.Context, last *api.GetOverviewResponse)
-	Consume(ctx context.Context, hlth *setup.HealthReporter) error
-	CreateOrUpdateApp(ctx context.Context, overview *api.GetOverviewResponse, app *api.Environment_Application, env *api.Environment, appsKnownToArgo map[string]*v1alpha1.Application)
-	isSelfManagedFilterActive(team string) (bool, error)
-	ConsumeArgo(ctx context.Context, hlth *setup.HealthReporter) error
-	DeleteArgoApps(ctx context.Context, argoApps map[string]*v1alpha1.Application, apps map[string]*api.Environment_Application)
-}
-
 type ArgoAppProcessor struct {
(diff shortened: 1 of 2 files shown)


services/rollout-service/pkg/argo/argo_test.go
```diff
@@ -782,7 +788,201 @@ func TestArgoConsume(t *testing.T) {
...
+				lastOverview:          tc.Overview,
+				ApplicationClient:     as,
+				trigger:               make(chan *api.GetOverviewResponse, 10),
+				argoApps:              make(chan *v1alpha1.ApplicationWatchEvent, 10),
+				ManageArgoAppsEnabled: tc.ArgoManageEnabled,
+				ManageArgoAppsFilter:  tc.ArgoManageFilter,
+			}
+			hlth.BackOffFactory = func() backoff.BackOff { return backoff.NewConstantBackOff(0) }
+
+			argoProcessor.CreateOrUpdateApp(ctx, tc.Overview, tc.Application, tc.Environment, tc.AppsKnownToArgo)
```
hannesg, line 977: This test case only tests the `mockArgoProcessor`. Do we actually need it?

leandro-freiheit: I wanted to test the filter on the mockArgoProcessor. Hmm, I can see if I can rearrange the function call.

leandro-freiheit: I changed the test to only test the function regarding the selfManage filter (testing both the error output and the isActive flag).

services/rollout-service/pkg/argo/argo.go
```diff
@@ -195,6 +198,16 @@ func (a ArgoAppProcessor) CreateOrUpdateApp(ctx context.Context, overview *api.G
 	}
 }
 
+func (a *ArgoAppProcessor) IsSelfManagedFilterActive(team string) (bool, error) {
+	if len(a.ManageArgoAppsFilter) > 1 && slices.Contains(a.ManageArgoAppsFilter, "*") {
```
hannesg, line 202: This code is not tested by any unit test. Try replacing this method body with `panic("no covered")` and all tests still pass. Please add tests.

leandro-freiheit: I removed the duplicate function in the test file, re-added the interface (so that we can test the function from the go file), added the interface as an argument to the function so that we can test both cases.

hannesg (approved):

PR #1923, opened 2024-09-04
chore(ci): Add integration tests for commit deployments

ahmed-nour-fdc:
Ref: SRX-SF7WYL

### DIFF
--- a/tests/integration-tests/commit_deployments_test.go
+++ b/tests/integration-tests/commit_deployments_test.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"math/rand"
 	"net/http"
 	"strings"
 	"testing"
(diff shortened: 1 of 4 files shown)


tests/integration-tests/commit_deployments_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/freiheit-com/kuberpult/pkg/api/v1"
+)
+
+const (
+	development = "development2"
+	staging     = "staging"
+)
+
+func TestCommitDeployments(t *testing.T) {
+	randNum := fmt.Sprintf("%d", rand.Intn(10000))
```
sven-urbanski-freiheit-com, line 40: Why do we need to randomize the test? 
If this is really necessary, we need to document why.
But I would much prefer to remove any randomness.

ahmed-nour-fdc: The test assumes that the apps do not exist, and builds on that assumption. If the test was to run multiple times, that assumption would be incorrect and the test would fail.

We might not have to the case where we want to run the integration tests more than once, so I will remove this randomization.

sven-urbanski-freiheit-com (approved):

PR #1929, opened 2024-09-05
feat(db): migrate GetVersion to DB

sven-urbanski-freiheit-com:
Ref: SRX-78FWXB

This change `GetVersion` to also work with release numbers instead of commit hashes, in order to be consistent with `GenerateManfest`.

### DIFF
--- a/charts/kuberpult/templates/cd-service.yaml
+++ b/charts/kuberpult/templates/cd-service.yaml
@@ -107,7 +107,8 @@ spec:
             cpu:  {{ .Values.db.requests.cpu }}
 {{- end }}
       - name: service
-        image: "{{ .Values.hub }}/{{ .Values.cd.image }}:{{ $.Chart.AppVersion }}"
+#        image: "{{ .Values.hub }}/{{ .Values.cd.image }}:{{ $.Chart.AppVersion }}"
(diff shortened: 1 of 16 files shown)


charts/kuberpult/templates/cd-service.yaml
```diff
@@ -107,7 +107,8 @@ spec:
             cpu:  {{ .Values.db.requests.cpu }}
 {{- end }}
       - name: service
-        image: "{{ .Values.hub }}/{{ .Values.cd.image }}:{{ $.Chart.AppVersion }}"
+#        image: "{{ .Values.hub }}/{{ .Values.cd.image }}:{{ $.Chart.AppVersion }}"
+        image: "europe-west3-docker.pkg.dev/fdc-public-docker-registry/kuberpult/kuberpult-cd-service:v8.14.1-3-g0b5595b8"
```
hannesg, line 111: Left from testing?

hannesg (commented):
Looks good except for the wrong images

charts/kuberpult/templates/cd-service.yaml
sven-urbanski-freiheit-com: yes, indeed. I removed the image changes now

hannesg (approved):

PR #1932, opened 2024-09-09
feat: add ci links to locks

miguel-crespo-fdc:
Ref: SRX-N70Z31


charts/kuberpult/tests/charts_test.go
```diff
@@ -632,6 +632,24 @@ db:
 			},
 			ExpectedMissing: []core.EnvVar{},
 		},
+		{
```
miguel-crespo-fdc, line 635: Should have been on the previous PR, but I forgot to add it

services/cd-service/pkg/repository/transformer.go
```diff
@@ -2215,13 +2237,21 @@ func (c *CreateEnvironmentApplicationLock) Transform(
 	if err != nil {
 		return "", err
 	}
+	if c.CiLink != "" && state.DBHandler.ShouldUseOtherTables() && !isValidLink(c.CiLink, c.AllowedDomains) {
```
sven-urbanski-freiheit-com, line 2240: what if it's an invalid link but `ShouldUseOtherTables` is false?
We should also handle that case, e.g. by returning an error that says this feature is only available with the db.

PR #1950, opened 2024-09-13
fix: invalid commit ids do not get ported into the database

miguel-crespo-fdc:
Ref: SRX-I88JML

### DIFF
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/freiheit-com/kuberpult/pkg/valid"
 	"io"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


services/cd-service/pkg/repository/repository.go
```diff
@@ -2362,6 +2363,12 @@ func (s *State) WriteAllReleases(ctx context.Context, transaction *sql.Tx, app s
 		if err != nil {
 			return fmt.Errorf("cannot get manifest for app %s and release %v: %v", app, releaseVersion, err)
 		}
+
+		if !valid.SHA1CommitID(repoRelease.SourceCommitId) {
+			logger.FromContext(ctx).Sugar().Warnf("Source commit ID %s is not valid. Skipping migration for release %d of app %s", repoRelease.SourceCommitId, releaseVersion, app)
+			continue
```
sven-urbanski-freiheit-com, line 2369: I don't think we should skip the release, but just set the sourceCommitId to null.

PR #1953, opened 2024-09-16
fix(db): consistent timestamps on transaction

miguel-crespo-fdc:
Every row that a transaction touches, gets created with the same timestamp.

Ref: SRX-I88JML


services/cd-service/pkg/repository/transformer_db_test.go
```diff
@@ -3190,3 +3190,156 @@ func TestTransaction(t *testing.T) {
...
+					Version:         2,
+				},
+				&DeployApplicationVersion{
+					Environment: envAcceptance,
+					Application: "test",
+					Version:     1,
+				},
+				&DeployApplicationVersion{
+					Environment: envAcceptance,
+					Application: "test",
```
AminSlk, line 3256: There are several parts that we used "test" for application name in this testcase. Maybe we can define a constant called applicationName at the top.

AminSlk (approved):

pkg/db/db.go
```diff
@@ -2691,17 +2724,15 @@ func (h *DBHandler) DBWriteEnvironmentLockInternal(ctx context.Context, tx *sql.
 	insertQuery := h.AdaptQuery(
 		"INSERT INTO environment_locks (eslVersion, created, lockID, envName, deleted, metadata) VALUES (?, ?, ?, ?, ?, ?);")
 
-	var timetoInsert time.Time
-	if useTimeInLock {
-		timetoInsert = envLock.Created
-	} else {
-		timetoInsert = time.Now().UTC()
+	now, err := h.DBReadTransactionTimestamp(ctx, tx)
```
sven-urbanski-freiheit-com, line 2727: what about the `usetimeinlock` parameter?

miguel-crespo-fdc: The usetimeinlock parameter now does not make sense. If we want to make the timestamps consistent, we can't have a different reference for locks. The created/timestamp column should always have the timestamp of insertion into the database. If this isn't the case, then locks will not work when we try to backtrack to a previous DB state. However, I would argue we should still store it on the metadata

sven-urbanski-freiheit-com (approved):

PR #1968, opened 2024-09-19
chore: make it easier to see release versions in test data

sven-urbanski-freiheit-com:
The random value in the testdata is not guaranteed to be unique, so we add the release number which is unique
and it is also helpful for the human observer.

Ref: SRX-78FWXB


bjoern-michaelsen-freiheit (approved):

bjoern-michaelsen-freiheit (approved):
> obverserver
typo in PR message ;)

PR #1980, opened 2024-09-24
fix: do not consider the overview cache multiple times in a releasetrain

sven-urbanski-freiheit-com:
This is a performance improvement that solves 2 issues with the overview cache:
1) It reduces the amount of times we write the overview in a release train from `n=number of apps` to `1 per env`. The 1 write contains multiple changed apps.
2) When creating a new app or when undeploying, it does not force a recalculation of the whole overview, but instead only updates the app.

Further cases that also force a complete recalculation of the overview cache (like creating a new environment) are not handled in this PR.

Note that some functions like `CalculateWarnings` had to be moved to another module but were not changed.

Ref: SRX-4WQBET


services/cd-service/pkg/repository/repository.go
```diff
@@ -2330,6 +2336,321 @@ func (s *State) WriteAllCommitEvents(ctx context.Context, transaction *sql.Tx, d
...
+
+	err = h.WriteOverviewCache(ctx, transaction, cache)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *State) UpdateTopLevelApp(ctx context.Context, transaction *sql.Tx, appName string, result *api.GetOverviewResponse, deleteApp bool) error {
```
AminSlk, line 2391: This function name doesn't indicate that it's changing overview and can be confusing. I think `UpdateTopLevelAppInOverview` would be better`.

services/cd-service/pkg/repository/repository.go
```diff
@@ -2330,6 +2336,321 @@ func (s *State) WriteAllCommitEvents(ctx context.Context, transaction *sql.Tx, d
...
+		return api.UndeploySummary_UNDEPLOY
+	}
+	if allNormal {
+		return api.UndeploySummary_NORMAL
+	}
+	return api.UndeploySummary_MIXED
+
+}
+
+func (s *State) UpdateOneAppEnv(ctx context.Context, transaction *sql.Tx, appName string, envName string, configParam *config.EnvironmentConfig) (*api.Environment_Application, error) {
```
AminSlk, line 2543: Same as above. The function name is confusing.

services/cd-service/pkg/repository/repository.go
```diff
@@ -2330,6 +2336,321 @@ func (s *State) WriteAllCommitEvents(ctx context.Context, transaction *sql.Tx, d
...
+	var envConfig = configParam
+	if envConfig == nil {
+		var err error
+		envConfig, err = s.GetEnvironmentConfig(ctx, transaction, envName)
+		if err != nil {
+			return nil, fmt.Errorf("could not get environment to update app %s in env %s: %w", appName, envName, err)
+		}
+	}
+
+	teamName, err := s.GetTeamName(ctx, transaction, appName)
```
AminSlk, line 2553: nit pick. moving this line immediately above the line 2568 would be better. Since checking the error immediately at the next line is better.

services/cd-service/pkg/cmd/server.go
```diff
@@ -344,6 +345,11 @@ func RunServer() {
 					logger.FromContext(ctx).Fatal("Could not pull repository to perform custom migrations", zap.Error(err))
 				}
 				logger.FromContext(ctx).Sugar().Warnf("running custom migrations, because KUBERPULT_DB_WRITE_ESL_TABLE_ONLY=false")
+
+				// we overwrite InsertApp in order to also update the overview:
+				dbHandler.InsertAppFun = func(ctx context.Context, transaction *sql.Tx, appName string, previousEslVersion db.EslVersion, stateChange db.AppStateChange, metaData db.DBAppMetaData) error {
```
hannesg, line 350: This creates a pretty weird control flow as the dbhandler is also used in the repo service. The first instance that does the migration will update the overviews and all later instances won't.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -4023,14 +4053,38 @@ func (c *envReleaseTrain) Transform(
...
+		envGroup := overview.EnvironmentGroups[i]
+
+		for i2 := range envGroup.Environments {
+			env := envGroup.Environments[i2]
+			if env.Name == envName {
+				return env
+			}
+		}
+	}
+	return nil
```
AminSlk, line 4085:
```suggestion
	for _, envGroup := range overview.EnvironmentGroups {
		for _, env := range envGroup.Environments {
			if env.Name == envName {
				return env
			}
		}
	}
	return nil
```

services/cd-service/pkg/cmd/server.go
AminSlk: I agree with the weird control flow. But, I assumed because of the recursive dependencies in go there's no quick workaround for that.

sven-urbanski-freiheit-com: `InsertAppFun` is only called when creating an app or deleting an app.
It will never be called in the migration, the migrations just call `DBInsertApplication` directly without updating the overview.

I agree the flow is not great, but I'm also not sure how to make this simpler.
And we have the same pattern a few lines below, so I think it's okay for now.

services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: done

services/cd-service/pkg/repository/repository.go

services/cd-service/pkg/repository/repository.go

services/cd-service/pkg/repository/repository.go

AminSlk (approved):

PR #1982, opened 2024-09-25
fix(DB): fixing lock dates

miguel-crespo-fdc:
Lock dates were getting written wrong into the overview.

Ref: SRX-FFMDKH


hannesg (approved):
Please add a test for this case

PR #1985, opened 2024-09-26
fix: allow disabling the request queue

sven-urbanski-freiheit-com:
This introduces a new helm option "cd.backendConfig.disableQueue`. When true, the cd-service processes incoming requests in parallel.

Ref: SRX-YNEAHU


miguel-crespo-fdc (approved):

hannesg (commented):
I would remove the info logs.

PR #1987, opened 2024-09-26
fix(db): get all environment configs at the start of a release train

bjoern-michaelsen-freiheit:
in db.go:
- EnvironmentFromRow extract maping from row to db environment object
- remove extra closeRows calls in DBSelectEnvironment, already handled by
   defer
- add a DBSelectEnvironmentsBatch to query for multple environments at once
  * currently hardcoded to bail at more than 1024 environments
  * would need paging for more

in repository.go:
- use DBSelectEnvironmentsBatch instead of looping over queries

in transformer.go:
- fetch environment configs as early as possible
- reuse those configs e.g. in prognosis

Ref: SRX-YNEAHU

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -98,6 +98,7 @@ const (
 const (
 	MigrationCommitEventUUID = "00000000-0000-0000-0000-000000000000"
 	MigrationCommitEventHash = "0000000000000000000000000000000000000000"
+	WhereInBatchMax          = 1024
 )
 
 func (h *DBHandler) ShouldUseEslTable() bool {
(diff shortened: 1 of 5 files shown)


pkg/db/db.go
```diff
@@ -4500,30 +4518,83 @@ LIMIT 1;
...
+    "public"."environments"
+  GROUP BY
+    name) AS latest
+JOIN
+  environments
+ON
+  latest.latest=environments.version
+  AND latest.name = environments.name
+WHERE
+  environments.name IN (?)
```
bjoern-michaelsen-freiheit, line 4556: How do we write a typesafe query with WHERE ... IN for postgresql and sqlite?

sven-urbanski-freiheit-com (approved):

PR #1989, opened 2024-09-27
fix(db): Do not write argo apps if disabled

sven-urbanski-freiheit-com:
While the helm parameter argocd.generateFiles=false technically worked, it would still request a lot of data that is not necessary. With this change, we reduce the number of queries to the database for the case argocd.generateFiles=false.

Ref: SRX-YNEAHU


ahmed-nour-fdc (approved):

jonaswittbrodt:
Could you move the check to before to before the span is opened?

PR #1992, opened 2024-09-27
fix(db): retrieve manifests before expensive step on prognosis

miguel-crespo-fdc:
Ref: SRX-YNEAHU

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -624,7 +624,7 @@ func (h *DBHandler) DBSelectAllManifestsForAllReleases(ctx context.Context, tx *
 		latest.biggestRelease=releases.releaseVersion
 		AND latest.appname=releases.appname;
 `)
-	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectReleaseByVersion")
+	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectAllManifestsForAllReleases")
 	defer span.Finish()
 	span.SetTag("query", selectQuery)
 	rows, err := tx.QueryContext(
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -621,7 +621,7 @@ func (h *DBHandler) DBSelectAllManifestsForAllReleases(ctx context.Context, tx *
 	JOIN
 		releases AS releases
 	ON
-		latest.biggestRelease=releases.releaseVersion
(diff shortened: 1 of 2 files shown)


pkg/db/db.go
```diff
@@ -602,6 +602,84 @@ func (h *DBHandler) DBSelectReleaseByVersion(ctx context.Context, tx *sql.Tx, ap
...
+func (h *DBHandler) DBSelectAllManifestsForAllReleases(ctx context.Context, tx *sql.Tx) (map[string]map[uint64]DBReleaseManifests, error) {
+	selectQuery := h.AdaptQuery(
+		`
+	SELECT DISTINCT
+		releases.appname,
+		releases.releaseVersion,
+		releases.manifests
+	FROM (
+		SELECT
+			MAX(releaseVersion) AS biggestRelease,
```
sven-urbanski-freiheit-com, line 614:
```suggestion
			MAX(releaseVersion) AS latestRelease,
```
(nitpick)

services/cd-service/pkg/repository/transformer.go
```diff
@@ -3769,7 +3781,9 @@ func (c *envReleaseTrain) prognosis(
 		}
 
 		if state.DBHandler.ShouldUseOtherTables() {
-			release, err := state.DBHandler.DBSelectReleaseByVersion(ctx, transaction, appName, versionToDeploy, true)
+			release, exists := allLatestManifests[appName][versionToDeploy]
+			//release, err = state.DBHandler.DBSelectReleaseByVersion(ctx, transaction, appName, versionToDeploy, true)
```
sven-urbanski-freiheit-com, line 3785:
```suggestion
```

services/cd-service/pkg/repository/transformer.go
```diff
@@ -3769,7 +3781,9 @@ func (c *envReleaseTrain) prognosis(
...
+
 			if err != nil {
```
sven-urbanski-freiheit-com, line 3786: this looks like it's never true

sven-urbanski-freiheit-com (approved):

services/cd-service/pkg/repository/transformer.go
miguel-crespo-fdc: Looks like it, but we have no guarantee that it isn't. The old code checks for release == nil, so this is the equivalent to that

PR #2001, opened 2024-10-02
fix(cd-service): update environment applications in overview correctly

AminSlk:
Ref: SRX-09Z4A6

* Update overview cache on create environment
* Update environment applications table in custom migrations

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -1639,6 +1639,10 @@ func (h *DBHandler) RunCustomMigrations(
 	if err != nil {
 		return err // better wrap the error in a descriptive message?
 	}
+	err = h.RunCustomMigrationEnvironmentApplications(ctx)
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/db/db.go
```diff
@@ -5096,6 +5100,36 @@ func (h *DBHandler) needsEnvironmentsMigrations(ctx context.Context, transaction
...
+		if err != nil {
+			return fmt.Errorf("could not get all applications, error: %w", err)
+		}
+		for _, envName := range allEnvironments.Environments {
+			env, err := h.DBSelectEnvironment(ctx, transaction, envName)
+			if err != nil {
+				return fmt.Errorf("could not get env: %s, error: %w", envName, err)
+			}
+			if env.Applications == nil || len(env.Applications) == 0 {
+				err = h.DBWriteEnvironment(ctx, transaction, envName, env.Config, allApplications.Apps)
```
sven-urbanski-freiheit-com, line 5123: is it correct that every environment gets the same list of apps?

sven-urbanski-freiheit-com (approved):

PR #2004, opened 2024-10-02
feat: new endpoint to get app details WIP

sven-urbanski-freiheit-com:
Ref: SRX-RH7CD5

### DIFF
--- a/services/cd-service/pkg/service/overview.go
+++ b/services/cd-service/pkg/service/overview.go
@@ -47,6 +47,84 @@ type OverviewServiceServer struct {
 	Context  context.Context
 	init     sync.Once
 	response atomic.Value
+
+	DBHandler *db.DBHandler
+}
+
+func (o *OverviewServiceServer) GetAppDetails(
+	ctx context.Context,
+	in *api.GetAppDetailsRequest) (*api.GetAppDetailsResponse, error) {
+
+	var appName = in.AppName
+	var result = &api.GetAppDetailsResponse{
+		Application: &api.Application{
+			UndeploySummary: 0,
+			Warnings:        nil,
+			Name:            appName,
+			Releases:        []*api.Release{},
+			SourceRepoUrl:   "",
+			Team:            "",
+		},
+	}
+	resultApp, err := db.WithTransactionT(o.DBHandler, ctx, 2, true, func(ctx context.Context, transaction *sql.Tx) (*api.Application, error) {
+		var result = &api.Application{}
+		var rels []int64
+		retrievedReleasesOfApp, err := o.DBHandler.DBSelectAllReleasesOfApp(ctx, transaction, appName)
+		if err != nil {
+			logger.FromContext(ctx).Sugar().Warnf("app without releases: %v", err)
+		}
+		rels = retrievedReleasesOfApp.Metadata.Releases
+		for _, id := range rels {
+			uid := uint64(id)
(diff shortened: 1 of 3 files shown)


services/frontend-service/pkg/handler/commit_deployments_test.go
```diff
@@ -82,7 +83,7 @@ func TestHandleCommitDeployments(t *testing.T) {
 			inputTail:          "123456/",
 			failGrpcCall:       false,
 			expectedStatusCode: http.StatusOK,
-			expectedResponse:   "{\"deploymentStatus\":{\"app1\":{\"deploymentStatus\":{\"dev\":\"DEPLOYED\",\"prod\":\"UNKNOWN\",\"stage\":\"PENDING\"}}}}\n",
+			expectedResponse:   "{\"deploymentStatus\":{\"app1\":{\"deploymentStatus\":{\"dev\":\"DEPLOYED\", \"prod\":\"UNKNOWN\", \"stage\":\"PENDING\"}}}}\n",
```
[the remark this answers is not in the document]
miguel-crespo-fdc: Only added white spaces

pkg/db/db.go
```diff
@@ -1667,6 +1667,54 @@ func (h *DBHandler) DBSelectLatestDeployment(ctx context.Context, tx *sql.Tx, ap
...
+	SELECT
+		deployments.eslVersion,
+		deployments.created,
+		deployments.appname,
+		deployments.releaseVersion,
+		deployments.envName,
+		deployments.metadata
+	FROM (
+		SELECT
+	MAX(eslVersion) AS latest,
```
sven-urbanski-freiheit-com, line 1685: we just observed this to be a rather slow query.
Which is fine for a new unused endpoint, just a note that we need to test the performance of this before rolling out.
I would suggest to make the `max` into a `order by` with `limit 1` (but we need to test the performance either way)

services/cd-service/pkg/service/overview_test.go
```diff
@@ -659,6 +660,183 @@ func TestOverviewService(t *testing.T) {
...
 }
 
+func TestGetApplicationDetails(t *testing.T) {
+	var dev = "dev"
+	var env = "development"
+	var appName = "test-app"
+	tcs := []struct {
+		Name             string
+		Setup            []repository.Transformer
+		Test             func(t *testing.T, svc *OverviewServiceServer, expectedResponse *api.GetAppDetailsResponse)
```
sven-urbanski-freiheit-com, line 670: this test function should be part of the actual tests, please avoid function pointers in test arrays

services/cd-service/pkg/service/overview.go
```diff
@@ -47,6 +50,177 @@ type OverviewServiceServer struct {
...
+				},
+			}
+		}
+
+		// Team Locks
+		teamLocks, err := o.DBHandler.DBSelectAllActiveTeamLocksForTeam(ctx, transaction, result.Team)
+		if err != nil {
+			return nil, fmt.Errorf("could not find team locks for app %s: %w", appName, err)
+		}
+		fmt.Println(teamLocks)
```
sven-urbanski-freiheit-com, line 170: please remove println

sven-urbanski-freiheit-com (commented):
looks good to me, but since it's my original PR, i can't technically approve it

miguel-crespo-fdc (approved):

PR #2029, opened 2024-10-10
fix: reduce number of queries to get releases

sven-urbanski-freiheit-com:
This reduces the number of queries to the releases table, roughly by a factor of 25 (average number of releases per app).

This also reduces the number of cases where we actually load all the manifests from the releases table.
And it fixes a few tests that had a nil context.

Ref: SRX-LEIHOZ

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -566,7 +566,7 @@ func (h *DBHandler) DBSelectAnyRelease(ctx context.Context, tx *sql.Tx, ignorePr
 		ctx,
 		selectQuery,
 	)
-	processedRows, err := h.processReleaseRows(ctx, err, rows, ignorePrepublishes, true)
+	processedRows, err := h.processReleaseRows(ctx, err, rows, ignorePrepublishes, false)
 	if err != nil {
 		return nil, err
 	}
@@ -822,7 +822,7 @@ func (h *DBHandler) DBSelectReleasesByAppLatestEslVersion(ctx context.Context, t
 		deleted,
 	)
 
-	return h.processReleaseRows(ctx, err, rows, ignorePrepublishes, true)
+	return h.processReleaseRows(ctx, err, rows, ignorePrepublishes, false)
 }
 
 func (h *DBHandler) DBSelectAllReleasesOfApp(ctx context.Context, tx *sql.Tx, app string) (*DBAllReleasesWithMetaData, error) {
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -623,15 +623,6 @@ func (h *DBHandler) DBSelectReleasesByVersions(ctx context.Context, tx *sql.Tx, 
 		return []*DBReleaseWithMetaData{}, nil
 	}
 	repeatedQuestionMarks := strings.Repeat(",?", len(releaseVersions)-1)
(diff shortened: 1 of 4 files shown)


jonaswittbrodt:
I don't think the manifests are actually used in this code path. Can we skip loading them them?

services/cd-service/pkg/repository/repository.go
```diff
@@ -2512,16 +2512,15 @@ func (s *State) UpdateTopLevelAppInOverview(ctx context.Context, transaction *sq
 		}
 		rels = retrievedReleasesOfApp
 	}
-	for _, id := range rels {
-		if rel, err := s.GetApplicationRelease(ctx, transaction, appName, id); err != nil {
-			return err
-		} else {
+
+	if releasesInDb, err := s.GetApplicationReleasesDB(ctx, transaction, appName, rels); err != nil {
```
AminSlk, line 2516: Not sure, but shouldn't it also work for the case when we are not using the db?

pkg/db/db.go
```diff
@@ -615,6 +615,48 @@ func (h *DBHandler) DBSelectReleasesWithoutEnvironments(ctx context.Context, tx
...
+		return []*DBReleaseWithMetaData{}, nil
+	}
+	repeatedQuestionMarks := strings.Repeat(",?", len(releaseVersions)-1)
+	selectQuery := h.AdaptQuery(fmt.Sprintf(
+		`
+SELECT eslVersion, created, appName, metadata, manifests, releaseVersion, deleted, environments
+FROM releases
+WHERE appName=? AND releaseVersion IN (?` + repeatedQuestionMarks + `)
+ORDER BY eslVersion DESC
+LIMIT ?;
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: solved

services/cd-service/pkg/repository/repository.go
sven-urbanski-freiheit-com: this functions is only called from the db package, so this code should never run without db.

pkg/db/db.go
```diff
@@ -942,11 +1014,15 @@ func (h *DBHandler) processReleaseRows(ctx context.Context, err error, rows *sql
...
-			return nil, fmt.Errorf("Error during json unmarshal of manifests for releases. Error: %w. Data: %s\n", err, metadataStr)
+		if withManifests {
+			err = json.Unmarshal(([]byte)(manifestStr), &manifestData)
+			if err != nil {
+				return nil, fmt.Errorf("Error during json unmarshal of manifests for releases. Error: %w. Data: %s\n", err, metadataStr)
+			}
+			row.Manifests = manifestData
+		} else {
+			row.Manifests = manifestData
 		}
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: done

services/cd-service/pkg/repository/repository.go
sven-urbanski-freiheit-com: nevermind, i added it, the test still call this function without the DB enabled

sven-urbanski-freiheit-com:
> I don't think the manifests are actually used in this code path. Can we skip loading them them?

Indeed we can. I found a few more places were we load the manifest for no good reason and disabled loading there.

PR #2030, opened 2024-10-10
fix(db): fix update overview when app does not exist

AminSlk:
Ref: SRX-LEIHOZ

### DIFF
--- a/pkg/db/overview.go
+++ b/pkg/db/overview.go
@@ -113,7 +113,28 @@ func (h *DBHandler) UpdateOverviewDeployment(ctx context.Context, transaction *s
 	}
 	appInEnv := getEnvironmentApplicationByName(env.Applications, deployment.App)
 	if appInEnv == nil {
-		return fmt.Errorf("could not find application %s in environment %s in overview", deployment.App, deployment.Env)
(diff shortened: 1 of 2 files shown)


pkg/db/overview.go
```diff
@@ -113,7 +113,28 @@ func (h *DBHandler) UpdateOverviewDeployment(ctx context.Context, transaction *s
...
+		if err != nil {
+			return fmt.Errorf("could not find application '%s' in apps table, got an error: %w", deployment.App, err)
+		}
+		if selectApp == nil {
+			return fmt.Errorf("could not find application '%s' in apps table: got no result", deployment.App)
+		}
+		env.Applications[deployment.App] = &api.Environment_Application{
+			Version:         0,
+			QueuedVersion:   0,
+			UndeployVersion: false,
```
sven-urbanski-freiheit-com, line 126: can we set the undeployverison from the DeploymentMetaData as well?

sven-urbanski-freiheit-com (approved):

pkg/db/overview.go
AminSlk: It's will be set further down in the code.

PR #2038, opened 2024-10-14
fix(tests): fix TestHealthReporterRetry, TestValidateTokenStatic

AminSlk:
Ref: SRX-AHGZTR
Ref: SRX-TI3JNJ

- TestHealthReporterRetry: There's a background job going on in the ReportHealth function which should be completed before we check the expectations. Sometimes it may be left behind(It's pretty rare). So, if we wait for that and then do the checks, it would solve the problem.
- TestValidateTokenStatic: It seems that we are testing not our code but the code of an underlying library `golang-jwt`. We shouldn't test the error messages that are returned by the library not us.

### DIFF
--- a/pkg/auth/azure_test.go
+++ b/pkg/auth/azure_test.go
@@ -52,11 +52,6 @@ func TestValidateTokenStatic(t *testing.T) {
 		noInit        bool
 	}{
 		{
-			Name:          "Not a token",
-			Token:         "asdf",
-			ExpectedError: errMatcher{"Failed to parse the JWT.\nError: token is malformed: token contains an invalid number of segments"},
-		},
-		{
 			Name:          "Not initialized",
 			Token:         "asdf",
 			noInit:        true,
--- a/pkg/auth/azure_test.go
+++ b/pkg/auth/azure_test.go
@@ -44,6 +44,17 @@ func (e errMatcher) Is(err error) bool {
 	return e.Error() == err.Error()
 }
 
+type prefixErrMatcher struct {
+	prefix string
+}
+
+func (e prefixErrMatcher) Error() string {
+	return e.prefix
+}
+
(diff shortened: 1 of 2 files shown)


pkg/setup/health_test.go
```diff
@@ -341,6 +341,7 @@ func TestHealthReporterRetry(t *testing.T) {
 			for _, st := range tc.Steps {
 				stepCh <- st
 				<-stateChange
+				time.Sleep(500 * time.Millisecond)
```
sven-urbanski-freiheit-com, line 344: in line 336, we send a message to `stateChange` after setting the report to healthy.
Why do we need to sleep here, when all the next check does is test if we area ready?

sven-urbanski-freiheit-com, line 344: in line 336, we send a message to `stateChange` after setting the report to healthy.
Why do we need to sleep here, when all the next check does is test if we area ready?

pkg/auth/azure_test.go
```diff
@@ -51,11 +51,6 @@ func TestValidateTokenStatic(t *testing.T) {
 		ExpectedError error
 		noInit        bool
 	}{
-		{
-			Name:          "Not a token",
-			Token:         "asdf",
-			ExpectedError: errMatcher{"Failed to parse the JWT.\nError: token is malformed: token contains an invalid number of segments"},
```
sven-urbanski-freiheit-com, line 57: I don't quite understand why we need to delete this test. If we provide "asdf", surely the library will not accept that?
I would rather expect a more generic error here than just deleting the test.

pkg/setup/health_test.go
AminSlk: Fixed it with adding a channel to `mockBackoff`

sven-urbanski-freiheit-com (approved):

PR #2041, opened 2024-10-14
chore(rollout-service): Adapt rollout service to get information from the GetAppDetails endpoint

miguel-crespo-fdc:
Important notes:
* The rollout service no longer listens for to the stream overview endpoint. It now now listens to the GetChangedApps endpoint, that gets triggered when there is some change to any app, or when an environment is edited
* The rollout service still makes a request to the overview service to obtain the overview. This is done because there is important environment data there.
* The rollout service functionality should remain UNCHANGED.
Ref: SRX-FIC65Q


services/cd-service/pkg/notify/notify.go
```diff
@@ -16,11 +16,14 @@ Copyright freiheit.com*/
...
-import "sync"
+import (
+	"sync"
+)
 
 type Notify struct {
-	mx       sync.Mutex
-	listener map[chan struct{}]struct{}
+	mx                 sync.Mutex
+	listener           map[chan struct{}]struct{}
```
sven-urbanski-freiheit-com, line 25:
```suggestion
	overviewListener           map[chan struct{}]struct{}
```

services/cd-service/pkg/notify/notify.go
```diff
@@ -53,3 +56,31 @@ func (n *Notify) Notify() {
 		}
 	}
 }
+
+func (n *Notify) SubscribeChangesApps() (<-chan []string, Unsubscribe) {
```
sven-urbanski-freiheit-com, line 60: please also rename the other subscribe function to something like "SubscribeOverview".
That can be done in another PR if you prefer.

sven-urbanski-freiheit-com, line 60: please create a type for either `chan []string` or `[]string` to make clear that this is a list of app names and nothing else.

services/cd-service/pkg/repository/repository.go
```diff
@@ -1122,6 +1132,18 @@ func (r *repository) Apply(ctx context.Context, transformers ...Transformer) err
...
+		var seen = make(map[string]bool)
+		for _, app := range changes.ChangedApps {
+			if _, ok := seen[app.App]; !ok {
+				seen[app.App] = true
+				changedAppNames = append(changedAppNames, app.App)
+			}
+		}
+		if len(changedAppNames) != 0 {
+			r.notify.NotifyChangedApps(changedAppNames)
+		}
```
sven-urbanski-freiheit-com, line 1146: this code looks the same as in ProcessQueueOnce. Please extract this into a function.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -2792,6 +2792,17 @@ func (c *CreateEnvironment) Transform(
...
+		//Should be empty on new environments
+		envApps, err := state.GetEnvironmentApplications(ctx, transaction, c.Environment)
+		if err != nil {
+			return "", fmt.Errorf("Unable to read environment, error: %w", err)
+
+		}
+		for _, app := range envApps {
+			t.AddAppEnv(app, c.Environment, "")
+		}
+
```
sven-urbanski-freiheit-com, line 2805: if envApps should be empty, then why are we doing this look?
I think we can skip it.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -3069,6 +3080,7 @@ func (c *DeployApplicationVersion) Transform(
 		if err != nil {
 			return "", fmt.Errorf("could not write oldest deployment for %v - %v", newDeployment, err)
 		}
+
```
sven-urbanski-freiheit-com, line 3083: unnecessary change

services/cd-service/pkg/service/overview.go
```diff
@@ -46,10 +46,11 @@ type OverviewServiceServer struct {
 	RepositoryConfig repository.RepositoryConfig
 	Shutdown         <-chan struct{}
 
-	notify   notify.Notify
-	Context  context.Context
-	init     sync.Once
-	response atomic.Value
+	notify                       notify.Notify
+	Context                      context.Context
+	init                         sync.Once
```
sven-urbanski-freiheit-com, line 51: please rename `init`. Maybe `OverviewStreamingInitFun`?

services/cd-service/pkg/service/overview.go
```diff
@@ -372,6 +398,29 @@ func (o *OverviewServiceServer) subscribe() (<-chan struct{}, notify.Unsubscribe
 	return o.notify.Subscribe()
 }
 
+func (o *OverviewServiceServer) subscribeChangedApps() (<-chan []string, notify.Unsubscribe) {
+	o.changedAppsStreamingInitFunc.Do(func() {
+		ch, unsub := o.Repository.Notify().SubscribeChangesApps()
+		// Channels obtained from subscribe are by default triggered
+		//
+		// This means, we have to wait here until the first overview is loaded.
```
sven-urbanski-freiheit-com, line 406:
```suggestion
		// This means, we have to wait here until the changedApps are loaded for the first time.
```

services/cd-service/pkg/notify/notify.go
```diff
@@ -31,25 +34,55 @@ func (n *Notify) Subscribe() (<-chan struct{}, Unsubscribe) {
...
+		delete(n.changeAppsListener, ch)
+	}
+}
+
+func (n *Notify) NotifyChangedApps(changedApps ChangedAppNames) {
+	n.mx.Lock()
+	defer n.mx.Unlock()
+	for ch := range n.changeAppsListener {
+		select {
+		case ch <- changedApps:
```
hannesg, line 84: This doesn't work here. The original overview notifier was able to simply skip updates for slow subscribers because it always sends out the entire overview. If you want to send partial updates, you have to make sure that partial updates don't get lost.

hannesg (changes requested):

services/cd-service/pkg/notify/notify.go
sven-urbanski-freiheit-com: yes, what's missing is sending all apps, whenever a client connects to the streaming grpc call.

services/cd-service/pkg/service/overview.go
```diff
@@ -349,8 +351,48 @@ func (o *OverviewServiceServer) StreamOverview(in *api.GetOverviewRequest,
...
+	stream api.OverviewService_StreamChangedAppsServer) error {
+	ch, unsubscribe := o.subscribeChangedApps()
+	defer unsubscribe()
+	done := stream.Context().Done()
+	for {
+		select {
+		case <-o.Shutdown:
+			return nil
+		case changedAppsNames := <-ch:
+			if len(changedAppsNames) == 0 { //This only happens when a channel is first triggered, so we send all apps
```
miguel-crespo-fdc, line 364: This ensures that the first message to some listener is always the full list of apps

sven-urbanski-freiheit-com (approved):

PR #2049, opened 2024-10-16
feat(git): Allow exporting only strictly required data for Argo CD

sven-urbanski-freiheit-com:
Ref: SRX-2CSAEK

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -867,7 +867,7 @@ func (h *DBHandler) DBSelectReleasesByAppLatestEslVersion(ctx context.Context, t
 func (h *DBHandler) DBSelectReleasesByAppOrderedByEslVersion(ctx context.Context, tx *sql.Tx, app string, ignorePrepublishes bool) (*DBReleaseWithMetaData, error) {
 	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectReleasesByAppOrderedByEslVersion")
 	defer span.Finish()
-	selectQuery := h.AdaptQuery(fmt.Sprintf(`
+	selectQuery := h.AdaptQuery(`
 SELECT
 	releases.eslVersion,
 	releases.created,
@@ -888,7 +888,7 @@ JOIN releases
     AND c.releaseVersion = releases.releaseVersion
 ORDER BY c.releaseVersion DESC
 LIMIT 1;
-`),
+`,
 	)
 	span.SetTag("query", selectQuery)
 	span.SetTag("appName", app)


pkg/db/db.go
```diff
@@ -883,6 +883,33 @@ func (h *DBHandler) DBSelectReleasesByAppOrderedByEslVersion(ctx context.Context
...
 
+func (h *DBHandler) DBSelectLastReleasesByApp(ctx context.Context, tx *sql.Tx, app string, deleted bool, ignorePrepublishes bool) (*DBReleaseWithMetaData, error) {
+	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectReleasesByApp")
+	defer span.Finish()
+	selectQuery := h.AdaptQuery(fmt.Sprintf(
+		"SELECT eslVersion, created, appName, metadata, releaseVersion, deleted, environments " +
+			" FROM releases " +
+			" WHERE appName=? AND deleted=?" +
+			" ORDER BY eslVersion DESC, releaseVersion DESC" +
+			" LIMIT 1;"))
```
AminSlk, line 894: If we have two rows like this:
```
eslVersion: 3 releaseVersion: 2
eslVersion: 2 releaseVersion: 3
```
This query will return the releaseVersion 2 but it should return the 3.
We should use joins for this query, like the one [here](https://github.com/freiheit-com/kuberpult/blob/73f2cfdce8e8d17a891244576c0a54bda39ee68a/pkg/db/db.go#L826)

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -1017,13 +1059,14 @@ func isLatestVersion(state *State, application string, version uint64) (bool, er
 
 // Finds old releases for an application: Checks for the oldest release that is currently deployed on any environment
 // Releases older that the oldest deployed release are eligible for deletion. releaseVersionsLimit
-func findOldApplicationVersions(ctx context.Context, transaction *sql.Tx, state *State, name string) ([]uint64, error) {
+func findOldApplicationVersions(ctx context.Context, transaction *sql.Tx, state *State, versionLimit int, appName string) ([]uint64, error) {
```
AminSlk, line 1062: I didn't understand why we added `versionLimit` to the inputs of the function. Because it's always equal to `int(state.ReleaseVersionsLimit)` and we already have `state` as an input

pkg/db/db.go
```diff
@@ -2418,14 +2448,32 @@ func (h *DBHandler) processAppsRow(ctx context.Context, rows *sql.Rows, err erro
...
+	app, err := h.DBSelectApp(ctx, tx, appName)
+	if err != nil {
+		return nil, err
+	}
+	if app == nil {
+		return nil, nil
+	}
+	if app.StateChange == AppStateChangeDelete {
+		return nil, nil
+	}
```
AminSlk, line 2464: Can't we move this to `processAppRow` and use `DBSelectApp` instead?

services/cd-service/pkg/service/overview.go
```diff
@@ -190,31 +190,47 @@ func (o *OverviewServiceServer) GetAppDetails(
...
-			deployment := &api.Deployment{
-				Version:         uint64(*currentDeployment.Version),
-				QueuedVersion:   0,
-				UndeployVersion: false,
-				DeploymentMetaData: &api.Deployment_DeploymentMetaData{
-					CiLink:       currentDeployment.Metadata.CiLink,
-					DeployAuthor: currentDeployment.Metadata.DeployedByName,
-					DeployTime:   currentDeployment.Created.String(),
-				},
+			environment, err := o.DBHandler.DBSelectEnvironment(ctx, transaction, envName)
```
sven-urbanski-freiheit-com, line 193: we'll have to keep an eye on this in the stress tests, because this is an additional call for each env.

AminSlk (approved):

PR #2053, opened 2024-10-17
fix(db): fixed select query that ignored eslversions with deleted flag

diogo-nogueira-freiheit:
In the DeleteEnvFromApp a query was made which selected the most recent eslversions for each release of the app.
This query ignored rows that has the deleted flag set to true. When the transformer later updated each release if the lated eslversion was deleted then an error would occur.

This query was fixed and, in the process, altered to be more efficient.

Ref: SRX-9MM6SC

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -789,26 +789,53 @@ func (h *DBHandler) processReleaseManifestRows(ctx context.Context, err error, r
 	return result, nil
 }
 
-func (h *DBHandler) DBSelectReleasesByApp(ctx context.Context, tx *sql.Tx, app string, deleted bool, ignorePrepublishes bool) ([]*DBReleaseWithMetaData, error) {
+// DBSelectReleasesByAppLatestEslVersion returns the latest eslversion
+// for each release of an app. It includes deleted releases and loads manifests.
+func (h *DBHandler) DBSelectReleasesByAppLatestEslVersion(ctx context.Context, tx *sql.Tx, app string, ignorePrepublishes bool) ([]*DBReleaseWithMetaData, error) {
 	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectReleasesByApp")
 	defer span.Finish()
-	selectQuery := h.AdaptQuery(fmt.Sprintf(
-		"SELECT eslVersion, created, appName, metadata, manifests, releaseVersion, deleted, environments " +
-			" FROM releases " +
-			" WHERE appName=? AND deleted=?" +
-			" ORDER BY releaseVersion DESC, eslVersion DESC, created DESC;"))
+	selectQuery := h.AdaptQuery(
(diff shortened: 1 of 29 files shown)


pkg/db/db.go
```diff
@@ -789,26 +789,53 @@ func (h *DBHandler) processReleaseManifestRows(ctx context.Context, err error, r
 	return result, nil
 }
 
-func (h *DBHandler) DBSelectReleasesByApp(ctx context.Context, tx *sql.Tx, app string, deleted bool, ignorePrepublishes bool) ([]*DBReleaseWithMetaData, error) {
+// DBSelectReleasesByAppLatestEslVersion returns the latest eslversion
+// for each release of an app. It includes deleted releases and loads manifests.
+func (h *DBHandler) DBSelectReleasesByAppLatestEslVersion(ctx context.Context, tx *sql.Tx, app string, ignorePrepublishes bool) ([]*DBReleaseWithMetaData, error) {
 	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectReleasesByApp")
```
sven-urbanski-freiheit-com, line 795: the span name should be the same as the function name

pkg/db/db.go
```diff
@@ -789,26 +789,53 @@ func (h *DBHandler) processReleaseManifestRows(ctx context.Context, err error, r
...
 		app,
-		deleted,
 	)
 
 	return h.processReleaseRows(ctx, err, rows, ignorePrepublishes, true)
 }
 
-func (h *DBHandler) DBSelectReleasesByAppLatestEslVersion(ctx context.Context, tx *sql.Tx, app string, deleted bool, ignorePrepublishes bool) ([]*DBReleaseWithMetaData, error) {
+func (h *DBHandler) DBSelectReleasesByAppOrderedByEslVersion(ctx context.Context, tx *sql.Tx, app string, deleted bool, ignorePrepublishes bool) ([]*DBReleaseWithMetaData, error) {
 	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectReleasesByApp")
```
sven-urbanski-freiheit-com, line 839: please update the span name

pkg/db/db.go
```diff
@@ -789,27 +789,54 @@ func (h *DBHandler) processReleaseManifestRows(ctx context.Context, err error, r
...
+				appname, releaseversion
+		) as currentEslReleases
+		JOIN
+		  releases
+		ON 
+			currentEslReleases.appname 			= releases.appname
+		AND
+			currentEslReleases.latesteslversion = releases.eslversion
+		AND
+			currentEslReleases.releaseversion 	= releases.releaseversion;`,
```
sven-urbanski-freiheit-com, line 826: please add the ORDER BY back, even if it's just for consistency.

sven-urbanski-freiheit-com (approved):

PR #2061, opened 2024-10-23
fix(charts): manifest-repo-export-service pod now stores manifest repo in an emptyDir volume 

diogo-nogueira-freiheit:
Altered the mount point of the emptyDir volume for the manifeset-repo-export-service pod.
It now mounts to the home folder so that the service can create the repo from scratch and be the owner.
Altered the size from 10Gi to 30Gi.
The relevant documentation was added to the chart.

Ref: SRX-F26M7C

### DIFF
--- a/charts/kuberpult/templates/manifest-repo-export-service.yaml
+++ b/charts/kuberpult/templates/manifest-repo-export-service.yaml
@@ -239,8 +239,7 @@ spec:
           # The repository volume, an emptyDir, is mounted to the kp directory.
           # The manifest-repo-export-service creates the repository inside this directory in /kp/repository.
           # We mount the volume to the parent because kubernetes volumes belong to root.
-          # This way the container can create /kp/repository himself.
-          # DO NOT USE /kp TO STORE ANYTHING ELSE
+          # This way the container can create /kp/repository itself without any permission issues.
           mountPath: /kp/
         - name: ssh
           mountPath: /etc/ssh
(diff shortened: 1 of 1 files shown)


charts/kuberpult/templates/manifest-repo-export-service.yaml
```diff
@@ -235,6 +235,13 @@ spec:
         - name: KUBERPULT_RELEASE_VERSIONS_LIMIT
           value: "{{ .Values.git.releaseVersionsLimit }}"
         volumeMounts:
+        - name: repository
+          # The repository volume, an emptyDir, is mounted to the kp directory.
+          # The manifest-repo-export-service creates the repository inside this directory in /kp/repository.
+          # We mount the volume to the parent because kubernetes volumes belong to root.
+          # This way the container can create /kp/repository himself.
+          # DO NOT USE /kp TO STORE ANYTHING ELSE
```
sven-urbanski-freiheit-com, line 243: is it really accurate that we can't store anything else to /kp?
I thought the problem was that the export-service has to be the one to create the subdirectory.

diogo-nogueira-freiheit: That is correct. However, this now means that the kp directory is the emptyDir volume. Meaning that if someone stores anything on this directory it will use the 30 Gi that we alocate for the git repository. This could lead to the repo service being restarted (happens when an empty dir is filled) without the repository itself beign 30Gi of size which could lead to a confusing debuging process.

sven-urbanski-freiheit-com (approved):

PR #2066, opened 2024-10-28
fix(db): fixed function InsertAppFun function override & cleaned up update of lightweight  apps on overview

miguel-crespo-fdc:
InserAppFun was only getting overridden when we performed custom migrations. This led to a bug where when we initiated  a kuberpult instance without performing migrations (i.e. normal dev-env deployment or make kuberpult locally), we would add an InsertAppFun that did not update the overview.

Also added sql migration for deleting the old overview cache.

Ref: SRX-ZYK5HN

### DIFF
--- a/services/cd-service/pkg/cmd/server.go
+++ b/services/cd-service/pkg/cmd/server.go
@@ -344,7 +344,12 @@ func RunServer() {
 			&service.Service{
 				Repository: repo,
 			}
+
 		if dbHandler.ShouldUseOtherTables() {
+			// we overwrite InsertApp in order to also update the overview:
+			dbHandler.InsertAppFun = func(ctx context.Context, transaction *sql.Tx, appName string, previousEslVersion db.EslVersion, stateChange db.AppStateChange, metaData db.DBAppMetaData) error {
+				return repo.State().DBInsertApplicationWithOverview(ctx, transaction, appName, previousEslVersion, stateChange, metaData)
+			}
 			//Check for migrations -> for pulling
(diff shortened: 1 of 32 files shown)


services/cd-service/pkg/repository/repository.go
```diff
@@ -2536,6 +2514,7 @@ func (s *State) UpdateEnvironmentsInOverview(ctx context.Context, transaction *s
 	} else {
 		result.EnvironmentGroups = mapper.MapEnvironmentsToGroups(envs)
 		for envName, config := range envs {
+			logger.FromContext(ctx).Sugar().Warnf("Overview for env: %s", envName)
```
sven-urbanski-freiheit-com, line 2517: Either remove this line or set it to `Info`

services/cd-service/pkg/cmd/server.go
```diff
@@ -344,7 +344,12 @@ func RunServer() {
 			&service.Service{
 				Repository: repo,
 			}
+
 		if dbHandler.ShouldUseOtherTables() {
+			// we overwrite InsertApp in order to also update the overview:
+			dbHandler.InsertAppFun = func(ctx context.Context, transaction *sql.Tx, appName string, previousEslVersion db.EslVersion, stateChange db.AppStateChange, metaData db.DBAppMetaData) error {
```
sven-urbanski-freiheit-com, line 350: please fix the typos in the commit message: `FUn that did not upadte`

sven-urbanski-freiheit-com (approved):

PR #2067, opened 2024-10-28
fix(manifest-repo-export-service): measure pushes write when nothing to push

diogo-nogueira-freiheit:
The `measurePushes` method sends metrics to datadog about wether the manifest repo export service successfully pushed to git.
This PR makes the service write the metric, as a success, even when there are no pushes.

Ref: SRX-PLAY6B

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -49,7 +49,7 @@ This metric measures each failure in the manifest-repo-export-service to push to
 It measures 0 if there are currently no failures, and 1 if there are.
 
 This metric is allways written, even if there is nothing for kuberpult to push.
-In case kuberpult has nothing to push this metric writes 0.
+In case kuberpult has nothing to push this metric writes 0 every `manifestRepoExport.eslProcessingIdleTimeSeconds` seconds.
 
 
 #### Monitoring Processing Delay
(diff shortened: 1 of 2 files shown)


docs/database.md
```diff
@@ -48,7 +48,8 @@ Metric `Kuberpult.manifest_export_push_failures`.
 This metric measures each failure in the manifest-repo-export-service to push to the git repository.
 It measures 0 if there are currently no failures, and 1 if there are.
 
-This metric is only written, if there is something for kuberpult to push.
+This metric is allways written, even if there is nothing for kuberpult to push.
+In case kuberpult has nothing to push this metric writes 0.
```
sven-urbanski-freiheit-com, line 52:
```suggestion
In case kuberpult has nothing to push this metric writes 0 every `manifestRepoExport.eslProcessingIdleTimeSeconds` seconds.
```

sven-urbanski-freiheit-com (approved):

docs/database.md
diogo-nogueira-freiheit: Doesn't the time change based on the used `backoff.Backoff`. Which in this case is being configured as a type of exponential backoff?

sven-urbanski-freiheit-com: that's true, maybe it could be phrased better.
But in general (ignoring error cases) it should be running exactly like that, because we call `sleepDuration.Reset()` before the sleep.

PR #2077, opened 2024-10-31
feat(UI): reload button for each app

miguel-crespo-fdc:
Screenshot 2024-10-31 at 11 53 25
Ref: SRX-E9K6SV


services/frontend-service/src/ui/components/ServiceLane/ServiceLane.tsx
```diff
@@ -315,6 +325,16 @@ export const ReadyServiceLane: React.FC<{
         />
     );
 
+    const reloadButton = (
```
sven-urbanski-freiheit-com, line 328: i think the button might fit better to the right side of the lane (where "updated just now" appears as well)

sven-urbanski-freiheit-com (approved):

PR #2091, opened 2024-11-08
fix(cd-service): Added new check for non existing env

diogo-nogueira-freiheit:
Fixes a nil pointer when a manifest for a non-existing environment is submitted in the /release endpoint.

Ref: SRX-QCRRU1

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -1973,6 +1973,9 @@ func (s *State) checkUserPermissions(ctx context.Context, transaction *sql.Tx, e
 	if err != nil {
 		return err
 	}
+	if config == nil {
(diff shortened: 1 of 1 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -1973,6 +1973,9 @@ func (s *State) checkUserPermissions(ctx context.Context, transaction *sql.Tx, e
 	if err != nil {
 		return err
 	}
```
sven-urbanski-freiheit-com, line 1975: Looks good.
Suggestion for Commit Message body:
`Fixes a nil pointer when a manifest for a non-existing environment is submitted in the /release endpoint."

sven-urbanski-freiheit-com (approved):

PR #2100, opened 2024-11-14
fix(rollout-service): improved logging

diogo-nogueira-freiheit:
The rollout service is crashing and the logs do not provide useful feedback.
This PR adds logs to improve visibility in the kuberpult background task shutdown process.

Ref: SRX-FEUSCN

### DIFF
--- a/pkg/setup/setup.go
+++ b/pkg/setup/setup.go
@@ -292,7 +292,7 @@ func runBackgroundTask(ctx context.Context, config BackgroundTaskConfig, cancel 
 	} else {
 		select {
 		case <-ctx.Done():
-			// task finished with no error and context was cancelled, normal shutdown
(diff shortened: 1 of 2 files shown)


pkg/setup/setup.go
```diff
@@ -289,6 +289,13 @@ func runBackgroundTask(ctx context.Context, config BackgroundTaskConfig, cancel
 	defer cancel()
 	if err := config.Run(ctx, reporter); err != nil {
 		logger.FromContext(ctx).Error("background.error", zap.Error(err), zap.String("job", config.Name))
+	} else {
+		select {
+		case <-ctx.Done():
+			// task finished with no error and context was cancelled, normal shutdown
```
sven-urbanski-freiheit-com, line 295: We do not need to be so careful with adding log lines when it's about rare cases like shutdown (or startup).
I would suggest to turn this comment into a log (maybe INFO level)

PR #2104, opened 2024-11-18
fix: Speed up metrics collection by using the overview cache

sven-urbanski-freiheit-com:
We already have the number of locks in the overview cache,
so there's no need to calculate it again.

This also removes some dead code.

Ref: SRX-Z8BHWD

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -268,6 +268,9 @@ func UpdateLockMetrics(ctx context.Context, transaction *sql.Tx, state *State, n
 	if err != nil {
 		return err
 	}
+	if overviewCache == nil {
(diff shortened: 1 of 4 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -278,10 +271,17 @@ func UpdateLockMetrics(ctx context.Context, transaction *sql.Tx, state *State, n
 			return fmt.Errorf("failed to read environment from the db: %v", err)
 		}
 
+		cache, err := state.DBHandler.ReadLatestOverviewCache(ctx, transaction)
+		if err != nil {
+			return err
+
+		}
```
AminSlk, line 278: Also, we should check if cache is nil or empty

AminSlk (approved):

PR #2120, opened 2024-11-20
fix(trivy): re-enable trivy scan & upgrade k8s.io dependency

AminSlk:
Ref: SRX-TXJ83H


.github/workflows/trivy-scan.yml
```diff
@@ -46,7 +46,6 @@ jobs:
           exit 1
         fi
     - name: Run trivy scan
```
sven-urbanski-freiheit-com, line 48: question: did you run `go mod tidy` in the end?

sven-urbanski-freiheit-com (approved):

.github/workflows/trivy-scan.yml
AminSlk: Yes, I did.

PR #2121, opened 2024-11-21
fix: do not run overview twice when creating environments

sven-urbanski-freiheit-com:
Note that the create environments endpoint is still recalculating the whole overview cache, which is inefficient. This will be fixed in another PR.

Ref: SRX-0F9ZN3

### DIFF
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -2517,42 +2517,78 @@ func (s *State) UpdateEnvironmentsInOverview(ctx context.Context, transaction *s
 			var groupName = mapper.DeriveGroupName(config, envName)
 			var envInGroup = getEnvironmentInGroup(result.EnvironmentGroups, groupName, envName)
 
-			argocd := &api.EnvironmentConfig_ArgoCD{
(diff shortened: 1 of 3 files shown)


services/cd-service/pkg/repository/repository.go
```diff
@@ -2517,113 +2517,121 @@ func (s *State) UpdateEnvironmentsInOverview(ctx context.Context, transaction *s
...
-			}
-			env := api.Environment{
-				DistanceToUpstream: 0,
-				Priority:           api.Priority_PROD,
-				Name:               envName,
-				Config: &api.EnvironmentConfig{
-					Upstream:         mapper.TransformUpstream(config.Upstream),
-					Argocd:           argocd,
-					EnvironmentGroup: &groupName,
+			err2 := s.UpdateEnvironmentInternal(ctx, transaction, config, envName, groupName, envInGroup)
```
AminSlk, line 2520: We're still recalculating the whole overview. Shouldn't we put a filter to update only the created environment?

sven-urbanski-freiheit-com: yes, but that's more complicated than i thought, so i split off this PR.

AminSlk (approved):

PR #2122, opened 2024-11-21
fix: create environment does not update complete overview

sven-urbanski-freiheit-com:
Ref: SRX-0F9ZN3

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -2840,7 +2840,8 @@ func (c *CreateEnvironment) Transform(
 	if state.DBHandler.ShouldUseOtherTables() {
 		// write to environments table
 		environmentApplications := make([]string, 0)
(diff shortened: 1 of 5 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -2878,7 +2878,8 @@ func (c *CreateEnvironment) Transform(
 		if err != nil {
 			return "", fmt.Errorf("Unable to read overview cache, error: %w", err)
 		}
-		err = state.UpdateEnvironmentsInOverview(ctx, transaction, overview)
+		logger.FromContext(ctx).Sugar().Warnf("SU DEBUG: updating env %s with group %v", c.Environment, c.Config.EnvironmentGroup)
```
AminSlk, line 2881: I think we can delete this log

AminSlk (approved):

PR #2145, opened 2024-11-28
fix(rollout-service): new retry strategy for kuberpult events

diogo-nogueira-freiheit:
Currently the rollout service loses updates from the cd-service if the self-manage thread is busy handling argo updates. This PR adds two different changes to help this problem:
* When events fail to be pushed they are stored and pushed along the next update;
* The channel used for pushes was given a capacity of 5 instead of 0 to improve push success rate;

Ref: SRX-FEUSCN

### DIFF
--- a/services/rollout-service/pkg/versions/versions.go
+++ b/services/rollout-service/pkg/versions/versions.go
@@ -19,9 +19,10 @@ package versions
 import (
 	"context"
 	"fmt"
-	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 	"strconv"
 	"time"
+
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 
 	"github.com/argoproj/argo-cd/v2/pkg/apiclient/application"
 	"github.com/freiheit-com/kuberpult/services/rollout-service/pkg/argo"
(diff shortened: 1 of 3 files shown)


services/rollout-service/pkg/versions/versions.go
```diff
@@ -281,8 +283,13 @@ func (v *versionClient) ConsumeEvents(ctx context.Context, processor VersionEven
 
 			}
 
-			l.Info("version.push")
-			v.ArgoProcessor.Push(ctx, &overview)
+			overview.AppDetails = appsToChange
+			if err := v.ArgoProcessor.Push(ctx, &overview); err != nil {
+				l.Sugar().Warnf("version.push failed: %v", err)
+			} else {
+				l.Info("version.push")
```
AminSlk, line 290: Is this log needed?

AminSlk (approved):

services/rollout-service/pkg/versions/versions.go
diogo-nogueira-freiheit: This log is essential telling us when there is a push and if it fails. Since we now have a retry strategy it is not an error when a push fails, but I think we should still be able to see it in the logs that pushes have been failing so that we now why changes are not being applied to argocd

PR #2151, opened 2024-12-01
fix(rollout-service): prioritize kuberpult events over argocd events in rollout background job

AminSlk:
Ref: SRX-0F9ZN3

### DIFF
--- a/services/rollout-service/pkg/argo/argo.go
+++ b/services/rollout-service/pkg/argo/argo.go
@@ -99,29 +99,37 @@ func (a *ArgoAppProcessor) Consume(ctx context.Context, hlth *setup.HealthReport
 	l := logger.FromContext(ctx).With(zap.String("self-manage", "consuming"))
 	appsKnownToArgo := map[string]map[string]*v1alpha1.Application{} //EnvName => AppName => Deployment
 	envAppsKnownToArgo := make(map[string]*v1alpha1.Application)
-	for {
-		select {
-		case argoOv := <-a.trigger:
-			l.Info("self-manage.trigger")
-			overview := argoOv.Overview
-			for currentApp, currentAppDetails := range argoOv.AppDetails {
-				span, ctx := tracer.StartSpanFromContext(ctx, "ProcessChangedApp")
-				defer span.Finish()
-				span.SetTag("kuberpult-app", currentApp)
-				for _, envGroup := range overview.EnvironmentGroups {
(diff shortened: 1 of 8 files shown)


services/rollout-service/pkg/argo/argo.go
```diff
@@ -122,26 +122,31 @@ func (a *ArgoAppProcessor) Consume(ctx context.Context, hlth *setup.HealthReport
...
+					l.Info("created/updated:kuberpult.application:" + ev.Application.Name + ",kuberpult.environment:" + envName)
+					envKnownToArgo[appName] = &ev.Application
+				case "DELETED":
+					l.Info("deleted:kuberpult.application:" + ev.Application.Name + ",kuberpult.environment:" + envName)
+					delete(envKnownToArgo, appName)
+				}
+				appsKnownToArgo[envName] = envKnownToArgo
+			case <-ctx.Done():
+				return nil
+			}
```
diogo-nogueira-freiheit, line 149: If we have multiple kuberpult events and don't have any argo watch events isn't there a chance we block here and cannot process kuberpult self-manage events for a long time?

services/rollout-service/pkg/service/broadcast.go
```diff
@@ -305,7 +305,7 @@ func (b *Broadcast) Start() ([]*BroadcastEvent, <-chan *BroadcastEvent, unsubscr
 	for key, app := range b.state {
 		result = append(result, app.getEvent(key.Application, key.Environment))
 	}
-	ch := make(chan *BroadcastEvent, 100)
+	ch := make(chan *BroadcastEvent, 100000)
```
diogo-nogueira-freiheit, line 308: Couldn't this very high capacity bring memory problems? I'm making the question because there will allways be multiple broadcast channels (not sure how many but at least 3). And this could become a very large load on the rollout service.

I think increasing broadcast channel sizes is a good idea, the number just seems a bit too high. How did you come up with it?

AminSlk: There are about 1000 applications and 70 envs, which mean 70000 appEnvs. and at the startup we will have a new message for each of them in the channel. I just assumed the worst case.
With rough estimation, I think it's gonna take 50Mi memory, which I think is ok, but maybe we can decrease it.

diogo-nogueira-freiheit (approved):

PR #2164, opened 2024-12-09
fix(cd-service): improve get app details deployments

diogo-nogueira-freiheit:
The `GetAppDetails` endpoint collects information about the deployments of an app.
One of these is the latest queued version. Currently the endpoint would evaluate one deployment at a time and, when it was decided to be a valid deployment, the database would be queried for this deployment's latest queued version.

This PR queries the database for all latest deployment attempts. Then when analyzing each deployment it just pulls from this information cached in a map.

This might slow down `GetAppDetails` requests that have no current deployments (since before we would not query the queued versions for invalid deployments). This will translate in improved performance. And in most cases, most apps will have deployments

Ref: SRX-016Y9E

### DIFF
--- a/pkg/db/db_test.go
+++ b/pkg/db/db_test.go
@@ -25,8 +25,10 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"slices"
 	"sort"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/db/db_test.go
```diff
@@ -25,8 +25,10 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"slices"
```
sven-urbanski-freiheit-com, line 28: for the commit message, I would add something like this:
"this will translate in improved performance. And in most cases, most apps will have deployments"

PR #2170, opened 2024-12-12
fix(db): separate current state of the releases from releases in history

AminSlk:
Ref: SRX-2HNB0H


database/migrations/postgres/1732637175997202_releases_autoincrement.up.sql
```diff
@@ -1,4 +1,12 @@
-ALTER TABLE IF EXISTS releases ADD COLUMN IF NOT EXISTS version INTEGER;
+DO $$
```
sven-urbanski-freiheit-com, line 1: Why are we changing an old migration? We should generally not do that, except if it's a bug fix.

database/migrations/postgres/1733415120086267_deployments_autoincrement.up.sql
```diff
@@ -23,7 +23,23 @@ SELECT setval('deployments_version_seq', coalesce(max(version), 0) + 1, false) F
 ALTER TABLE IF EXISTS deployments
 ALTER COLUMN version SET DEFAULT nextval('deployments_version_seq');
 
-ALTER TABLE IF EXISTS deployments DROP CONSTRAINT IF EXISTS deployments_pkey;
+DO $$
```
sven-urbanski-freiheit-com, line 26: same, why are we changing an old migration? Please add a new migration instead

database/migrations/postgres/1733867056189286_environments_autoincrement.up.sql
```diff
@@ -24,7 +24,23 @@ SELECT setval('environments_version_seq', coalesce(max(row_version), 0) + 1, fal
 ALTER TABLE IF EXISTS environments
 ALTER COLUMN row_version SET DEFAULT nextval('environments_version_seq');
 
-ALTER TABLE IF EXISTS environments DROP CONSTRAINT IF EXISTS environments_pkey;
+DO $$
```
sven-urbanski-freiheit-com, line 27: same here

pkg/db/db_releases.go
```diff
@@ -0,0 +1,681 @@
...
+	Environments  []string
+}
+
+// SELECTS
+
+func (h *DBHandler) DBSelectAnyRelease(ctx context.Context, tx *sql.Tx, ignorePrepublishes bool) (*DBReleaseWithMetaData, error) {
+	selectQuery := h.AdaptQuery(`
+	SELECT created, appName, metadata, releaseVersion, environments 
+	FROM releases 
+	LIMIT 1;`)
```
sven-urbanski-freiheit-com, line 62:
```suggestion
		SELECT created, appName, metadata, releaseVersion, environments 
		FROM releases 
		LIMIT 1;`)
```

please separate the query visually from the rest of the code.

sven-urbanski-freiheit-com: (same for other queries)

pkg/db/db_releases.go
```diff
@@ -0,0 +1,681 @@
...
+		err = h.insertReleaseRow(ctx, transaction, release)
+		if err != nil {
+			return err
+		}
+	} else {
+		err = h.updateReleaseRow(ctx, transaction, release)
+		if err != nil {
+			return err
+		}
+	}
```
sven-urbanski-freiheit-com, line 252: This should probably be an "upsert" instead of 3 queries. Something like this:
```
insert into apps (created, appname, statechange, metadata)
VALUES (now(), 'echo-3', 'AppStateChangeModify', '')
ON CONFLICT(version, appname)
DO UPDATE Set statechange = EXCLUDED.statechange, metadata = EXCLUDED.metadata ;
```

pkg/db/db_releases.go
```diff
@@ -0,0 +1,681 @@
...
+			release.App,
+			release.ReleaseNumber,
+			err)
+	}
+	return nil
+}
+
+func (h *DBHandler) updateReleaseRow(ctx context.Context, transaction *sql.Tx, release DBReleaseWithMetaData) error {
+	insertQuery := h.AdaptQuery(`
+		UPDATE releases SET created=?, manifests=?, metadata=?, environments=?
```
sven-urbanski-freiheit-com, line 328: (nitpick)
```suggestion
		UPDATE releases
		SET created=?, manifests=?, metadata=?, environments=?
```

pkg/db/db_releases.go
```diff
@@ -0,0 +1,681 @@
...
+		WHERE appname=? AND releaseVersion=?;`)
+	span, ctx := tracer.StartSpanFromContext(ctx, "updateReleaseRow")
```
sven-urbanski-freiheit-com, line 330: creating the span should always be the first line in a function.

pkg/db/db_releases.go
```diff
@@ -0,0 +1,681 @@
...
+		ctx,
+		selectQuery,
+	)
+
+	return h.processReleaseManifestRows(ctx, err, rows)
+}
+
+func (h *DBHandler) DBSelectReleasesByAppLatestEslVersion(ctx context.Context, tx *sql.Tx, app string, ignorePrepublishes bool) ([]*DBReleaseWithMetaData, error) {
+	selectQuery := h.AdaptQuery(`
+	SELECT created, appName, metadata, manifests, releaseVersion, environments
```
sven-urbanski-freiheit-com, line 165: it's really nice to see such simple SQL queries!

sven-urbanski-freiheit-com (changes requested):
.

database/migrations/postgres/1732637175997202_releases_autoincrement.up.sql
AminSlk: it's actually a bug fix. The issue is that it wasn't idempotent. So, after the new migrations are applied and then we rerun the migrations we get an error. The integration tests were failing because of that.
But, it wouldn't be a problem if these migrations are already applied somewhere.

database/migrations/postgres/1733415120086267_deployments_autoincrement.up.sql
AminSlk: Same.

database/migrations/postgres/1733867056189286_environments_autoincrement.up.sql
AminSlk: Same.

pkg/db/db_releases.go
AminSlk: Done

pkg/db/db_releases.go
AminSlk: Done

pkg/db/db_releases.go
```diff
@@ -0,0 +1,681 @@
...
+		release.ReleaseNumber,
+		release.App,
+		manifestJson,
+		metadataJson,
+		deleted,
+		environmentStr,
+	)
+	if err != nil {
+		return fmt.Errorf(
+			"could not insert release for app '%s' and version '%v' into DB. Error: %w\n",
```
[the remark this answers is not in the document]
AminSlk: Done

pkg/db/db_releases.go
AminSlk: Done

pkg/db/db_releases.go
AminSlk: Done.

sven-urbanski-freiheit-com (approved):

PR #2176, opened 2024-12-17
fix(cd-service): less calls to db select app in releasetrains

AminSlk:
Ref: SRX-625RAW

### DIFF
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -3223,6 +3223,33 @@ func (s *State) GetApplicationTeamOwner(ctx context.Context, transaction *sql.Tx
 	}
 }
 
+func (s *State) GetAllApplicationsTeamOwner(ctx context.Context, transaction *sql.Tx) (map[string]string, error) {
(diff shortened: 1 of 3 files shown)


services/cd-service/pkg/repository/repository.go
```diff
@@ -3223,6 +3223,33 @@ func (s *State) GetApplicationTeamOwner(ctx context.Context, transaction *sql.Tx
...
+		}
+		for _, app := range apps {
+			teamOwner, err := s.GetApplicationTeamOwnerFromManifest(app)
+			if err != nil {
+				return result, err
+			}
+			result[app] = teamOwner
+		}
+		return result, nil
+	}
```
sven-urbanski-freiheit-com, line 3250: this code will never be run in the cd-service, please remove it.

sven-urbanski-freiheit-com (approved):

services/cd-service/pkg/repository/repository.go
AminSlk: We still have some tests that use this part of the code

PR #2179, opened 2025-01-02
feat(overview): Implement new endpoint for getting all app locks

miguel-crespo-fdc:
Added new endpoint to get all application locks

Ref: SRX-JTRV7H

### DIFF
--- a/services/cd-service/pkg/service/overview.go
+++ b/services/cd-service/pkg/service/overview.go
@@ -329,34 +329,31 @@ func (o *OverviewServiceServer) GetAllAppLocks(ctx context.Context,
 		response := api.GetAllAppLocksResponse{
 			AllAppLocks: make(map[string]*api.AllAppLocks),
 		}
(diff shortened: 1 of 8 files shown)


services/cd-service/pkg/service/overview.go
```diff
@@ -312,6 +312,55 @@ func (o *OverviewServiceServer) GetOverview(
...
+		}
+
+		response := api.GetAllAppLocksResponse{
+			AllAppLocks: make(map[string]*api.AllAppLocks),
+		}
+		for _, appName := range allAppNames {
+			if err != nil {
+				return nil, err
+			}
+			appLocks, err := o.DBHandler.DBSelectAllActiveAppLocksForApp(ctx, transaction, appName)
```
AminSlk, line 336: Can't we get all app locks for all apps at once instead of iterating over apps? It might cause performance issues.

AminSlk (approved):

PR #2182, opened 2025-01-03
chore: remove app locks from overview

miguel-crespo-fdc:
Ref: SRX-JTRV7H


database/migrations/postgres/1735910780961964_delete_overview_cache_app_locks.up.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM overview_cache;
```
miguel-crespo-fdc, line 1: We changed the overview structure. As such, we need to recalculate the overview

services/frontend-service/src/ui/Pages/Locks/LocksPage.test.tsx
```diff
@@ -668,26 +607,26 @@ describe('Test Team locks', () => {
...
-                            locks: [
-                                {
-                                    message: 'lockfoo',
-                                    lockId: 'ui-v2-123',
-                                    createdAt: new Date(1995, 11, 16),
-                                },
-                            ],
-                        },
-                    },
+                    // appLocks: {
```
AminSlk: Why are these lines commented? Can't we just delete them?

AminSlk (approved):

PR #2197, opened 2025-01-13
fix(cd-service): put a maximum on the number of threads for releasetrain

AminSlk:
Ref: SRX-AC1K9Y


services/cd-service/pkg/repository/transformer.go
```diff
@@ -3717,9 +3723,17 @@ func (c *ReleaseTrain) Transform(
 	span.SetTag("environments", len(envNames))
 	if state.DBHandler.ShouldUseOtherTables() && isEnvGroup {
 		var wg sync.WaitGroup
```
hannesg, line 3725: I think using an errgroup would be a lot simpler it has a max-concurrency feature build in.

miguel-crespo-fdc (approved):

PR #2204, opened 2025-01-17
chore(cd-service): remove env/team locks from overview

AminSlk:
Ref: SRX-4ARVJW

* Removed all env locks and team locks from overview
* Added new endpoint to retrieve all env and team locks
* Adapted all places to use the new endpoint instead of overview


services/cd-service/pkg/service/overview.go
```diff
@@ -357,6 +357,67 @@ func (o *OverviewServiceServer) GetAllAppLocks(ctx context.Context,
 	})
 }
 
+func (o *OverviewServiceServer) GetAllEnvLocks(ctx context.Context,
```
miguel-crespo-fdc, line 360: This endpoint does not have a very good name. I would expect that this endpoint to return only the env locks, but it also returns the team locks. I think the name should convey this

PR #2208, opened 2025-01-20
feat: more explicit logging about skipped notifies

bjoern-michaelsen-freiheit:
Ref: SRX-TQ1N6K


services/rollout-service/pkg/revolution/revolution.go
```diff
@@ -117,25 +117,56 @@ func (s *Subscriber) subscribeOnce(ctx context.Context, b *service.Broadcast) er
...
+	}
+	return true
+}
 
-func shouldNotify(old *service.BroadcastEvent, nu *service.BroadcastEvent) bool {
+func shouldNotify(ctx context.Context, old *service.BroadcastEvent, nu *service.BroadcastEvent) bool {
+	l := logger.FromContext(ctx).With(zap.String("revolution", "processing"))
 	// check for fields that must be present to generate the request
-	if nu.ArgocdVersion == nil || nu.IsProduction == nil || nu.ArgocdVersion.SourceCommitId == "" {
+	if !shouldNotifyValidate(l, "new", nu) {
```
miguel-crespo-fdc, line 151: "new" seems to be hardcoded here, we could just remove this parameter and hardcode it to the log message inside the function

bjoern-michaelsen-freiheit: ack, done. was a leftover from when I thought we were validating the old and new event in a similar way ... => removed

miguel-crespo-fdc (approved):

PR #2211, opened 2025-01-21
chore(cd-service): remove overview cache completely

AminSlk:
This removes the overview_cache table, but keeps the overview as a grpc call.

Note: If this commit has to be rolled-back, the table `overview_cache` needs to be recreated

Ref: SRX-HHHRTQ


database/migrations/postgres/1737376354546130_remove_overview_cache.up.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS overview_cache;
```
sven-urbanski-freiheit-com, line 1: looks good, I just changed the commit message. Please check if it's good.

sven-urbanski-freiheit-com (approved):

PR #2216, opened 2025-01-23
chore: implement sync status on cd-service and manifest export

miguel-crespo-fdc:
cd-service now writes to git_sync_data table upon transformer execution
manifest-export-service now writes to git_sync_data table upon transformer execution or failure
Ref: SRX-IV81DS

### DIFF
--- a/database/migrations/postgres/1737404419146074_git_sync_status.up.sql
+++ b/database/migrations/postgres/1737404419146074_git_sync_status.up.sql
@@ -1,10 +1,9 @@
 CREATE TABLE IF NOT EXISTS git_sync_status
 (
-    eslVersion SERIAL,
     created TIMESTAMP,
     transformerID  INTEGER,
     envName VARCHAR,
     appName VARCHAR,
     status INTEGER,
-    PRIMARY KEY(eslVersion, appName, envName)
+    PRIMARY KEY(appName, envName)
 );
--- a/database/migrations/postgres/1737404419146074_git_sync_status.up.sql
+++ b/database/migrations/postgres/1737404419146074_git_sync_status.up.sql
@@ -1,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS git_sync_status
+(
+    eslVersion SERIAL,
+    created TIMESTAMP,
+    transformerID  INTEGER,
+    envName VARCHAR,
+    appName VARCHAR,
+    status INTEGER,
+    PRIMARY KEY(eslVersion, appName, envName)
+);
--- a/pkg/db/db_sync_status.go
+++ b/pkg/db/db_sync_status.go
@@ -18,6 +18,8 @@ const (
 	SYNC_FAILED
 )
 
+const BULK_INSERT_BATCH_SIZE = 2
+
 type GitSyncData struct {
 	AppName       string
 	EnvName       string
@@ -68,18 +70,9 @@ func (h *DBHandler) DBWriteNewSyncEventBulk(ctx context.Context, tx *sql.Tx, id 
 	if err != nil {
 		return fmt.Errorf("DBWriteNewSyncEventBulk unable to get transaction timestamp: %w", err)
 	}
-	insertAllQuery := "INSERT INTO git_sync_status (created, transformerid, envName, appName, status) VALUES"
-	for idx, currComb := range envApps {
-		format := " ('%s', %d, '%s', '%s', %d)"
-		if idx == len(envApps)-1 {
-			format += ";"
-		} else {
-			format += ","
-		}
-		insertAllQuery += fmt.Sprintf(format, now.Format(time.RFC3339), id, currComb.EnvName, currComb.AppName, status)
-	}
-	_, err = tx.Exec(
-		insertAllQuery)
+
+	err = h.executeBulkInsert(ctx, tx, envApps, *now, id, status)
+
 	if err != nil {
 		return fmt.Errorf("could not write sync event into DB. Error: %w\n", err)
 	}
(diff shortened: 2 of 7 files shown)


database/migrations/postgres/1737404419146074_git_sync_status.up.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS git_sync_status
+(
```
sven-urbanski-freiheit-com, line 2: I think we should remove the eslVersion.

Our new approach is to split tables into "current" and "history".

However, for the git status, I don't think we want to store the history.
I would suggest to only store the current version in this table.

pkg/db/db_sync_status.go
```diff
@@ -0,0 +1,244 @@
...
+	if tx == nil {
+		return fmt.Errorf("DBWriteNewSyncEventBulk: no transaction provided")
+	}
+	now, err := h.DBReadTransactionTimestamp(ctx, tx)
+	if err != nil {
+		return fmt.Errorf("DBWriteNewSyncEventBulk unable to get transaction timestamp: %w", err)
+	}
+
+	err = h.executeBulkInsert(ctx, tx, envApps, *now, id, status, BULK_INSERT_BATCH_SIZE)
+
```
sven-urbanski-freiheit-com, line 91: nitpick: odd empty lie
```suggestion
```

pkg/db/db_sync_status.go
```diff
@@ -0,0 +1,244 @@
...
+	return nil
+}
+
+type EnvApp struct {
+	AppName string
+	EnvName string
+}
+
+func (h *DBHandler) DBReadUnsyncedAppsForTransfomerID(ctx context.Context, tx *sql.Tx, id TransformerID) ([]EnvApp, error) {
+	span, ctx := tracer.StartSpanFromContext(ctx, "DBReadUnsyncedAppsForTransfomerID")
```
sven-urbanski-freiheit-com, line 104: please use the function `tracing.StartSpanFromContext`, so the span is marked as error

sven-urbanski-freiheit-com: (in all functions that return an error)

pkg/db/db_sync_status.go
```diff
@@ -0,0 +1,244 @@
...
+	if tx == nil {
+		return fmt.Errorf("DBBulkUpdateUnsyncedApps: no transaction provided")
+	}
+
+	allCombs, err := h.DBReadUnsyncedAppsForTransfomerID(ctx, tx, id)
+	if err != nil {
+		return fmt.Errorf("DBBulkUpdateUnsyncedApps unable to read unsynced apps: %w", err)
+	}
+	if len(allCombs) == 0 {
+		logger.FromContext(ctx).Sugar().Warnf("Could not update all unsynced apps. Did not find any unsynced apps.")
```
sven-urbanski-freiheit-com, line 166: that's not really a problem, is it?
Either remove the warning or turn it into "info".

pkg/db/db_sync_status.go
```diff
@@ -0,0 +1,244 @@
...
+			_, err := tx.ExecContext(ctx, currentQuery)
+			if err != nil {
+				return err
+			}
+			currentQuery = queryPrefix
+		} else {
+			format += ","
+			currentQuery += fmt.Sprintf(format, now.Format(time.RFC3339), id, currComb.EnvName, currComb.AppName, status)
+		}
+	}
```
sven-urbanski-freiheit-com, line 242: I think this code would be more readible, if we just repeat the whole query.
So instead of
`insert into ... VALUES () () ();`
It would be 
`insert into ... VALUES (); insert into ... VALUES (); insert into ... VALUES ();`

pkg/db/db_test.go
```diff
@@ -4093,6 +4093,366 @@ func TestReadAllManifestsAllReleases(t *testing.T) {
...
+					TransformerID: EslVersion(2),
+					SyncStatus:    UNSYNCED,
+				},
+				{
+					AppName:       "app-3",
+					EnvName:       "env-3",
+					TransformerID: EslVersion(3),
+					SyncStatus:    UNSYNCED,
+				},
+				{
```
sven-urbanski-freiheit-com, line 4123: could we add a 2nd case here?
Could be exactly the same, but one addition appEnv that is already synced.

pkg/db/db_test.go
```diff
@@ -4093,6 +4093,366 @@ func TestReadAllManifestsAllReleases(t *testing.T) {
...
+				if err != nil {
+					if diff := cmp.Diff(tc.expectedError, err, cmpopts.EquateErrors()); diff != "" {
+						t.Fatalf("error mismatch (-want, +got):\n%s", diff)
+
+					}
+					return nil
+				}
+
+				apps, err := dbHandler.DBReadUnsyncedAppsForTransfomerID(ctx, transaction, TransformerID(0))
+				if err != nil {
```
sven-urbanski-freiheit-com, line 4324: please also compare the error if it is nil

pkg/db/db_test.go
```diff
@@ -4093,6 +4093,366 @@ func TestReadAllManifestsAllReleases(t *testing.T) {
...
+				t.Fatalf("error while running the transaction error: %v", err)
+			}
+		})
+	}
+}
+
+func TestBulkInsertFunction(t *testing.T) {
+	tcs := []struct {
+		Name          string
+		NumberOfApps  int
```
sven-urbanski-freiheit-com, line 4257:
```suggestion
		ExpectedNumberOfApps  int
```

sven-urbanski-freiheit-com (approved):

PR #2230, opened 2025-01-30
feat(manifest-export-service): get sync status non-streaming endpoint

miguel-crespo-fdc:
Ref: SRX-7N10GA


pkg/db/db_sync_status.go
```diff
@@ -174,24 +174,32 @@ func (h *DBHandler) DBBulkUpdateUnsyncedApps(ctx context.Context, tx *sql.Tx, id
 	return h.executeBulkInsert(ctx, tx, allCombs, *now, id, status, BULK_INSERT_BATCH_SIZE)
 }
 
-func (h *DBHandler) DBRetrieveSyncStatus(ctx context.Context, tx *sql.Tx, appName, envName string) (*GitSyncData, error) {
+func (h *DBHandler) DBRetrieveAppsByStatus(ctx context.Context, tx *sql.Tx, status SyncStatus) ([]GitSyncData, error) {
```
miguel-crespo-fdc, line 177: the diff made this weird :/

pkg/api/v1/api.proto
```diff
@@ -58,6 +59,28 @@ message GetCommitInfoResponse {
...
+
+message GetGitSyncStatusResponse {
+  repeated EnvApp unsynced = 1;
+  repeated EnvApp sync_failed = 2;
+}
+
+message EnvApp {
+  string application_name = 1;
+  string environment_name = 2;
+  GitSyncStatus status    = 3;
```
AminSlk, line 81: I don't understand. If these are going to be in the `unsynced` or `sync_failed` list, then why would we need the status field here? We already know that they're unsynced or sync_failed!

AminSlk (approved):

PR #2239, opened 2025-02-04
fix(cd): migration to delete foreign key on deployments and deployment history

miguel-crespo-fdc:
Parallel release trains spawn new transactions. This means that for a release train we will have one outer transaction and N inner transactions (based on the number of parallel processing go routines and number of environments to process). As the outer transaction has not yet been committed, inner transactions will not be able to view the most recent event written to the event_sourcing_light table. This ALWAYS results in a FK violation if anything is deployed by the release train, as the transformer ID that the deployments try to reference does not exist on the current database view of the inner transaction. 

Rolling back the outer transaction causes inner transactions to also be rolled back, so simply removing these fk_constrains works and kuberpult won't be left in an inconsistent state (i.e. deployments exist that don't reference any transformer on the event_sourcing_light table). 

In my opinion, removing the FK constraint is the only solution, as we want to keep using parallel transactions to perform the release trains for different environments in parallel.

Ref: SRX-AC1K9Y

### DIFF
--- a/database/migrations/postgres/1720428454051313_fk_deployments.up.sql
+++ b/database/migrations/postgres/1720428454051313_fk_deployments.up.sql
@@ -1,7 +1,7 @@
 DO $$ BEGIN IF NOT EXISTS (
     SELECT column_name
     FROM information_schema.columns
-    WHERE table_name='deployments' and column_name='transformerEslversion'
+    WHERE table_name='deployments' and column_name='transformereslversion'
 ) THEN
 ALTER TABLE deployments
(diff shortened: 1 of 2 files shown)


database/migrations/postgres/1720428454051313_fk_deployments.up.sql
```diff
@@ -1,10 +1,2 @@
-DO $$ BEGIN IF NOT EXISTS (
```
miguel-crespo-fdc, line 1: This is needed to make this migration idempotent. ElsID dos not exist if we run this twice (it gets renamed on later migrations)

AminSlk (approved):

PR #2253, opened 2025-02-10
feat(export-service): skip failed esl events endpoint. Migrated to current and history for failed esl events

miguel-crespo-fdc:
* We now have two failed esl events tables. One for keeping history (same as before) and another one for keeping the currently failed esl events
* Implemented endpoint for skipping the failed esl events (removes them from the event_sourcing_light_failed table and keeps the Git Sync Status unchanged)

Ref: SRX-6GJRWL


database/migrations/postgres/1738926689513374_separate_esl_failed_events.up.sql
```diff
@@ -0,0 +1,45 @@
+-- rename deployments table to deployments_history if it doesn't exist
```
AminSlk, line 1: Comments are not adapted :)
`event_sourcing_light_failed` table

PR #2258, opened 2025-02-13
chore(cd-service): remove git related stuff from transformers

AminSlk:
Ref: SRX-12OQXN

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"math"
 	"net/url"
+	"os"
 	"reflect"
 	"regexp"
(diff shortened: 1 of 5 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -3413,69 +2457,7 @@ type Overview struct {
...
-			if version == nil {
-				continue
-			}
-
-			resp = append(resp, Overview{App: appName, Version: *version})
-		}
-	}
-	return resp, nil
-}
-
```
EdSwordsmith, line 3458: I'm currently working on refactoring this function so it uses the DB instead of git. So I need you not to delete this.

EdSwordsmith (approved):

PR #2259, opened 2025-02-14
chore: return map instead of lists for GitSyncStatus

miguel-crespo-fdc:
Ref: SRX-XM1UN2


services/frontend-service/src/ui/components/GitSyncStatusDescription/GitSyncStatusDescription.tsx
```diff
@@ -14,27 +14,20 @@ along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>
 
 Copyright freiheit.com*/
 
-export enum GitSyncStatus {
```
miguel-crespo-fdc, line 17: We now use the enum on the api

AminSlk (approved):

PR #2279, opened 2025-02-27
feat(cli): add delete environment command

EdSwordsmith:
Add command for deleting environments to the CLI
Ref: SRX-8XR09K

### DIFF
--- a/cli/pkg/environments/parsing.go
+++ b/cli/pkg/environments/parsing.go
@@ -48,7 +48,7 @@ func readDeleteEnvironmentArgs(args []string) (*DeleteEnvironmentCommandLineArgu
 		return nil, fmt.Errorf("error while parsing command line arguments, error: %w", err)
 	}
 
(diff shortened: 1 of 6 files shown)


cli/pkg/environments/parsing.go
```diff
@@ -0,0 +1,82 @@
...
+
+	fs := flag.NewFlagSet("flag set", flag.ContinueOnError)
+
+	fs.Var(&cmdArgs.environment, "environment", "the environment you are trying to delete")
+
+	if err := fs.Parse(args); err != nil {
+		return nil, fmt.Errorf("error while parsing command line arguments, error: %w", err)
+	}
+
+	if len(fs.Args()) != 0 { // kuberpult-cli release does not accept any positional arguments, so this is an error
```
miguel-crespo-fdc, line 49:
```suggestion
	if len(fs.Args()) != 0 { // kuberpult-cli delete environment does not accept any positional arguments, so this is an error
```

miguel-crespo-fdc (approved):

PR #2280, opened 2025-02-28
fix(cd): remove git repo interactions

EdSwordsmith:
Ref: SRX-8DV0NY

PR #2282, opened 2025-03-03
chore(cd-service): remove all git related codes from cd-service

AminSlk:
Ref: SRX-L1AZGF
This contains changes of #2280 and more.


services/cd-service/pkg/sqlitestore/sqlite.c
```diff
@@ -1,215 +0,0 @@
-/*This file is part of kuberpult.
```
miguel-crespo-fdc, line 1: Should we be deleting this?

miguel-crespo-fdc (approved):

PR #2284, opened 2025-03-03
fix(manifest-export): deployments don't check for locks anymore on export service

miguel-crespo-fdc:
Fixed bug where the export service would crash with minimized git data was on and we undeployed an application

We don't check for locks because either the Deployment:

* Was generated by a release train, in which case the release train
transformer checks what was deployed by the releases train,
* Was generated by a CreateApplicationVersion, that also checks what deployments were generated by that transformer.
* Was generated by a CreateUndeployApplicationVersion, that also checks what deployments were generated by that transformer.
* Was generated by the UI, which always has  api.LockBehavior = IGNORE.

This way we don't have to check for which transformer generated the current we are processing, which would be
costly for big release trains that generate a lot of deployments.

Ref: SRX-HNENMA


services/manifest-repo-export-service/pkg/repository/transformer_test.go
```diff
@@ -2382,119 +2381,6 @@ func TestCreateUndeployLogic(t *testing.T) {
 				},
 			},
 		},
-		{
-			Name: "Try Create undeploy application version, but it is locked",
```
miguel-crespo-fdc, line 2386: This test does not make sense anymore. If it is locked, the deployment should not appear

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -1855,8 +1828,14 @@ func (u *UndeployApplication) Transform(
 
 		undeployFile := fs.Join(versionDir, "undeploy")
 		_, err = fs.Stat(undeployFile)
-		if err != nil && errors.Is(err, os.ErrNotExist) {
-			return "", fmt.Errorf("UndeployApplication(repo): error cannot un-deploy application '%v' the release on '%v' is not un-deployed: '%v'", u.Application, env, undeployFile)
+		if err != nil { //Undeploy version does not exist if we minimize git data
```
miguel-crespo-fdc, line 1831: We don't write the undeploy version file if we minimize git data.

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -364,54 +364,20 @@ func (c *DeployApplicationVersion) Transform(
...
-		app, err := state.DBHandler.DBSelectExistingApp(ctx, transaction, c.Application)
-		if err != nil {
-			return "", err
-		}
-		var teamName = ""
-		if app != nil && app.Metadata.Team != "" {
-			teamName = app.Metadata.Team
-		}
+	/*
+		We don't check for locks because either the Deployment:
```
AminSlk, line 369: I think there's no need for this comment in this part of code. It could just be in the PR description.

AminSlk (approved):

PR #2286, opened 2025-03-03
chore(metrics): git sync status metrics are reported more often

miguel-crespo-fdc:
Ref: SRX-8FMYI8

### DIFF
--- a/services/cd-service/pkg/service/git.go
+++ b/services/cd-service/pkg/service/git.go
@@ -257,10 +257,6 @@ func (s *GitServer) ReadSyncStatuses(ctx context.Context) (*api.GetGitSyncStatus
 		}
 		response.AppStatuses = toApiStatuses(append(unsyncedStatuses, syncFailedStatuses...))
 
-		err = repository.MeasureGitSyncStatus(len(unsyncedStatuses), len(syncFailedStatuses))
-		if err != nil { //Don't fail transaction because we can't send data to datadog.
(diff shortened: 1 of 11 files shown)


services/manifest-repo-export-service/pkg/cmd/server_test.go
```diff
@@ -171,86 +141,3 @@ func SetupRepositoryTestWithDBOptions(t *testing.T, writeEslOnly bool) (reposito
 	}
 	return repo, dbHandler
 }
-
-func TestMeasureGitSyncStatus(t *testing.T) {
```
miguel-crespo-fdc, line 175: Test was moved to repository_test.go

services/cd-service/pkg/service/git.go
```diff
@@ -256,8 +256,14 @@ func (s *GitServer) ReadSyncStatuses(ctx context.Context) (*api.GetGitSyncStatus
 			return err
 		}
 		response.AppStatuses = toApiStatuses(append(unsyncedStatuses, syncFailedStatuses...))
+
+		err = repository.MeasureGitSyncStatus(len(unsyncedStatuses), len(syncFailedStatuses))
```
AminSlk, line 260: Why are we sending these metrics both in the `ReadSyncStatus` method and in `processEsls`? isn't the `processEsls` enough?
I think we should update the metrics only when the values are changing, not when we're reading them.

PR #2288, opened 2025-03-04
chore(manifest-repo): Remove checks from CreateApplicationVersion

miguel-crespo-fdc:
We need to remove Kuberpult logic from export service. This means that we only error out when some interaction with the file system fails.

Ref: SRX-E3IK25

### DIFF
--- a/services/manifest-repo-export-service/pkg/repository/transformer.go
+++ b/services/manifest-repo-export-service/pkg/repository/transformer.go
@@ -762,9 +762,6 @@ func (c *CreateApplicationVersion) Transform(
 ) (string, error) {
 	version := c.Version
 	fs := state.Filesystem
-	if !valid.ApplicationName(c.Application) {
-		return "", GetCreateReleaseAppNameTooLong(c.Application, valid.AppNameRegExp, uint32(valid.MaxAppNameLen))
-	}
 
 	releaseDir := releasesDirectoryWithVersion(fs, c.Application, version)
 	appDir := applicationDirectory(fs, c.Application)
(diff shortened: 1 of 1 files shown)


services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -1009,19 +970,6 @@ func writeNextPrevInfo(ctx context.Context, sourceCommitId string, otherCommitId
 	return nil
 }
 
-func isLatestVersion(ctx context.Context, state *State, transaction *sql.Tx, application string, version uint64) (bool, error) {
```
miguel-crespo-fdc, line 1012: dead code

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -903,12 +871,10 @@ func (c *CreateApplicationVersion) Transform(
...
 			return "", err
 		}
+
 		tCtx.AddAppEnv(c.Application, env, teamOwner)
 
-		if _, exists := deploymentsMap[env]; !exists { //If this transformer did not generate any deployments, skip the deployment transformer
-			continue
-		}
-		if hasUpstream && config.Upstream.Latest && isLatest {
+		if _, exists := deploymentsMap[env]; exists { //If this transformer did not generate any deployments, skip the deployment transformer
```
miguel-crespo-fdc, line 877: Only check we need is to figure out if the CreateApplicationVersion transformer generated any deployments for

AminSlk (approved):

PR #2289, opened 2025-03-04
feat(charts): add option to disable export service

AminSlk:
Ref: SRX-A90GM4

### DIFF
--- a/charts/kuberpult/templates/manifest-repo-export-service.yaml
+++ b/charts/kuberpult/templates/manifest-repo-export-service.yaml
@@ -18,7 +18,7 @@
 {{ fail ".Values.db.dbOption does not contain a valid value (postgreSQL)."}}
 {{ end -}}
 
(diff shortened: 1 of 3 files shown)


charts/kuberpult/templates/manifest-repo-export-service.yaml
```diff
@@ -19,7 +19,7 @@
 {{ end -}}
 
 # the export service is only enabled, if the DB is enabled
```
miguel-crespo-fdc, line 21: This comment is not really accurate anymore, maybe we should delete it?

miguel-crespo-fdc (approved):

charts/kuberpult/templates/manifest-repo-export-service.yaml
AminSlk: Adapted the comment

PR #2297, opened 2025-03-11
feat(cd-service): bring back custom migrations

miguel-crespo-fdc:
Reverting PR where we removed the custom Migrations.
We now also offer a helm chart parameter that to check for custom migrations. If flag is disabled, the cd-service does not wait for the manifest export service to startup.

Ref: SRX-E3IK25


charts/kuberpult/tests/charts_test.go
```diff
@@ -639,6 +647,44 @@ db:
...
+		{
+			Name: "Check for custom Migrations",
+			Values: `
+git:
+  url:  "testURL"
+ingress:
+  domainName: "kuberpult-example.com"
+db:
+  dbOption: "postgreSQL"
+  checkCustomMigrations: false
```
miguel-crespo-fdc, line 659: New tests

charts/kuberpult/values.yaml
```diff
@@ -162,6 +162,9 @@ db:
     rollout:
       maxOpen: 100
       maxIdle: 20
+  # Kuberpult only checks for custom migrations if this flag is enabled. Otherwise, it skips them entirely, assuming
+  # that there is no information we want to migrate from an existing manifest repository to the database.
```
sven-urbanski-freiheit-com, line 166: We should provide clear guidelines on when this flag should be used.
```suggestion
  # that there is no information we want to migrate from an existing manifest repository to the database.
  # You should only enable this flag, if you want to do a one-step upgrade of kuberpult from a (relatively old) version that still uses git as storage.
  # If you are already using the database, ignore this flag.
```

database/migrations/postgres/1738234757185160_remove_custom_migrations.up.sql
```diff
@@ -1 +0,0 @@
-DROP TABLE IF EXISTS custom_migration_cutoff;
```
sven-urbanski-freiheit-com, line 1: please make sure that all releases (on github) that still have one of these sql files, are set to "prerelease:true"

services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -628,3 +702,7 @@ func checkReleaseVersionLimit(limit uint) error {
 	}
 	return nil
 }
+
+func shouldRunCustomMigrations(checkCustomMigrations, minimizeGitData bool) bool {
+	return checkCustomMigrations && !minimizeGitData //If minimize git ata is enabled we can't make sure we have all the information on the repository to perform all the migrations
```
sven-urbanski-freiheit-com, line 707:
```suggestion
	return checkCustomMigrations && !minimizeGitData //If `minimizeGitData` is enabled we can't make sure we have all the information on the repository to perform all the migrations
```

sven-urbanski-freiheit-com (approved):

PR #2300, opened 2025-03-13
feat(manifest-export): configurable empty dir size

miguel-crespo-fdc:
Making empty dir size for repository volume mount  configurable

Ref : SRX-4RL2SE

### DIFF
--- a/charts/kuberpult/templates/cd-service.yaml
+++ b/charts/kuberpult/templates/cd-service.yaml
@@ -50,7 +50,6 @@ spec:
       labels:
         app: kuberpult-cd-service
 {{- if .Values.datadogTracing.enabled }}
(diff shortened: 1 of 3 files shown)


charts/kuberpult/tests/charts_test.go
```diff
@@ -1481,6 +1497,67 @@ db:
 	}
 }
 
+func TestVolumes(t *testing.T) {
+	tcs := []struct {
+		Name            string
+		Values          string
+		VolumeMountSize string
+		target          core.Volume
+		VolumeName      string
```
AminSlk, line 1506: `VolumeName` in all cases is "repository". doesn't have to be inside the testcases struct. Can be a constant value.

AminSlk (approved):

PR #2301, opened 2025-03-13
fix(cd-service): isolate some transformers

AminSlk:
Ref: SRX-FIMM1N

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -1313,7 +1313,7 @@ type CleanupOldApplicationVersions struct {
 }
 
 func (c *CleanupOldApplicationVersions) GetDBEventType() db.EventType {
-	panic("CleanupOldApplicationVersions GetDBEventType")
+	return db.EvtCleanupOldApplicationVersions
 }
 
 func (c *CleanupOldApplicationVersions) SetEslVersion(id db.TransformerID) {
@@ -2794,7 +2794,7 @@ type envReleaseTrain struct {
 }
 
 func (c *envReleaseTrain) GetDBEventType() db.EventType {
-	panic("envReleaseTrain GetDBEventType")
+	return db.EvtEnvReleaseTrain
 }
 
 func (c *envReleaseTrain) SetEslVersion(id db.TransformerID) {
@@ -3296,7 +3296,7 @@ type skippedServices struct {
 }
 
 func (c *skippedServices) GetDBEventType() db.EventType {
-	panic("GetDBEventType for skippedServices")
+	return db.EvtSkippedServices
 }
 
 func (c *skippedServices) SetEslVersion(id db.TransformerID) {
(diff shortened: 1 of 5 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -365,8 +366,25 @@ type transformerRunner struct {
 	DeletedRootApps []RootApp
 }
 
+var isolatedTransformersLock sync.RWMutex
+var isolatedTransformerNames = []string{"UndeployApplication", "DeleteEnvFromApp", "DeleteEnvironment"}
+
 func (r *transformerRunner) Execute(ctx context.Context, t Transformer, transaction *sql.Tx) error {
 	r.Stack = append(r.Stack, nil)
+	tranformerTypeName := reflect.TypeOf(t).Name()
```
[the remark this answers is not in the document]
sven-urbanski-freiheit-com: You just need to make sure that all transformers implement this function, but I think they do already

services/cd-service/pkg/repository/transformer.go
```diff
@@ -365,8 +366,25 @@ type transformerRunner struct {
...
+	isIsolated := false
```
sven-urbanski-freiheit-com, line 375:
```suggestion
	requiresIsolation := false
```

services/cd-service/pkg/repository/transformer.go
```diff
@@ -365,8 +366,25 @@ type transformerRunner struct {
...
+	for _, isolatedTransformerName := range isolatedTransformerNames {
+		if isolatedTransformerName == tranformerTypeName {
+			isIsolated = true
+		}
+	}
```
sven-urbanski-freiheit-com, line 380: I think this would work for now, but it's not future proof.
If we scale up the cd-service to 2 pods, then this solution is not effective anymore.
At minimum, we need a story to come up with a long term solution.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -365,8 +366,25 @@ type transformerRunner struct {
...
+	if isIsolated {
+		isolatedTransformersLock.Lock()
+		defer isolatedTransformersLock.Unlock()
+	} else {
+		isolatedTransformersLock.RLock()
+		defer isolatedTransformersLock.RUnlock()
+	}
```
sven-urbanski-freiheit-com, line 387: I think this requires a bit of explanation:
```suggestion
	if isIsolated {
	        // we protect all "destructive" operations by a read-write lock, so that only 1 destructive operation can be run in parallel:
		isolatedTransformersLock.Lock()
		defer isolatedTransformersLock.Unlock()
	} else {
	        // we also use a read lock, so that destructive and non-destructive transformers cannot run in parallel:
		isolatedTransformersLock.RLock()
		defer isolatedTransformersLock.RUnlock()
	}
```

miguel-crespo-fdc (approved):

PR #2303, opened 2025-03-17
feat(manifest-export): WIP

miguel-crespo-fdc:


pkg/db/db_argo_events.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+type ArgoEvent struct {
+	App       string
+	Env       string
+	JsonEvent []byte
+	Discarded bool
+}
+
+func (h *DBHandler) InsertArgoEvents(ctx context.Context, tx *sql.Tx, events []*ArgoEvent) error {
```
AminSlk, line 36: I would name this UpsertArgoEvents or WriteArgoEvents

services/rollout-service/pkg/service/service.go
```diff
@@ -45,10 +48,29 @@ var (
...
+	ArgoEventsBatchSize int
+}
+
+func ConsumeEvents(ctx context.Context, params *ConsumeEventsParameters) error {
+	hlth := params.HealthReporter
+	appClient := params.AppClient
+	dispatcher := params.Dispatcher
+	a := params.ArgoAppProcessor
+	ddMetrics := params.DDMetrics
+	handler := params.DBHandler
```
AminSlk, line 71: Not sure how difficult to implement this but I think the better way is to define a struct Called `EventsConsumer` with these fields. Then `ConsumeEvents` should be a method of that struct, Like this:
```
func (e *EventConsumer) ConsumeEvents(ctx context.Context) error {
...
```
and then use these params whenever needed like this: `e.AppClient`

services/rollout-service/pkg/service/service.go
```diff
@@ -104,6 +127,29 @@ func ConsumeEvents(ctx context.Context, appClient SimplifiedApplicationServiceCl
 						logger.FromContext(ctx).Sugar().Warnf("could not send argo_events_fill_rate metric to datadog! Err: %v", ddError)
 					}
 				}
+
+				if sentEvent && params.PersistArgoEvents {
```
AminSlk, line 131: You could just use `argoEvent != nil` here. No need to define a new variable

AminSlk (approved):

PR #2304, opened 2025-03-17
fix(UI): rollback/advancing on side bar item

miguel-crespo-fdc:
Brought back feature of showing off many versions we were rolling back or advancing on side bar deploy actions.
Screenshot 2025-03-17 at 14 04 17

Ref: SRX-41ZF5J

### DIFF
--- a/services/frontend-service/src/ui/components/SideBar/SideBar.tsx
+++ b/services/frontend-service/src/ui/components/SideBar/SideBar.tsx
@@ -34,6 +34,7 @@ import {
     useApplications,
     useAppDetails,
     AppDetailsResponse,
+    InvalidateAppLocks,
 } from '../../utils/store';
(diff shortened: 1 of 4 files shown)


services/frontend-service/src/ui/components/SideBar/SideBar.tsx
```diff
@@ -510,6 +538,11 @@ export const SideBar: React.FC<{ className?: string }> = (props) => {
                     appNamesToInvalidate.push(action.action.deploy.application);
                 }
                 if (action.action?.$case === 'deleteEnvironmentApplicationLock') {
+                    appLocksToInvalidate.push({
```
miguel-crespo-fdc, line 541: App Locks were not being removed from store, which lead to a UI bug where we would get a warning that we were deploying services over app locks that had previously been deleted

AminSlk (approved):

PR #2318, opened 2025-03-26
feat(compliance): add commit hash to the retrieved data

miguel-crespo-fdc:
Ref: SRX-IQ7NL4

### DIFF
--- a/pkg/db/db_releases.go
+++ b/pkg/db/db_releases.go
@@ -110,6 +110,33 @@ func (h *DBHandler) DBSelectReleasesByVersions(ctx context.Context, tx *sql.Tx, 
 	return h.processReleaseRows(ctx, err, rows, ignorePrepublishes, false)
 }
 
+func (h *DBHandler) DBSelectAllReleasesWithVersion(ctx context.Context, tx *sql.Tx, releaseVersions []uint64, ignorePrepublishes bool) ([]*DBReleaseWithMetaData, error) {
+	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectReleasesByVersions")
+	defer span.Finish()
+	if len(releaseVersions) == 0 {
+		return []*DBReleaseWithMetaData{}, nil
+	}
+	repeatedQuestionMarks := strings.Repeat(",?", len(releaseVersions)-1)
+	selectQuery := h.AdaptQuery(`
+		SELECT created, appName, metadata, releaseVersion, environments FROM releases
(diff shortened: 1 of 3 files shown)


services/cd-service/pkg/service/overview.go
```diff
@@ -732,50 +733,113 @@ func (o *OverviewServiceServer) StreamDeploymentHistory(in *api.DeploymentHistor
...
 		}
-
+		type ReleaseInfo struct {
+			appName        string
+			releaseVersion uint64
+		}
+		startDateReleases := in.StartDate.AsTime().AddDate(0, -1, 0)
+		var releases = make(map[ReleaseInfo]string)
+		//Get releases for which we found any relevant deployment. We want to extract the commit hash for that release
 		query := o.DBHandler.AdaptQuery(`
```
AminSlk, line 749: I think we can move all these query parts to a function in db_releases.go

services/cd-service/pkg/service/overview.go
```diff
@@ -732,50 +733,113 @@ func (o *OverviewServiceServer) StreamDeploymentHistory(in *api.DeploymentHistor
...
+			SELECT appName, metadata, releaseVersion FROM releases
```
AminSlk, line 750: Not sure, but shouldn't we look at the `releases_history` table instead. For example if a release is deleted, do we want to count it in the compliance or not?

services/cd-service/pkg/service/overview.go
miguel-crespo-fdc: The second query we can't move, as we need to send the streaming data as we are processing the rows

pkg/db/db_releases.go
```diff
@@ -634,3 +634,74 @@ func (h *DBHandler) processAllAppsReleaseVersionsRows(ctx context.Context, err e
...
+	}
+	if transaction == nil {
+		return nil, fmt.Errorf("DBSelectCommitHashesTimeWindow: no transaction provided")
+	}
+
+	var releases = make(map[ReleaseKey]string)
+	//Get releases for which we found any relevant deployment. We want to extract the commit hash for that release
+	query := h.AdaptQuery(`
+			SELECT appName, metadata, releaseVersion FROM releases_history
+			WHERE releaseversion IS NOT NULL AND created >= (?) AND created <= (?);
```
AminSlk, line 657:
```suggestion
			WHERE releaseversion IS NOT NULL AND created >= (?) AND created <= (?) ORDER BY version;
```

AminSlk (approved):

PR #2327, opened 2025-04-03
fix(rollout-service): fix max age value in dora configs

AminSlk:
Ref: SRX-IQ7NL4

### DIFF
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -458,6 +458,7 @@ revolution:
     token: ""
     # Maximum number of requests send in parallel.
     concurrency: 20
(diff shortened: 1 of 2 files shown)


charts/kuberpult/values.yaml
```diff
@@ -458,6 +458,7 @@ revolution:
     token: ""
     # Maximum number of requests send in parallel.
     concurrency: 20
+    maxAge: 0s
```
sven-urbanski-freiheit-com, line 461: I think  this should be `2h` to keep the default the same.

sven-urbanski-freiheit-com (approved):

PR #2329, opened 2025-04-03
fix(rollout): fix dora events. improve observability of rollout service

miguel-crespo-fdc:
Ref: SRX-BPKKYY


charts/kuberpult/values.yaml
```diff
@@ -457,6 +470,10 @@ revolution:
     token: ""
     # Maximum number of requests send in parallel.
     concurrency: 20
+
+    # If enabled, revolution isn't actually contacted with new events and Kuberpult simply logs the events it would have sent.
+    # Useful if you don't want to overload Revolution with events.
```
sven-urbanski-freiheit-com, line 475:
```suggestion
    # Useful if you don't want to overload Revolution with events while debugging.
```

services/rollout-service/pkg/service/service.go
```diff
@@ -102,23 +101,14 @@ func (e *ArgoEventConsumer) ConsumeEvents(ctx context.Context) error {
 					logger.FromContext(ctx).Sugar().Warnf("argo apps channel at full capacity of %d. Discarding event: %v", cap(e.ArgoAppProcessor.ArgoApps), ev)
 				}
 
-				if e.DDMetrics != nil { //If DD is enabled, send metrics
+				if e.DDMetrics != nil && e.ArgoAppProcessor.ArgoAppsMetricsEnabled { //If DD is enabled, send metrics
```
sven-urbanski-freiheit-com, line 104: Please add an early exit (probably in the start of this function) if `e.ArgoAppProcessor == nil`

services/rollout-service/pkg/argo/argo.go
```diff
@@ -102,15 +110,18 @@ func (a *ArgoAppProcessor) Consume(ctx context.Context, hlth *setup.HealthReport
...
 			return nil
 		default:
 			select {
 			case argoOv := <-a.trigger:
 				l.Info("self-manage.trigger")
 				a.ProcessArgoOverview(ctx, l, appsKnownToArgo, argoOv)
+				a.GaugeKuberpultEventsQueueFillRate(ctx)
 			case ev := <-a.ArgoApps:
 				a.ProcessArgoWatchEvent(ctx, l, appsKnownToArgo, ev)
+				a.GaugeArgoAppsQueueFillRate(ctx)
```
sven-urbanski-freiheit-com, line 124: this function is only called once. It should be called twice for both adding and  removing from the channel.

services/rollout-service/pkg/revolution/revolution.go
```diff
@@ -189,17 +206,24 @@ func (s *Subscriber) notify(ctx context.Context, ev *service.BroadcastEvent) fun
...
 		span.SetTag("environment", ev.Environment)
 		span.SetTag("application", ev.Application)
+
 		body, err := json.Marshal(event)
 		if err != nil {
 			return fmt.Errorf("marshal event: %w", err)
 		}
+
+		if s.dryRun {
+			logger.FromContext(ctx).Sugar().Infof("Dry Run enabled! Would send following DORA event to revolution: %s", string(body))
```
sven-urbanski-freiheit-com, line 224: should this be info or warn level?

services/rollout-service/pkg/argo/argo.go
miguel-crespo-fdc: It is done on func (e *ArgoEventConsumer) ConsumeEvents. I've changed it to be more explicit

PR #2337, opened 2025-04-10
feat(db): release error handling, db lock type, removed read-before-write

sven-urbanski-freiheit-com:
0) With this change, the endpoint to create new releases (/release and /api/release) will now return an error (status 400) when a manifest for a non-existing environment was supplied.

1) Replaced in-memory locks with db locks

Note that this feature is considered experimental.

With this change, we have three options in the helm chart ("cd.lockType") to lock:
* "go": with an in-memory (golang) lock (default)
* "db": with a postgres advisory lock
* "none": without any locks

The first option was already implemented, the other two options are new. The postgres locks are an improvement over the go-locks, because they are not limited to the current pod/process. Apart from that they work the same.

Both lock types ensure that only one destructive action (undeploy, delete env from app, delete env) is running at a time.

2) Removed "Read-before-Writes"

There were 2 points where the cd-service read from the DB before writing:
When adding an app to an environment, and when removing an app from an environment.
Both have been simplified to one sql statement.

Ref: SRX-8JRR7Q

### DIFF
--- a/infrastructure/scripts/create-testdata/create-many-releases.sh
+++ b/infrastructure/scripts/create-testdata/create-many-releases.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+# create-many-releases.sh MIN MAX
+# Where MIN is the first release number used, and MAX is the last. Make sure that MAX>=MIN and MIN>=0.
+# This script creates a few releases, while in parallel we call undeploy.
+# This is useful to test the lock feature (db.lockType).
+# To check if the test is successful or not:
+# select applications from environments where name='development'; select appname from releases where appname='echo-???'
+# The test is successful if EACH application either appear in both tables, or in neither. See query.sql for the test.
+
+cd "$(dirname "$0")" || exit 1
+
+set -u
+
+MIN=${1}
+MAX=${2}
+
+./create-environments.sh
+
+# Setup: ensure the app exists:
+for appId in $(seq -w "${MIN}" "${MAX}" ); do
+  RELEASE_VERSION=99999 ./create-release.sh "echo-${appId}";
+done
+
+echo "Done with setup, starting test..."
+
+# Test:
+for appId in $(seq -w "${MIN}" "${MAX}" ); do
+  app="echo-${appId}"
+  RELEASE_VERSION=1 ./create-release.sh "${app}" &
+  (
+    echo '{
+"actions": [
+  {
+    "prepare_undeploy": {
+      "application": "'"${app}"'"
+    }
+  },
+  {
+    "undeploy": {
+      "application": "'"${app}"'"
+    }
+  }
+]
+}' | \
+    evans --header author-name=YXV0aG9y --header author-email=YXV0aG9yQGF1dGhvcg== --host localhost --port 8443 -r cli call api.v1.BatchService.ProcessBatch \
+    && echo "app ${appId} deletion success" \
+    || echo "app ${appId} deletion failed"
+  ) &
+done
+
+echo "waiting for all calls to finish..."
+wait
+echo "ALL DONE."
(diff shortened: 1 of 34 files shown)


infrastructure/scripts/create-testdata/create-many-releases.sh
```diff
@@ -0,0 +1,49 @@
...
+# This script creates a few releases, while in parallel we call undeploy.
+# This is useful to test the lock feature (db.lockType).
+# For repeated runs, adapt the "MIN" and "MAX" variables below.
+# To check if the test is successful or not:
+# select applications from environments where name='development'; select appname from releases where appname='echo-???'
+# The test is successful if EACH application either appear in both tables, or in neither.
+
+cd "$(dirname "$0")" || exit 1
+
+MIN=120
```
AminSlk, line 12: Would be good if the script took the min and max from the input params

infrastructure/scripts/create-testdata/create-many-releases.sh
```diff
@@ -0,0 +1,49 @@
...
+  for appId in $(seq -w $MIN $MAX ); do
+    RELEASE_VERSION=99999 ~/projects/fdc/kuberpult/infrastructure/scripts/create-testdata/create-release.sh "echo-${appId}";
+  done
+
+  echo "Done with setup, starting test..."
+
+# Test:
+  for appId in $(seq -w $MIN $MAX ); do
+    app="echo-${appId}"
+    RELEASE_VERSION=$v ~/projects/fdc/kuberpult/infrastructure/scripts/create-testdata/create-release.sh "${app}" &
```
AminSlk, line 28: Isn't the $v always 1 here?

infrastructure/scripts/create-testdata/create-release.sh
```diff
@@ -80,10 +80,10 @@ data:
   releaseVersion: "${release_version[@]}"
 ---
 EOF
-  echo "wrote file ${file}"
+#  echo "wrote file ${file}"
   manifests+=("--form" "manifests[${env}]=@${file}")
 done
-echo commit id: "${commit_id}"
+#echo commit id: "${commit_id}"
```
AminSlk, line 86: Do we need these commented codes?

services/cd-service/pkg/service/batch.go
```diff
@@ -408,17 +429,32 @@ func (d *BatchServer) ProcessBatch(
 	if requiresIsolation { // This solution is not scalable and doesn't work when we have multiple cd-service pods. Ref: SRX-8JRR7Q
```
AminSlk, line 429: We can move this comment to the line 433. Also, we can remove the reference to the story.

infrastructure/scripts/create-testdata/create-release.sh
sven-urbanski-freiheit-com: I suppose not, I removed them now

services/cd-service/pkg/service/batch.go
sven-urbanski-freiheit-com: done

AminSlk (approved):

infrastructure/scripts/create-testdata/create-many-releases.sh
```diff
@@ -0,0 +1,50 @@
...
+set -u
+
+MIN=${1}
+MAX=${2}
+
+./create-environments.sh
+
+# Setup: ensure the app exists:
+for appId in $(seq -w "${MIN}" "${MAX}" ); do
+  RELEASE_VERSION=99999 ~/projects/fdc/kuberpult/infrastructure/scripts/create-testdata/create-release.sh "echo-${appId}";
```
AminSlk, line 22: There shouldn't be absolute path here. For some other person it may not be in `project/fdc...`

sven-urbanski-freiheit-com: done

AminSlk (approved):

sven-urbanski-freiheit-com:
This PR depends on #2346

miguel-crespo-fdc (approved):

PR #2365, opened 2025-05-05
feat(ui): Suggested Lifetime in the locks tables

miguel-crespo-fdc:
Screenshot 2025-05-05 at 17 49 41

Ref: SRX-JU2G64


services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -1557,3 +1557,21 @@ class GitSyncStatusGetter {
         return mostInteresting;
     }
 }
+
+export const GetTargetFutureDate = (current: Date | undefined, increment: string): Date | undefined => {
```
AminSlk, line 1561: It doesn't seem correct to put this function in the `store.tsx`. It doesn't have anything do with `storing` things:)
I would say for now we can put it in `LockDisplay.tsx`.

AminSlk (approved):

PR #2370, opened 2025-05-10
chore(db): do not use environment applications column anymore

AminSlk:
Ref: SRX-XQ21LY

### DIFF
--- a/services/cd-service/pkg/service/git_test.go
+++ b/services/cd-service/pkg/service/git_test.go
@@ -275,7 +275,7 @@ func TestGetProductDB(t *testing.T) {
 			},
 			expectedErrors: []StageError{
 				{
-					err: errMatcher{"unable to get applications for environment 'staging': environment staging not found"},
+					err: nil,
 				},
 				{
 					err: nil,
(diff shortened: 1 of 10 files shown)


services/cd-service/pkg/service/git_test.go
```diff
@@ -442,43 +442,6 @@ func TestGetProductDBFailureCases(t *testing.T) {
 			timestamp:     &ts,
 			expectedErr:   errMatcher{"Can not have both an environment and environmentGroup to get the product summary for"},
 		},
-		{
-			Name:      "invalid environment used",
```
AminSlk, line 446: If an invalid environment is sent to GetProductSummary, we can just return an empty response. No need for returning error.

services/cd-service/pkg/service/overview.go
```diff
@@ -228,42 +228,28 @@ func (o *OverviewServiceServer) GetAppDetails(
 				continue
 			}
 
-			environment, ok := envMap[envName]
-			if !ok {
-				logger.FromContext(ctx).Sugar().Warnf("could not obtain environment %s for app %s: %w", envName, appName, err)
-				continue
-			}
-			foundApp := false // only apps that are active on that environment should be returned here
```
AminSlk, line 236: We're already checking that a deployment Release exists for that environment in the lines above. Therefore, this check is redundant.

PR #2389, opened 2025-05-28
fix: Run release trains with only one writing transaction

sven-urbanski-freiheit-com:
The goal here is to avoid that only *parts* of a release train are successful.

This introduce a new experimental helm parameter
"cd.experimentalParallelismOneTransaction".
When set to true the release trains to environment groups are changed:
* The go routines (1 per env) now only produce the prognosis (list of versions to be deployed).
* The go routines use a read only transaction.
* Only the requests go routine is applying the changes with a writing transaction.

Ref: SRX-JEU54H

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -2684,6 +2684,7 @@ func (c *ReleaseTrain) Transform(
 			type ChannelData struct {
 				prognosis *ReleaseTrainEnvironmentPrognosis
 				train     *envReleaseTrain
+				error     error
 			}
 			var prognosisResultChannel = make(chan *ChannelData, channelSize)
 			for _, envName := range envNames {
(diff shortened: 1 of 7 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -2675,12 +2677,60 @@ func (c *ReleaseTrain) Transform(
...
-			envNameLocal := envName
-			releaseTrainErrGroup.Go(func() error {
-				return c.runEnvReleaseTrainBackground(ctx, state, t, envNameLocal, trainGroup, envGroupConfigs, configs, allLatestReleases)
-			})
+		if state.ParallelismOneTransaction {
+			type ChannelData struct {
+				prognosis *ReleaseTrainEnvironmentPrognosis
+				train     *envReleaseTrain
+			}
+			var prognosisResultChannel = make(chan *ChannelData, 42)
```
AminSlk, line 2685: Why is 42 a fixed number here? We could just get len(envNames)

services/cd-service/pkg/repository/transformer.go
```diff
@@ -2675,12 +2677,60 @@ func (c *ReleaseTrain) Transform(
...
+					prognosisResultChannel <- &ChannelData{
+						prognosis: prognosis,
+						train:     train,
+					}
+					return nil
+				})
+			}
+			for i := 0; i < expectedNumPrognoses; i++ {
+				// waiting for prognoses to be done
+				var result = <-prognosisResultChannel
```
AminSlk, line 2719: Since this channel is buffered, the code here will not wait for all of the routines to finish. It just continues with any result that it gets. I think we can use something like 
```
releaseTrainErrGroup.Wait()
```
Before the for line.

sven-urbanski-freiheit-com: I solved it in a different way.
It's actually ok that it "continues with any result" because it loops over all results and the order does not matter.
I changed it now, so that the go routines never return an error, they just write the error to the channel.
This way, we do not have to wait for `.Wait()`, because it is already clear that the routines are done from the content of the channel.

AminSlk (approved):

PR #2445, opened 2025-06-26
chore(earthly): remove earthly from pkg

AminSlk:
Ref: SRX-XQY8JM

### DIFF
--- a/pkg/Makefile
+++ b/pkg/Makefile
@@ -25,8 +25,12 @@ MIN_COVERAGE?=47
 SERVICE_DIR:=/kp/pkg
 include $(ROOT_DIR)/infrastructure/make/go/include.mk
 
-all: lint test
+all: gen lint test
 test: gen unit-test
+
+.PHONY: lint
+lint: deps gen
+	docker run --rm -w $(SERVICE_DIR) -v ".:$(SERVICE_DIR)" $(DEPS_IMAGE) sh -c 'golangci-lint run --timeout=15m -j4 --tests=false --skip-files=".*\.pb\.go$$" ./... || $(SKIP_LINT_ERRORS) && exhaustruct -test=false $$(go list ./... | grep -v "github.com/freiheit-com/kuberpult/pkg/api" | grep -v "github.com/freiheit-com/kuberpult/pkg/publicapi")'
 
 gen: grpc-api-gen publicapi
 


infrastructure/docker/deps/Makefile
```diff
@@ -18,5 +18,7 @@ include ../../../Makefile.variables
 
 .PHONY: build
 build:
+ifeq ($(IMAGE_TAG),latest)
```
EdSwordsmith, line 21: We removed this because it would make the ci fail if the builder image wasn't built in that PR

infrastructure/make/go/include.mk
```diff
@@ -13,7 +13,7 @@ MIN_COVERAGE?=99.9 # should be overwritten by every service
 
 .PHONY: deps
 deps:
-	IMAGE_TAG=latest $(MAKE) -C $(ROOT_DIR)/infrastructure/docker/deps build
+	IMAGE_TAG=$(IMAGE_TAG) $(MAKE) -C $(ROOT_DIR)/infrastructure/docker/deps build
```
EdSwordsmith, line 16: Same here

pkg/Makefile
```diff
@@ -15,23 +15,17 @@
 # Copyright freiheit.com
 
 # NOTE: this will be replaced with the etf-golang makefile
-include ../Makefile.variables
+ROOT_DIR=..
 MAKEFLAGS += --no-builtin-rules
-
 export CGO_ENABLED=1
```
EdSwordsmith, line 20: This is the default value in the `include.mk`, maybe we can just add the export keyword there and remove this line from here

pkg/Makefile
```diff
@@ -52,5 +46,6 @@ proto: api/v1/api_grpc.pb.go publicapi
 
 build-pr: all
 build-main: all
+build-all: all
```
EdSwordsmith, line 49: This target isn't needed

pkg/Makefile
```diff
@@ -15,23 +15,17 @@
...
-lint:
-	earthly +lint
-test:
-	earthly -P +test "--GO_TEST_ARGS=$(GO_TEST_ARGS)"
+include $(ROOT_DIR)/infrastructure/make/go/include.mk
+
+all: lint test
+test: unit-test
 
 gen: grpc-api-gen publicapi
```
EdSwordsmith, line 30: We could use a target for generating these without needing `buf` installed, maybe using the deps docker image and copying the generated files from there

EdSwordsmith (changes requested):

pkg/Earthfile
```diff
@@ -1,35 +0,0 @@
...
-    FROM +deps
-    ARG GO_TEST_ARGS
-    WORKDIR pkg
-    WITH DOCKER --compose ../docker-compose-unittest.yml
-        RUN --no-cache \
-            echo "Started docker compose for unit tests." ;\
-            go test $GO_TEST_ARGS ./... -coverprofile=coverage.out
-    END
-    RUN go tool cover -html=coverage.out -o coverage.html
-    RUN ../infrastructure/coverage/check-coverage-go.sh coverage.out 47.8 pkg
```
sven-urbanski-freiheit-com, line 29: when you remove this, you need to add `MIN_COVERAGE=47.8` to the makefile.

AminSlk: It's already being done in the `include.mk` file.

EdSwordsmith (approved):

PR #2454, opened 2025-06-27
feat: use deployment frontloading in release trains

sven-urbanski-freiheit-com:
Ref: SRX-MQKPEG


services/cd-service/pkg/repository/transformer.go
```diff
@@ -3320,33 +3330,70 @@ func (c *envReleaseTrain) prognosis(ctx context.Context, state *State, transacti
...
+		if existingDeployment != nil && existingDeployment.Version != nil {
+			var tmp2 = (uint64)(*existingDeployment.Version)
+			existingVersion = &tmp2
+		}
+		var oldReleaseCommitId = ""
+		if existingVersion != nil {
+			oldReleaseCommitId, _ = getCommitID(ctx, transaction, state, *existingVersion, appName)
+			// continue anyway, this is only for events
+		}
+
```
sven-urbanski-freiheit-com, line 3376: This is the core of this PR.
We load a bunch of data here, now in the prognosis, not when applying.

PR #2462, opened 2025-07-01
chore: remove earthly from manifest repo service

EdSwordsmith:
Ref: SRX-DHOQ6P


Makefile
```diff
@@ -77,6 +77,7 @@ compose-down:
 
 kuberpult: compose-down
 	IMAGE_TAG=local make -C services/cd-service docker
+	IMAGE_TAG=local make -C services/manifest-repo-export-service docker
```
sven-urbanski-freiheit-com, line 80: please extract this into its own target, so you don't have to repeat the same lines 3 times.

PR #2496, opened 2025-07-09
fix(cd): change revision from string to uint64

miguel-crespo-fdc:
Ref: SRX-NAIP6Q


database/migrations/postgres/1751468673639069_revision.up.sql
```diff
@@ -1,34 +1,38 @@
...
+    ALTER TABLE releases_history DROP CONSTRAINT releases_pkey;
+    ALTER TABLE releases_history ADD PRIMARY KEY (version, releaseVersion, revision);
 END IF;
 END $$;
 
 DO $$
 BEGIN
     IF NOT EXISTS (SELECT 1 FROM information_schema.columns WHERE table_name = 'deployments' AND column_name='revision') THEN
-ALTER TABLE IF EXISTS deployments ADD COLUMN revision VARCHAR NOT NULL DEFAULT '0';
+    ALTER TABLE IF EXISTS deployments ADD COLUMN revision int NOT NULL DEFAULT 0;
```
miguel-crespo-fdc, line 22: Deployments do not need revision as part of the primary key as the release version was also not in the PK (appName, envName are enough)

PR #2497, opened 2025-07-09
fix: refactor getCommitDeploymentInfo

sven-urbanski-freiheit-com:
Ref: SRX-8CT290

### DIFF
--- a/services/cd-service/pkg/service/commit_deployment.go
+++ b/services/cd-service/pkg/service/commit_deployment.go
@@ -97,6 +97,7 @@ SELECT appname, envname, releaseVersion
 FROM deployments
 WHERE releaseVersion IS NOT NULL;
 `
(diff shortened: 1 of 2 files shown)


services/cd-service/pkg/service/commit_deployment.go
```diff
@@ -126,6 +89,56 @@ WHERE releaseVersion IS NOT NULL;
 	}, nil
 }
 
+func getDeploymentsWithoutReleaseVersion(ctx context.Context, transaction *sql.Tx, err error, applicationReleases map[string]map[types.EnvName]uint64) error {
```
miguel-crespo-fdc, line 92: Nitpick: is the name of this function correct ? seems like we are getting every Deployment WITH some release version

sven-urbanski-freiheit-com:
should be rebased on #2498 before merging

PR #2510, opened 2025-07-14
fix: datadog span connections and postgres visualization

sven-urbanski-freiheit-com:
Instead of marking every span "as DB",
which had the unintended side effect
of disconnecting traces of different services (http frontend -> grpc cd),
we now use the datadog-sql wrapper.

Ref: SRX-8CT290

This reverts #2494.

### DIFF
--- a/pkg/db/db_testutil.go
+++ b/pkg/db/db_testutil.go
@@ -69,6 +69,9 @@ func ConnectToPostgresContainer(ctx context.Context, t testing.TB, migrationsPat
 
 		MaxIdleConnections: 0,
 		MaxOpenConnections: 0,
+
+		DatadogEnabled:     false,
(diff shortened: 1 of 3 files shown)


pkg/db/db_environments.go
```diff
@@ -153,26 +152,25 @@ func (h *DBHandler) DBSelectEnvironmentsBatch(ctx context.Context, tx *sql.Tx, e
 }
 
 func (h *DBHandler) DBSelectAllEnvironments(ctx context.Context, transaction *sql.Tx) ([]types.EnvName, error) {
-	span, ctx, onErr := tracing.StartSpanFromContext(ctx, "DBSelectAllEnvironments")
```
miguel-crespo-fdc, line 156: Should this be reverted?

PR #2546, opened 2025-07-24
Mc/manifest xport revisions

miguel-crespo-fdc:


infrastructure/docker/builder/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Rebuild counter: 1
+# Rebuild counter: 2
```
sven-urbanski-freiheit-com, line 1: this shouldn't be necessary anymore after #2544

services/frontend-service/src/ui/utils/Links.tsx
```diff
@@ -152,12 +162,12 @@ export const DisplaySourceLink: React.FC<{ displayString: string; commitId: stri
...
 };
 
-export const DisplayManifestLink: React.FC<{ displayString: string; app: string; version: number }> = (
+export const DisplayManifestLink: React.FC<{ displayString: string; app: string; version: ReleaseNumbers }> = (
     props
 ): JSX.Element | null => {
     const { displayString, app, version } = props;
 
-    let manifestLink = '/ui/manifest?app=' + app + '&release=' + version;
+    let manifestLink = '/ui/manifest?app=' + app + '&release=' + version.version + '&revision=' + version.revision;
```
sven-urbanski-freiheit-com, line 170: i think you need to check here if revisionsEnabled==true, otherwise it would break for setups without revisions.

sven-urbanski-freiheit-com (approved):

PR #2550, opened 2025-07-25
feat: prepare for seperate deployment attempt history

bjoern-michaelsen-freiheit:


database/migrations/postgres/1753262713869421_separate_deployment_attempts_history.up.sql
```diff
@@ -0,0 +1,38 @@
...
+    revision VARCHAR
+);
+CREATE INDEX IF NOT EXISTS deployment_attempts_history_env_app ON deployment_attempts_history (envname, appname);
+
+INSERT INTO deployment_attempts_history (created, envname, appname, releaseversion, revision)
+SELECT created, envname, appname, queuedreleaseversion, revision
+FROM deployment_attempts
+ORDER BY created, eslVersion;
+
+CREATE TABLE IF NOT EXISTS deployment_attempts_latest
```
sven-urbanski-freiheit-com, line 17:
```suggestion
CREATE TABLE IF NOT EXISTS deployment_attempts
```
Adding "latest" would be against the standard we have now. It makes sense to add "latest" semantically, but I think consistency is more important.

sven-urbanski-freiheit-com (approved):

PR #2560, opened 2025-07-29
chore(CI): Move migrations from builder to service images

miguel-crespo-fdc:


infrastructure/make/go/include.mk
```diff
@@ -10,7 +10,7 @@ IMAGE_NAME?=$(DOCKER_REGISTRY_URI)/kuberpult-$(SERVICE):$(IMAGE_TAG)
 MAIN_IMAGE_NAME=$(DOCKER_REGISTRY_URI)/kuberpult-$(SERVICE):main
 SERVICE_DIR?=/kp/services/$(SERVICE)
 MIN_COVERAGE?=99.9 # should be overwritten by every service
-CONTEXT?=.
+CONTEXT?=../../
```
sven-urbanski-freiheit-com, line 13: I don't think it's required that we override this context, right?
In that case:

```suggestion
CONTEXT=../../
```

sven-urbanski-freiheit-com (approved):

PR #2570, opened 2025-07-31
fix(ci): Only run docker once

sven-urbanski-freiheit-com:
When running `make docker` we now only run one docker command (per service). This leads to faster build times.

Previously we did:
1) docker run with the builder image to compile
2) save the executable locally
3) actual docker build that copies the executable

Ref: SRX-O4HMIP


Makefile
```diff
@@ -71,7 +71,7 @@ cleanup-main:
 
 .PHONY: builder
 builder:
-	IMAGE_TAG=main make -C infrastructure/docker/builder build
+	IMAGE_TAG=local make -C infrastructure/docker/builder build
```
AminSlk, line 74: Why should this value always be local?

AminSlk (approved):

Makefile
sven-urbanski-freiheit-com: This target is only used locally, and at other places (e.g. a few lines below in `prepare-compose`) we expect the builder tag to be `local`.

PR #2579, opened 2025-08-04
feat(manifest-export): Delete AA environment configs

miguel-crespo-fdc:
Ref: SRX-YZQDJW


services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -2139,10 +2125,76 @@ func (c *DeleteAAEnvironmentConfig) GetEslVersion() db.TransformerID {
...
+		return "", fmt.Errorf("error opening file: %w", err)
+	}
+	var envConfig config.EnvironmentConfig
+
+	err = json.Unmarshal(data, &envConfig)
+	if err != nil {
+		return "", err
+	}
+
+	argoCdAppFile := fs.Join("argocd", string(argocd.V1Alpha1), fmt.Sprintf("%s.yaml", *envConfig.ArgoCdConfigs.CommonEnvPrefix+"-"+string(c.Environment)+"-"+string(c.ConcreteEnvironmentName)))
```
sven-urbanski-freiheit-com, line 2153: The same file is also written in "afterTransform". Is this line really necessary?
We shouldn't write the same thing in two different places.

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -2139,10 +2125,76 @@ func (c *DeleteAAEnvironmentConfig) GetEslVersion() db.TransformerID {
...
+
+	err = writeEnvironmentConfigurationToManifestRepo(fs, configFile, envConfig)
+	if err != nil {
+		return "", fmt.Errorf("error writing environment configuration to manifest repository: %w", err)
+	}
+	return fmt.Sprintf("removed configuration for AA environment %q - %q", c.Environment, c.ConcreteEnvironmentName), nil
+}
+
+func writeEnvironmentConfigurationToManifestRepo(fs billy.Filesystem, configFile string, envConfig config.EnvironmentConfig) error {
+	err := fs.Remove(configFile) // New config was just getting appended to the old configuration . Removing the file and recreating it resulted in the correct output in the file.
```
sven-urbanski-freiheit-com, line 2181: The comment could be clearer
```suggestion
	err := fs.Remove(configFile) // If we do not remove the file, we get an "append" not "overwrite". Removing the file and recreating it resulted in the correct output in the file.
```

services/manifest-repo-export-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: At least this line that is just putting together the path of the file, should be extracted into its own function. (So that it's a bit more obvious that other code is calling this too).

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -2139,10 +2125,81 @@ func (c *DeleteAAEnvironmentConfig) GetEslVersion() db.TransformerID {
...
+	}
+
+	err = writeEnvironmentConfigurationToManifestRepo(fs, configFile, envConfig)
+	if err != nil {
+		return "", fmt.Errorf("error writing environment configuration to manifest repository: %w", err)
+	}
+	return fmt.Sprintf("removed configuration for AA environment %q - %q", c.Environment, c.ConcreteEnvironmentName), nil
+}
+
+func getArgoCdAAEnvFileName(fs billy.Filesystem, commonEnvPrefix, parentEnvironmentName, concreteEnvironmentName types.EnvName) string {
```
sven-urbanski-freiheit-com, line 2180: this function should also be user in afterTransform

PR #2580, opened 2025-08-05
feat: Show process delays in UI

bjoern-michaelsen-freiheit:
Ref: SRX-3VPEV1


services/manifest-repo-export-service/pkg/service/git.go
```diff
@@ -246,6 +249,31 @@ func findCommitID(
...
+	if esl.Created.IsZero() {
+		return 0, 0
+	}
+	count, err := dbHandler.DBCountEslEventsNewer(ctx, transaction, esl.EslVersion)
+	now := time.Now().UTC()
+	diff := now.Sub(esl.Created).Seconds()
+	if err != nil {
+		return diff, 1
+	}
+	return diff, count
```
sven-urbanski-freiheit-com, line 274: you are using "count" even though there is an error, that seems wrong.

sven-urbanski-freiheit-com (approved):

services/manifest-repo-export-service/pkg/service/git.go
bjoern-michaelsen-freiheit: Huh? if err != nil then we return already in line 272

PR #2602, opened 2025-08-12
fix(db): optimize query to get all deployments

sven-urbanski-freiheit-com:
Ref: SRX-XCOQX6

### DIFF
--- a/services/manifest-repo-export-service/pkg/db_history/db_history.go
+++ b/services/manifest-repo-export-service/pkg/db_history/db_history.go
@@ -114,13 +114,5 @@ func processAllLatestDeploymentsForEnv(_ context.Context, rows *sql.Rows) (Deplo
 		}
 		result[types.AppName(curr.App)] = curr
 	}
-	err := rows.Close()
-	if err != nil {
-		return nil, fmt.Errorf("deployments: row closing error: %v", err)
-	}
-	err = rows.Err()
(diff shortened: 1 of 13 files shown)


services/manifest-repo-export-service/pkg/db_history/db_history.go
```diff
@@ -0,0 +1,126 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("error during json unmarshal in deployments. Error: %w. Data: %s", err, jsonMetadata)
+		}
+		if releaseVersion.Valid {
+			conv := uint64(releaseVersion.Int64)
+			curr.ReleaseNumbers.Version = &conv
+		}
+		result[types.AppName(curr.App)] = curr
+	}
+	err := rows.Close()
```
bjoern-michaelsen-freiheit, line 117: lets not do that here: some return path have the rows closed at the end, others have not. Lets not do that here (or do it here only with a defer so that we attempt to close the rows on all return paths).

bjoern-michaelsen-freiheit (approved):

services/manifest-repo-export-service/pkg/db_history/db_history.go
sven-urbanski-freiheit-com: true, we already close the rows in the caller, we don't need it here.

PR #2609, opened 2025-08-26
fix(helm): add virtualservice to helm chart options

AminSlk:
Ref: SRX-5B70L6

### DIFF
--- a/charts/kuberpult/templates/ingress.yaml
+++ b/charts/kuberpult/templates/ingress.yaml
@@ -234,7 +234,7 @@ spec:
 {{- end }}
 {{- end }}
 ---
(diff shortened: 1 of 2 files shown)


charts/kuberpult/templates/ingress.yaml
```diff
@@ -233,3 +233,72 @@ spec:
       secretName: {{ required ".ingress.iap.secretName is required" .Values.ingress.iap.secretName }}
 {{- end }}
 {{- end }}
+---
+{{- if .Values.ingress.virtualservice.create }}
```
sven-urbanski-freiheit-com, line 237:
```suggestion
{{- if .Values.ingress.experimentalVirtualservice.create }}
```

PR #2622, opened 2025-09-04
chore(charts): make experimentalParallelism not experimental

AminSlk:
Ref: SRX-OPH8E8


charts/kuberpult/values.yaml
```diff
@@ -117,7 +116,7 @@ cd:
 # Disabling the queue is as of now an experimental feature.
 # With the queue, the cd-service processes only one request at a time, which is very much required when using git.
 # With the database, this is not required anymore.
-    disableQueue: false
+    disableQueue: true
```
sven-urbanski-freiheit-com, line 119: I think we should remove this as well, rather than setting it to true. (but it must be done in a non-breaking-change way).

PR #2623, opened 2025-09-08
fix: allow direct deployment of non-latest

bjoern-michaelsen-freiheit:
Ref: SRX-COS61V

### DIFF
--- a/services/cd-service/pkg/repository/transformer.go
+++ b/services/cd-service/pkg/repository/transformer.go
@@ -637,7 +637,9 @@ func (c *CreateApplicationVersion) Transform(
 			return "", err
 		}
 		t.AddAppEnv(c.Application, env, teamOwner)
-		if ((hasUpstream && config.Upstream.Latest && isLatest) || slices.Contains(c.DeployToDownstreamEnvironments, env)) && !c.IsPrepublish {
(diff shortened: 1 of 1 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -637,7 +637,7 @@ func (c *CreateApplicationVersion) Transform(
 			return "", err
 		}
 		t.AddAppEnv(c.Application, env, teamOwner)
-		if ((hasUpstream && config.Upstream.Latest) || slices.Contains(c.DeployToDownstreamEnvironments, env)) && isLatest && !c.IsPrepublish {
+		if ((hasUpstream && config.Upstream.Latest && isLatest) || slices.Contains(c.DeployToDownstreamEnvironments, env)) && !c.IsPrepublish {
```
sven-urbanski-freiheit-com, line 640: I think we can extract some variables here to make this more readable, something like this:
```suggestion
        envIsConfiguredLatest:=hasUpstream && config.Upstream.Latest && isLatest
        downstreamDeploymentRequested := slices.Contains(c.DeployToDownstreamEnvironments, env)
		if (envIsConfiguredLatest || downstreamDeploymentRequested) && !c.IsPrepublish {
```

sven-urbanski-freiheit-com (approved):

PR #2637, opened 2025-09-24
fix(chore): unify row closing (part 1)

ghost:


pkg/db/db.go
```diff
@@ -256,6 +256,13 @@ func closeRows(rows *sql.Rows) error {
 	return nil
 }
 
+func closeRowsAndLog(rows *sql.Rows, ctx context.Context, prefix string) {
+	err := rows.Close()
+	if err != nil {
```
sven-urbanski-freiheit-com, line 261: some of the replaced code also did a null check:
```suggestion
    if rows != nil {
        return
    }
	err := rows.Close()
	if err != nil {
```

sven-urbanski-freiheit-com (approved):

PR #2652, opened 2025-10-21
fix: date issue in the history page

bqcuong:
Ref: SRX-ON2WF2


bjoern-michaelsen-freiheit (approved):

services/frontend-service/src/ui/components/FormattedDate/FormattedDate.tsx
```diff
@@ -87,3 +92,23 @@ export const FormattedDate: React.FC<{ createdAt: Date; className?: string }> =
         </span>
     );
 };
+
+export const FormattedExactDate: React.FC<{ createdAt: Date; className?: string; prefixText?: string }> = ({
+    createdAt,
+    className,
+    prefixText,
+}) => {
```
sven-urbanski-freiheit-com: Another component just for the prefixText seems a bit much.
Please make `FormattedDate` use `FormattedExactDate` to avoid redundancy

bqcuong: Not only for the prefixText. The content of main text and tooltip text are different in each case of `FormattedDate` and `FormattedExactDate`

PR #2656, opened 2025-10-23
 fix: Ensure deleting an unknown environment returns 404

bqcuong:
Ref: SRX-XQP9BR


services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: please change the commit message (PR title) to something like
`fix: Ensure deleting an unknown environment returns 404`

bjoern-michaelsen-freiheit (approved):
approved with sven  comment

services/cd-service/pkg/repository/transformer.go
bqcuong: updated!

PR #2658, opened 2025-10-23
fix: Speed up removing env from app

bqcuong:
Ref: SRX-OW5TMV

Updated UI: `development` had a deployment before, and has been currently removed from `sreteam`. `aa-test` never has a manifest.
image

### DIFF
--- a/services/cd-service/pkg/repository/transformer_db_test.go
+++ b/services/cd-service/pkg/repository/transformer_db_test.go
@@ -2561,7 +2561,10 @@ func TestDeleteEnvFromAppWithDB(t *testing.T) {
 		&CreateEnvironment{
 			Environment: "env",
 			Config: config.EnvironmentConfig{
-				Upstream:         nil,
+				Upstream: &config.EnvironmentConfigUpstream{
+					Environment: "",
+					Latest:      true,
+				},
 				ArgoCd:           nil,
 				EnvironmentGroup: conversion.FromString("mygroup"),
 			},
@@ -2719,6 +2722,7 @@ func TestDeleteEnvFromAppWithDB(t *testing.T) {
 					"env2": "testenvmanifest2",
 				},
 				10: {
+					"env":  "testenvmanifest",
 					"env2": "testenvmanifest2",
 				},
 			},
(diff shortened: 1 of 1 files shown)


services/cd-service/pkg/repository/transformer.go
```diff
@@ -1229,7 +1230,19 @@ func (u *DeleteEnvFromApp) Transform(
...
+
+	// Find the "oldest" one among the latest deployments on all environments
+	oldestDeployment, err := state.DBHandler.DBSelectOldestDeploymentForApplication(ctx, transaction, u.Application)
+	if err != nil {
+		return "", err
+	}
+
+	// Only update the recent releases (counting from the oldest release having a deployment)
 	for _, dbReleaseWithMetadata := range releases {
+		if oldestDeployment != nil && !types.GreaterOrEqual(dbReleaseWithMetadata.ReleaseNumbers, oldestDeployment.ReleaseNumbers) {
```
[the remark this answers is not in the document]
bqcuong: Fixed!

services/cd-service/pkg/repository/transformer_db_test.go
```diff
@@ -2701,13 +2810,35 @@ func TestDeleteEnvFromAppWithDB(t *testing.T) {
...
+					for releaseNumber, expectedManifests := range tc.ExpectedManifestsByEachRelease {
+						var release *db.DBReleaseWithMetaData
+						for _, rel := range releases {
+							if *rel.ReleaseNumbers.Version == releaseNumber {
+								release = rel
+								break
+							}
+						}
+						if release == nil {
+							return fmt.Errorf("expect release but not found: %v", err2)
```
sven-urbanski-freiheit-com: err2 is always nil here, please add something else to the returned error, like  the app name

services/cd-service/pkg/repository/transformer_db_test.go
```diff
@@ -2701,13 +2810,35 @@ func TestDeleteEnvFromAppWithDB(t *testing.T) {
 				if err2 != nil {
 					return fmt.Errorf("error retrieving release: %v", err2)
 				}
-				for _, release := range releases {
-					for env, manifest := range tc.ExpectedManifests {
-						if diff := cmp.Diff(manifest, release.Manifests.Manifests[env]); diff != "" {
-							return fmt.Errorf("error mismatch Manifests - want, +got:\n%s", diff)
+
+				if tc.ExpectedManifestsByEachRelease != nil {
```
sven-urbanski-freiheit-com: This if statement is not clear to me.
Why do we sometimes test all releases, and sometimes only those where we specified the release version?

I would like to see 2 changes:
1) remove the field "expectedManifests" entirely and adapt all test cases (in this function) to use "ExpectedManifestsByEachRelease"
2) and ensure that all releases returned from DBSelectReleasesByAppLatestEslVersion are specified. If they testdata does not have all releases, return an error.

bqcuong: fixed!

PR #2664, opened 2025-10-30
fix: use custom type for application names

bqcuong:
Ref: SRX-LP12J1


sven-urbanski-freiheit-com (approved):
let's get this merged!

PR #2669, opened 2025-11-05
fix: sync basket

ghost:
When a user opens the "Delete Environments" dialog for an application, any environments they have already selected for deletion (which are stored as pending batch actions) will now be checked by default. This improves the user experience by making the dialog's initial state consistent with the current batch action list. Just keeping the state from the last open dialog would not be sufficient as batch actions can be _removed_ from the basket when the dialog is closed.

Ref: SRX-GJ3E9Y

### DIFF
--- a/services/frontend-service/src/ui/utils/store.tsx
+++ b/services/frontend-service/src/ui/utils/store.tsx
@@ -637,7 +637,7 @@ export const deleteAllActions = (): void => {
     UpdateAction.set({ actions: [] });
 };
 
-export const useDeleteEnvironmentsForAppActions = (app: string): string[] => {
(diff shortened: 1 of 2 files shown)


services/frontend-service/src/ui/utils/store.tsx
```diff
@@ -637,6 +637,20 @@ export const deleteAllActions = (): void => {
     UpdateAction.set({ actions: [] });
 };
 
+export const useDeleteEnvironmentsForAppActions = (app: string): string[] => {
```
sven-urbanski-freiheit-com: nitpick: maybe rename this
```suggestion
export const useDeleteEnvironmentActionsForApp = (app: string): string[] => {
```

PR #2689, opened 2025-11-18
fix: reduce number of DBSelectAllAppLocks calls

bqcuong:
Ref: SRX-9LQHKD


services/cd-service/pkg/repository/repository.go
```diff
@@ -719,6 +719,36 @@ func (s *State) GetEnvironmentLocksFromDB(ctx context.Context, transaction *sql.
 	return result, nil
 }
 
+func (s *State) GetApplicationLocksForEnv(ctx context.Context, transaction *sql.Tx, environment types.EnvName) (map[types.AppName]map[string]Lock, error) {
+	if transaction == nil {
+		return nil, fmt.Errorf("GetApplicationLocksForEnvFromDB: No transaction provided")
+	}
```
sven-urbanski-freiheit-com: this check is unnecesarry, we can't really handled that.
```suggestion
```

services/cd-service/pkg/repository/repository.go
```diff
@@ -719,6 +719,36 @@ func (s *State) GetEnvironmentLocksFromDB(ctx context.Context, transaction *sql.
 	return result, nil
 }
 
+func (s *State) GetApplicationLocksForEnv(ctx context.Context, transaction *sql.Tx, environment types.EnvName) (map[types.AppName]map[string]Lock, error) {
```
sven-urbanski-freiheit-com: Please add a span for this function, e.g.:
```suggestion
func (s *State) GetApplicationLocksForEnv(ctx context.Context, transaction *sql.Tx, environment types.EnvName) (map[types.AppName]map[string]Lock, error) {
    span, ctx := StartSpan...
    defer span.Finish(error)
```

PR #2859, opened 2026-04-09
fix: add client uuid to http header for better debugging

bqcuong:
Ref: SRX-X41I6Z

### DIFF
--- a/pkg/ctxkeys/keys.go
+++ b/pkg/ctxkeys/keys.go
@@ -1,0 +1,7 @@
+package ctxkeys
+
+type ctxClientUUID struct{}
+
+var (
+	CtxClientUUIDKey = &ctxClientUUID{}
+)
--- a/pkg/ctxkeys/keys.go
+++ b/pkg/ctxkeys/keys.go
@@ -1,3 +1,19 @@
+/*This file is part of kuberpult.
+
+Kuberpult is free software: you can redistribute it and/or modify
+it under the terms of the Expat(MIT) License as published by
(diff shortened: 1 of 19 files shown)


pkg/ctxkeys/keys.go
```diff
@@ -0,0 +1,23 @@
...
+
+Copyright freiheit.com*/
+
+package ctxkeys
+
+type ctxClientUUID struct{}
+
+var (
+	CtxClientUUIDKey = &ctxClientUUID{}
+)
```
ccreinartz11, line 23: ❓ Why do we need a new pkg for this header?

bqcuong: Simple answer: To avoid the cycling imports.

Before, these context headers are defined in different packages, e.g., `auth` package: https://github.com/freiheit-com/kuberpult/blob/main/pkg/auth/auth.go#L36

And we also use a context header in the `logger` package: https://github.com/freiheit-com/kuberpult/pull/2859/changes#diff-c1dc2eb1e149a1dc7ecd5cb474e167fb456eb7ad0ea0256cf62fba693a89af79R58

And most of other packages are using `logger` package, which introduces import cycles.
I think storing these context keys in a standalone package would be optimal solution.

ccreinartz11: agreed. can you create a story about moving all context keys that are currently used into the new pkg that you introduced?

ccreinartz11 (approved):

PR #2862, opened 2026-04-13
fix: bracket rendering into bracket directory in manifest-repo-service

sven-urbanski-freiheit-com:
Bracket rendering is intended as a performance optimization for Argo CD.
You can group any of your apps into "brackets", by providing the bracket parameter
in the `/api/release` endpoint. This reduces the number of apps that Argo needs to handle.
    
Kuberpult can now render the manifests into another directory
`environments/<env>/brackets/<bracket>/<app>.yaml`
if `manifestRepoExport.rendering.experimentalRenderBrackets:true`.
    
To point the root-apps (in `argocd/v1alpha/<env>.yaml`) to the brackets
set `manifestRepoExport.rendering.experimentalRootAppsPointToBrackets:true`.

Ref: SRX-ISPEFN


pkg/db/db_apps.go
```diff
@@ -485,7 +485,7 @@ func (h *DBHandler) processAppsRow(ctx context.Context, rows *sql.Rows, err erro
 	return row, nil
 }
 
-func (h *DBHandler) processOneAppInternal(rows *sql.Rows) (*DBAppWithMetaData, error) {
+func (h *DBHandler) processOneAppInternal(_ context.Context, rows *sql.Rows) (*DBAppWithMetaData, error) {
```
bqcuong: it seems `context.Context` is not used in this function

pkg/db/db_brackets_history.go
```diff
@@ -166,19 +166,27 @@ func DBSelectBracketHistoryLatest(ctx context.Context, h *DBHandler, tx *sql.Tx)
...
 	}
+	var result *BracketRow
 	if rows.Next() {
 		result, err = processBracketHistoryRow(rows)
 		if err != nil {
 			err2 := closeRows(rows)
 			return nil, errors.Join(err, err2)
 		}
+	} else {
+		return nil, nil
```
bqcuong, line 189: if no bracket is found, should we return an error instead of `nil`?
```
return nil, fmt.Errorf("no bracket found...")
```

charts/kuberpult/templates/manifest-repo-export-service.yaml
```diff
@@ -14,6 +14,19 @@
...
+{{- if (and (not (.Values.manifestRepoExport.rendering.experimentalRenderApps)) (not (.Values.manifestRepoExport.rendering.experimentalRootAppsPointToBrackets))) }}
+{{ fail "Cannot point to apps when not rendering apps: Values.manifestRepoExport.rendering.experimentalRootAppsPointToBrackets=false requires experimentalRenderApps to be true."}}
+{{ end -}}
+
+{{- if (and (not (.Values.manifestRepoExport.rendering.experimentalRenderBrackets)) (.Values.manifestRepoExport.rendering.experimentalRootAppsPointToBrackets)) }}
+{{ fail "Cannot point to brackets when not rendering brackets: Values.manifestRepoExport.rendering.experimentalRootAppsPointToBrackets=true requires experimentalRenderBrackets to be true."}}
+{{ end -}}
+
+{{- if (and (.Values.manifestRepoExport.rendering.experimentalRenderBrackets) (not (.Values.git.minimizeExportedData))) }}
+{{ fail "Cannot render brackets with minimizeExportedDat=false"}}
```
bqcuong:
```suggestion
{{ fail "Cannot render brackets with minimizeExportedData=false"}}
```

charts/kuberpult/tests/charts_test.go
```diff
@@ -2170,10 +2181,52 @@ rollout:
...
+  url: "testURL"
+  minimizeExportedData: false
+manifestRepoExport:
+  enabled: true
+  rendering:
+    experimentalRenderApps: true
+    experimentalRenderBrackets: true
+    experimentalRootAppsPointToBrackets: false
+`,
+			ExpectedError: infixErrMatcher{"Cannot render brackets with minimizeExportedDat=false"},
```
bqcuong:
```suggestion
			ExpectedError: infixErrMatcher{"Cannot render brackets with minimizeExportedData=false"},
```

services/manifest-repo-export-service/pkg/argocd/directories.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/freiheit-com/kuberpult/pkg/types"
+)
+
+type BracketDirectoryNames struct {
+	BracketDirectory string // just the directory
+	BracketPath      string // complete directory and filename
+}
+
+// BracketPaths returns path related to bracket rendering
+func BracketPaths(env types.EnvName, bracket types.ArgoBracketName, appName types.AppName) *BracketDirectoryNames {
```
bqcuong: Should we create a new file for the ArgoCD bracket path?
We have similar paths defined already in `transformer.go`, such as:
```
// environmentApplicationDirectory returns environments/<env>/applications/<app>/
func environmentApplicationDirectory(fs billy.Filesystem, environment types.EnvName, application string) string {
	return fs.Join("environments", string(environment), "applications", application)
}
```

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -465,6 +476,75 @@ func (c *DeployApplicationVersion) Transform(
...
+	}
+	return "", nil
+}
+
+func (c *DeployApplicationVersion) CleanupBracketFile(ctx context.Context, state *State, transaction *sql.Tx, fsys billy.Filesystem) error {
+	previousBracketHistory, err := db.DBSelectBracketHistoryLatestBeforeId(ctx, state.DBHandler, transaction, c.GetEslVersion())
+	if err != nil {
+		return err
+	}
+	if previousBracketHistory == nil {
```
bqcuong, line 519: could we simplify the code like this?
```
if previousBracketHistory != nil {
   logging.Warn(ctx, "bracket history found",
   ...
}
```

docker-compose.yml
```diff
@@ -58,7 +58,7 @@ services:
   manifest-repo-export-service:
     image: europe-west3-docker.pkg.dev/fdc-public-docker-registry/kuberpult/kuberpult-manifest-repo-export-service:local
     environment:
-      - LOG_LEVEL=INFO
+      - LOG_LEVEL=warn
```
bqcuong: Just curious question: why do we want warning level specifically for manifest-repo-export-service while running Kuberpult locally?
For cd-service and frontend-service, they are still with info level.

bqcuong (approved):

pkg/db/db_brackets_history.go
sven-urbanski-freiheit-com: If me made this case into an error, then the caller would need to check what kind of error it is. Having no brackets is valid, at least when turning on the brackets feature for the first time.
We can discuss if we can find a better general approach to this, but for now many of our db functions handle it like this.

services/manifest-repo-export-service/pkg/argocd/directories.go
sven-urbanski-freiheit-com: I move it to transformer.go
Edit: This does not work because of a cyclic dependency.

docker-compose.yml
sven-urbanski-freiheit-com: I set it back to info for now

PR #2863, opened 2026-04-13
fix: add makefile check

sven-urbanski-freiheit-com:
in order to verify that IMAGE_TAG_KUBERPULT is set when running helm tests

Ref: SRX-ISPEFN

### DIFF
--- a/charts/kuberpult/Makefile
+++ b/charts/kuberpult/Makefile
@@ -57,7 +57,10 @@ else
 		sh -c "ct lint --debug --chart-yaml-schema=ci/chart_schema.yaml --lint-conf=ci/lintconf.yaml --chart-dirs=. --charts=. --remote=git@github.com:freiheit-com/kuberpult.git"
 endif
 
-test-helm: $(TGZ_FILE) Chart.yaml
+check-precondition:
+	@if [ -z "$(IMAGE_TAG_KUBERPULT)" ]; then echo "IMAGE_TAG_KUBERPULT not set, exiting"; exit 1; fi
+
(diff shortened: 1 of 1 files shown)


charts/kuberpult/Makefile
```diff
@@ -57,7 +57,10 @@ else
 		sh -c "ct lint --debug --chart-yaml-schema=ci/chart_schema.yaml --lint-conf=ci/lintconf.yaml --chart-dirs=. --charts=. --remote=git@github.com:freiheit-com/kuberpult.git"
 endif
 
-test-helm: $(TGZ_FILE) Chart.yaml
+check-precondition:
+	@if [ -z "$(IMAGE_TAG_KUBERPULT)" ]; then echo "IMAGE_TAG_KUBERPULT not set, exiting"; exit 1; fi
```
bqcuong, line 61: Why and when would the `IMAGE_TAG_KUBERPULT` be empty? Can we fix the root cause of this issue?
If this happens in the CI builds, we do not want empty `IMAGE_TAG_KUBERPULT`, otherwise, the builds will be failing, right?

sven-urbanski-freiheit-com: The CI sets it, and it works there.
I added this check so that we get a quicker and clearer error when running locally. Locally we have to set the tag manually.

bqcuong: For local, is it possible to leverage the existing `make version` target to  get `IMAGE_TAG_KUBERPULT`? Manual configuration seems a bit unnecessarily complex.

bqcuong (approved):

PR #2871, opened 2026-04-23
fix: run releasetrain for specific git tag for obsolete releases

bqcuong:
Ref: [SRX-XNE3NZ](https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/STORY/-OqjFN2FgJHSls1FAWi7/)

image

### DIFF
--- a/pkg/db/db_apps.go
+++ b/pkg/db/db_apps.go
@@ -74,6 +74,26 @@ func (h *DBHandler) DBSelectApp(ctx context.Context, tx *sql.Tx, appName types.A
 	return h.processAppsRow(ctx, rows, err)
 }
 
+func (h *DBHandler) DBSelectAllAppsMetadataAtTimestamp(ctx context.Context, tx *sql.Tx, ts time.Time) (_ map[types.AppName]*DBAppWithMetaData, err error) {
+	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectAllAppsMetadataAtTimestamp")
+	defer func() {
+		span.Finish(tracer.WithError(err))
+	}()
+	selectQuery := h.AdaptQuery(`
+	SELECT DISTINCT
+		ON (appname) appname,
+		stateChange,
+		metadata,
+		argoBracket
+	FROM apps_history
+	WHERE
+		created <= ?;
+	`)
+	span.SetTag("query", selectQuery)
+	rows, err := tx.QueryContext(ctx, selectQuery, ts)
+	return h.processAppsRows(ctx, rows, err)
+}
+
 func (h *DBHandler) DBSelectAllAppsMetadata(ctx context.Context, tx *sql.Tx) (_ map[types.AppName]*DBAppWithMetaData, err error) {
 	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectAllAppsMetadata")
 	defer func() {
--- a/pkg/db/db_apps.go
+++ b/pkg/db/db_apps.go
@@ -87,7 +87,8 @@ func (h *DBHandler) DBSelectAllAppsMetadataAtTimestamp(ctx context.Context, tx *
 		argoBracket
 	FROM apps_history
 	WHERE
-		created <= ?;
+		created <= ?
+	ORDER BY appname, created DESC;
 	`)
 	span.SetTag("query", selectQuery)
 	rows, err := tx.QueryContext(ctx, selectQuery, ts)
--- a/pkg/db/db_releases.go
+++ b/pkg/db/db_releases.go
@@ -172,6 +172,38 @@ func (h *DBHandler) DBSelectReleaseByVersionAtTimestamp(ctx context.Context, tx 
 
 type AppVersionEnvironments map[types.AppName]map[string][]types.EnvName // first key is the appName
 
+func (h *DBHandler) DBSelectAllEnvironmentsForAllReleasesAtTimestamp(ctx context.Context, tx *sql.Tx, ts time.Time) (_ AppVersionEnvironments, err error) {
+	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectAllEnvironmentsForAllReleasesAtTimestamp")
+	defer func() {
+		span.Finish(tracer.WithError(err))
+	}()
+	query := h.AdaptQuery(`
+	SELECT DISTINCT appname
+	FROM (
+		SELECT DISTINCT ON (appname, releaseversion, revision)
+			appname,
+			releaseVersion,
+			environments,
+			revision,
+			deleted
+		FROM releases_history
+		WHERE created <= ?
+		ORDER BY
+			appname,
+			releaseversion,
+			revision,
+			version DESC
+	) AS latest_releases
+	WHERE deleted = false;
+	`)
+	rows, err := tx.QueryContext(ctx, query, ts)
+	if err != nil {
+		return nil, fmt.Errorf("could not query apps with releases at timestamp: %w", err)
+	}
+
+	return h.processReleaseEnvironmentRows(ctx, err, rows)
+}
+
 func (h *DBHandler) DBSelectAllEnvironmentsForAllReleases(ctx context.Context, tx *sql.Tx) (_ AppVersionEnvironments, err error) {
 	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectAllManifestsForAllReleases")
 	defer func() {
--- a/pkg/db/db_releases.go
+++ b/pkg/db/db_releases.go
@@ -970,6 +970,90 @@ func (h *DBHandler) DBSelectCommitHashesTimeWindow(ctx context.Context, transact
 	return releases, nil
 }
 
+func (h *DBHandler) DBSelectCommitIdAppReleaseVersionsAtTimestamp(ctx context.Context, transaction *sql.Tx, versionByApp map[types.AppName]types.ReleaseNumbers, ts time.Time) (_ map[types.AppName]string, err error) {
+	span, ctx := tracer.StartSpanFromContext(ctx, "DBSelectCommitIdAppReleaseVersions")
+	defer func() {
+		span.Finish(tracer.WithError(err))
+	}()
+	result := make(map[types.AppName]string)
+	if len(versionByApp) < 1 {
+		return result, nil
+	}
+	queryID := rand.IntN(1 << 31) // this function should be called no more than once per transaction, but just to be save ...
+	tableQuery := h.AdaptQuery(`CREATE TEMP TABLE IF NOT EXISTS temp_query_app_releaseversions(queryId INTEGER, appName VARCHAR NOT NULL, releaseVersion INTEGER, revision INTEGER);`)
+	_, err = transaction.ExecContext(ctx, tableQuery)
+	if err != nil {
+		return nil, fmt.Errorf("could not create query app releases table. Error: %w", err)
+	}
+	insertQuery := h.AdaptQuery(`INSERT INTO temp_query_app_releaseversions VALUES (?, ?, ?, ?)` + strings.Repeat(`, (?, ?, ?, ?)`, len(versionByApp)-1) + `;`)
+	args := make([]interface{}, len(versionByApp)*4)
+	i := 0
+	for appName, releaseVersion := range versionByApp {
+		args[i] = queryID
+		i++
+		args[i] = appName
+		i++
+		args[i] = releaseVersion.Version
+		i++
+		args[i] = releaseVersion.Revision
+		i++
+	}
+	_, err = transaction.ExecContext(
+		ctx,
+		insertQuery,
+		args...,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("could not insert into query app releases table. Query: %v, Error: %w", insertQuery, err)
+	}
+	selectQuery := h.AdaptQuery(`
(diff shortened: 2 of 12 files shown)


services/frontend-service/src/ui/components/ProductVersion/ProductVersion.tsx
```diff
@@ -316,7 +316,7 @@ export const ProductVersion: React.FC = () => {
                         </select>
                         <select className="env_drop_down" onChange={changeEnv} value={environment}>
                             <option value="default" disabled>
-                                Select an Environment or Environment Group
```
bqcuong, line 319: As discussed, it makes no sense to run release trains for the whole env group in this page, as the upstream envs might have different versions. So that we don't know we should take which version for running the release train. Better simplify this with env selection only.

pkg/db/db_releases.go
```diff
@@ -168,72 +167,46 @@ func (h *DBHandler) DBSelectReleaseByVersionAtTimestamp(ctx context.Context, tx
...
+			appname,
+			releaseversion,
+			revision
+		) appname,
+		releaseVersion,
+		environments,
+		revision
+	FROM releases_history
+	WHERE
+		deleted = false
```
sven-urbanski-freiheit-com, line 192: is it correct to remove deleted columns here?

bqcuong: yes, it is. as we're looking for releases when they are created.

services/cd-service/pkg/repository/releasetrain.go
```diff
@@ -556,7 +559,8 @@ func (c *envReleaseTrain) prognosis(ctx context.Context, state *State, transacti
 
 	appsPrognoses := make(map[types.AppName]ReleaseTrainApplicationPrognosis)
 
-	allLatestDeploymentsTargetEnv, err := state.DBHandler.DBSelectAllLatestDeploymentsOnEnvironment(ctx, transaction, envName)
+	// deployment versions in target env will be compared to the ones in upstream env
```
sven-urbanski-freiheit-com, line 577:
```suggestion
	// deployment versions in target env will be compared to the ones in upstream env
	// we need the current deployments, that is why `commitHash=""`
```

The empty commit hash deserve a comment

services/cd-service/pkg/repository/releasetrain.go
```diff
@@ -1022,6 +1058,57 @@ func (c *envReleaseTrain) applyPrognosis(
...
+			deletedRelease := appPrognosis.DeletedRelease
+
+			envConfigs, err := state.GetAllEnvironmentConfigs(ctx, transaction)
+			if err != nil {
+				return "", fmt.Errorf("could not get env configs: %w", err)
+			}
+
+			manifests := make(map[types.EnvName]string)
+			for envName, manifest := range deletedRelease.Manifests.Manifests {
+				// Skip if not deployed to this env. We'll let the
```
sven-urbanski-freiheit-com: fix the comment please

services/cd-service/pkg/repository/releasetrain.go
```diff
@@ -1022,6 +1058,57 @@ func (c *envReleaseTrain) applyPrognosis(
...
+				envConfig, ok := envConfigs[envName]
+				if !ok { // the env does not exist anymore
```
sven-urbanski-freiheit-com: this deserve a log message, e.g.
"tried to revive release, but skipped environment because it does not exist anymore"

services/cd-service/pkg/repository/releasetrain.go
```diff
@@ -1022,6 +1058,57 @@ func (c *envReleaseTrain) applyPrognosis(
...
+					continue
+				}
+
+				if envConfig.Upstream.Latest {
+					// the app will be deployed automatically to upstream envs with CreateApplicationVersion transformer.
+					// As we only want to deploy to the target env, upstream envs should be skipped
+					continue
```
sven-urbanski-freiheit-com: taking the upstream.latest flag as indicator to see, if you want to keep the env, seems wrong.
Yes, upstream envs are usually not targets for release trains - but they can be.

Why not just compare the name of the environment? That should be enough to decide if you want to keep it or not.

services/cd-service/pkg/repository/repository.go
```diff
@@ -675,14 +675,62 @@ func (s *State) DeleteQueuedVersion(ctx context.Context, transaction *sql.Tx, en
...
+			return nil, err
+		}
+		return s.DBHandler.DBSelectCommitIdAppReleaseVersionsAtTimestamp(ctx, transaction, appReleaseVersions, *ts)
+	}
+	return s.DBHandler.DBSelectCommitIdAppReleaseVersions(ctx, transaction, appReleaseVersions)
+}
+
+func (s *State) GetAllLatestReleases(ctx context.Context, transaction *sql.Tx, commitHash string) (map[types.AppName][]types.ReleaseNumbers, error) {
+	if commitHash != "" {
+		ts, err := s.GetCommitHashTimestamp(ctx, transaction, commitHash)
```
sven-urbanski-freiheit-com: you are introducing multiple places in the code that call `GetCommitHashTimestamp`, that's not efficient.
Please call it once in the prognosis or train, and then use the result here.

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -5776,6 +5776,177 @@ func TestReleaseTrainsWithCommitHash(t *testing.T) {
...
+						WriteCommitData: true,
+						Version:         uint64(versionTwo),
+					},
+					&UndeployApplication{
+						Application: appName,
+					},
+				},
+			},
+			// when we run a release train to production (commitHash refs to Stage 0)
+			// the app should be revived on the production (but not on upstream envs, i.e., staging)
```
sven-urbanski-freiheit-com: it should be revived on the target envs, independent of the upstream config

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -5776,6 +5776,177 @@ func TestReleaseTrainsWithCommitHash(t *testing.T) {
 				},
 			},
 		},
+		{
+			Name: "A release train with commit hash should revive an undeployed app",
```
sven-urbanski-freiheit-com, line 5780: I'm missing a test case about a release train that revives a deleted environment from an app.

services/cd-service/pkg/repository/releasetrain.go
bqcuong: This is implemented based on your suggestion: `Either way, once the release train to staging is done, it also has to work for the next train from staging to prod.`

We have to keep the manifests for both `staging` and `prod` but not the `dev`. We decided to exclude only the upstream envs with `latest=true` (i.e., `dev`) from the release's manifests.

Just a curious question: How can we run a release train to an "upstream.latest=true" env? I mean from where?

services/cd-service/pkg/repository/releasetrain.go
```diff
@@ -1022,6 +1058,57 @@ func (c *envReleaseTrain) applyPrognosis(
...
+				WriteCommitData:       c.Parent.WriteCommitData,
+				CiLink:                c.Parent.CiLink,
+				AllowedDomains:        c.Parent.AllowedDomains,
+				TransformerEslVersion: c.TransformerEslVersion,
+				IsPrepublish:          deletedRelease.Metadata.IsPrepublish,
+				ArgoBracket:           appPrognosis.ArgoBracket,
+			}
+			_, err = newRelease.Transform(ctx, state, t, transaction)
+			if err != nil {
+				return "", fmt.Errorf("could not create release for %s: %w", appName, err)
```
sven-urbanski-freiheit-com:
```suggestion
				return "", fmt.Errorf("could not revive release %v for app %s: %w", deletedRelease.ReleaseNumbers, appName, err)
```

services/cd-service/pkg/repository/repository.go
bqcuong: updated

services/cd-service/pkg/repository/transformer_test.go
bqcuong: added

services/cd-service/pkg/repository/releasetrain.go
bqcuong: fixed

services/cd-service/pkg/repository/releasetrain.go
```diff
@@ -882,6 +899,40 @@ func (c *envReleaseTrain) prognosis(ctx context.Context, state *State, transacti
...
+		var revivedRelease *db.DBReleaseWithMetaData
+		var argoBracket types.ArgoBracketName
+		if ts != nil {
+			// check if the release still exists for the app
+			release, err := state.DBHandler.DBSelectReleaseByVersion(ctx, transaction, appName, versionToDeploy, false)
+			if err != nil {
+				return failedPrognosis(err)
+			}
+
+			var recreateRelease = false
```
sven-urbanski-freiheit-com:
```suggestion
			var reviveRelease = false
```

services/cd-service/pkg/service/batch.go
```diff
@@ -559,14 +560,18 @@ func (d *BatchServer) ProcessBatch(
 
 	if d.Config.LockType == LockTypeDb {
 		isShared := !requiresIsolation
-		err = d.DBHandler.WithAdvisoryLock(ctx, isShared, db.LockIsolateTransformers, func(ctx context.Context) error {
-			return d.Repository.Apply(ctx, transformers...)
+		err = logger.Wrap(ctx, func(ctx context.Context) error {
```
sven-urbanski-freiheit-com: if we really need the Wrap function, this wouldn't be the right place to call it.
I'd suggest to remove it for now.

services/cd-service/pkg/service/batch.go
```diff
@@ -559,14 +560,18 @@ func (d *BatchServer) ProcessBatch(
...
+			return d.DBHandler.WithAdvisoryLock(ctx, isShared, db.LockIsolateTransformers, func(ctx context.Context) error {
+				return d.Repository.Apply(ctx, transformers...)
+			})
 		})
 	} else {
 		if d.Config.LockType == LockTypeNone {
 			logging.Info(ctx, "not locking at all")
 		}
-		err = d.Repository.Apply(ctx, transformers...)
+		err = logger.Wrap(ctx, func(ctx context.Context) error {
```
sven-urbanski-freiheit-com: same here

pkg/db/db_apps.go
```diff
@@ -74,6 +74,26 @@ func (h *DBHandler) DBSelectApp(ctx context.Context, tx *sql.Tx, appName types.A
 	return h.processAppsRow(ctx, rows, err)
 }
 
+func (h *DBHandler) DBSelectAllAppsMetadataAtTimestamp(ctx context.Context, tx *sql.Tx, ts time.Time) (_ map[types.AppName]*DBAppWithMetaData, err error) {
```
sven-urbanski-freiheit-com, line 77: approved with small comments.

Let's try it out together locally before we merge this big change.

pkg/db/db_apps.go
```diff
@@ -74,6 +74,26 @@ func (h *DBHandler) DBSelectApp(ctx context.Context, tx *sql.Tx, appName types.A
...
+	}()
+	selectQuery := h.AdaptQuery(`
+	SELECT DISTINCT
+		ON (appname) appname,
+		stateChange,
+		metadata,
+		argoBracket
+	FROM apps_history
+	WHERE
+		created <= ?;
```
sven-urbanski-freiheit-com: order by missing?

services/cd-service/pkg/repository/releasetrain.go
sven-urbanski-freiheit-com: please use `deployCounter` here. `deployedApps` seems like it can be removed

services/cd-service/pkg/repository/releasetrain.go
```diff
@@ -266,7 +277,8 @@ func (c *ReleaseTrain) Transform(
 	types.Sort(envNames)
 	span.SetTag("environments", len(envNames))
 
-	allLatestReleases, err := state.GetAllLatestReleases(ctx, transaction)
+	ts, err := state.GetCommitHashTimestamp(ctx, transaction, c.CommitHash)
```
sven-urbanski-freiheit-com, line 280: error handling is missing!

services/cd-service/pkg/repository/transformer_test.go
```diff
@@ -5776,6 +5776,265 @@ func TestReleaseTrainsWithCommitHash(t *testing.T) {
...
```
sven-urbanski-freiheit-com, line 5780: please add a small test for "invalid commitHash". In this case (commitHash requested, but does not exists) the train should just exit with an error.

pkg/db/db_apps.go
bqcuong: I dont think the order is necessary

services/cd-service/pkg/repository/transformer_test.go
bqcuong: this is already covered in the test: https://github.com/freiheit-com/kuberpult/blob/main/services/cd-service/pkg/repository/transformer_test.go#L2270-L2302

sven-urbanski-freiheit-com (approved):

PR #2873, opened 2026-04-23
feat: allow rollout-service to render brackets

sven-urbanski-freiheit-com:
This allows the rollout/reposerver to work with brackets, and allow a smooth transition from normal apps to brackets.

The cd-service splits up its output to serve some envs as normal, and some with brackets.
The rollout-service picks up on that and immediately creates brackets, when an environment is configured for that.
It also keeps the normal apps until the bracket apps have been created, and only then removes the normal apps.

New Helm parameters:
`rollout.experimentalBrackets.enabled` (default: `false`)
`rollout.experimentalBrackets.clusters` (map of env → bool)

This is considered "experimental" because this options will not stay long, due to the deprecation of the rollout service.

New "kind-tests" (and unit tests) have been added to make sure the rollout-service behaves as expected together with the cd-service. These integration tests are not run in the CI yet, because their setup is quite different.

Ref: SRX-VRKPNL


services/cd-service/pkg/service/overview.go
```diff
@@ -568,9 +571,26 @@ func (o *OverviewServiceServer) StreamChangedApps(in *api.GetChangedAppsRequest,
 				if err != nil {
 					return err
 				}
+				// Strip bracket-enabled envs from app deployments so the rollout-service
+				// does not track individual apps for those envs.
+				for envName := range response.Deployments {
```
sven-urbanski-freiheit-com, line 628: Q: is this a problem for the frontend-service?

charts/kuberpult/values.yaml
```diff
@@ -354,6 +354,22 @@ rollout:
...
     kuberpultEventsMetricsEnabled: false
+  experimentalBrackets:
+    # Master switch for bracket rendering in the rollout service.
+    # When false, the clusters parameter is ignored.
+    enabled: false
+    # Switches an environment to bracket mode.
+    # Environments that are not listed here will be handled as normal apps in the rollout service.
+    # Example:
+    #   <env1>: true
+    #   <env2>: true
```
sven-urbanski-freiheit-com, line 375: Q: does this config work for a single a/a env?

sven-urbanski-freiheit-com: confirmed: tested with A/A envs and updated docs

ccreinartz11 (changes requested):
## freiheit.com Code Review

Found 15 issue(s) — 2 error(s), 8 warning(s), 5 suggestion(s).

### Findings

| Severity | File | Line | Category | Message |
|---|---|---|---|---|
| error | `charts/kuberpult/install-kuberpult-helm.sh` | 21 | Secrets & Credentials | `echo "argocd token: $token"` prints a live bearer token to stdout in plaintext. In any CI/CD pipeline this token will appear in build logs. Remove this line or replace it with a redacted message such as `echo "argocd token obtained"`. |
| error | `charts/kuberpult/run-kind.sh` | 239 | Information Disclosure | `echo "installing argocd $(cat ./argocd-values.yml)"` prints the full ArgoCD configuration file to stdout, including SSH known-hosts entries, RBAC rules, and server URLs. In CI environments this will appear in build logs. Remove or replace with a summary message. |
| warning | `charts/kuberpult/install-kuberpult-helm.sh` | 19 | Script Robustness | `token=$(argocd account generate-token ...)` executes before `set -eu`. A failure in token generation will be silently ignored. Move `set -eu` / `set -o pipefail` to before any commands that should fail-fast. |
| warning | `charts/kuberpult/lib.sh` | 130 | Hardcoded Labels | `kubectl describe pod -l app=kuberpult-cd-service` is hardcoded inside `portForwardAndWait`, a general-purpose function. When called for other deployments (frontend-service, argocd-server) the describe command targets the wrong pod. Parameterise the label or remove this debug line. |
| warning | `docker-compose.yml` | 624 | Developer Experience | `KUBERPULT_EXPERIMENTAL_BRACKETS_CLUSTERS=development` is hardcoded in the local docker-compose rollout-service environment. Every developer running `make kuberpult` will have bracket mode silently enabled for `development`. Consider defaulting to empty and documenting how to opt-in locally. |
| warning | `pkg/api/v1/api.proto` | 675 | Weak Contract | `BracketDeployment.version` is a `string` encoding a colon-separated list of release numbers (e.g. `"1:2:3"`). This format is undocumented in the proto field comment and forces every client to parse the string manually with no schema enforcement. Consider `repeated uint64 versions = 1;` or a wrapper message. |
| warning | `services/cd-service/pkg/service/overview.go` | 849 | Log Level | `logging.Warn(stream.Context(), "StreamChangedApps called", ...)` logs a normal operational event at Warn severity. This will flood production on-call alerts. Change to `logging.Info` or `logging.Debug`, or remove before merge. |
| warning | `services/cd-service/pkg/service/overview.go` | 841 | Missing Update Path | `getChangedBrackets` only processes apps that changed in the current batch. If a bracket membership changes (app added/removed to a bracket) without any individual-app deployment change in the same batch, the bracket's view in the rollout-service will become stale. |
| warning | `services/rollout-service/pkg/argo/argo.go` | 1610 | Fragile Identity Check | `isBracket := currentAppDetails.Application.ArgoBracket == currentApp` — a bracket named `"foo"` whose member app is also named `"foo"` would be misidentified. Prefer an explicit boolean field or a stricter condition. |
| warning | `services/rollout-service/pkg/versions/versions.go` | 94 | Log Level | `logging.Warn(ctx, "getversion called", ...)` is emitted on every single version lookup. This is a trace-level event that will generate enormous log volume and obscure real warnings. Change to `logging.Debug` or remove before merge. |
| warning | `services/rollout-service/pkg/versions/versions.go` | 348 | Log Level | Multiple `logging.Warn` calls inside the bracket processing loop (`"changed bracket loop"`, `"bracket in same version"`, `"env not in bracketclusters"`) are used for normal operational paths. These should be `logging.Info` or `logging.Debug` as appropriate. |
| suggestion | `charts/kuberpult/lib.sh` | 107 | Dead Code | Commented-out `kubectl describe pod` block inside `waitForDeployment`. Dead code in a shared library adds maintenance confusion — remove it or track it in a TODO referencing an issue. |
| suggestion | `services/cd-service/pkg/service/overview.go` | 833 | Concurrent Map Mutation | `delete(response.Deployments, envName)` mutates the map while ranging over it. Go permits this, but add a brief comment (`// safe: Go allows map deletion during range`) so future readers don't flag it as a bug. |
| suggestion | `services/rollout-service/pkg/versions/versions.go` | 122 | Retry Count | `getBracketVersion` calls `db.WithTransactionT` with retry count `1`. Other callers use `2` or `3`. Under DB contention a single retry may cause unnecessary failures. Consider raising to `2` for consistency. |
| suggestion | `services/rollout-service/pkg/versions/versions.go` | 414 | Duplicated Logic | `isBracketEnv` is defined separately in `cd-service/pkg/service/overview.go` and `rollout-service/pkg/argo/argo.go`, with additional inline filtering in `versions.go`. A shared helper in `pkg/types` or a new `pkg/brackets` package would eliminate duplication and ensure consistent semantics. |

sven-urbanski-freiheit-com:
Regarding the "Fragile Identity Check": This just can't happen in a real scenario. We avoid this by always adding the bracket name as an app with zero deployments in versions.go function `addBracketToChange`.
So all brackets have an app of the same name as the bracket.

ccreinartz11:
> The cd-service splits up its output to serve some envs as normal, and some with brackets. The rollout-service picks up on that and immediately creates brackets, when an environment is configured for that. 

How does this affect the manifest-export service? SInce that is "behind" the CD service, it would also be affected, I suppose?

charts/kuberpult/install-kuberpult-helm.sh
```diff
@@ -1,21 +1,34 @@
...
 set -eu
 set -o pipefail
 
+#  experimentalBrackets:
+#    enabled: true
+#    clusters:
+#      development: true
+#      staging: false
+
+
```
ccreinartz11: 🔧 remove

charts/kuberpult/install-kuberpult-helm.sh
```diff
@@ -121,3 +141,15 @@ make release-tag
...
+kubectl get deployment
+kubectl get pods
+
+print "port forwarding to cd service..."
+waitForDeployment "default" "app=kuberpult-cd-service"
+portForwardAndWait "default" deployment/kuberpult-cd-service 8082 8080
+
+waitForDeployment "default" "app=kuberpult-frontend-service"
+portForwardAndWait "default" "deployment/kuberpult-frontend-service" "8081" "8081"
+print "connection to frontend service successful"
```
ccreinartz11, line 149: ❓ why is this addition required?

charts/kuberpult/values.yaml
```diff
@@ -354,6 +354,24 @@ rollout:
...
+    # clusters:
+    #   <env1>: true
+    #   <env2>: true
+    # For active/active environments, you can enable them 1 by 1. The naming scheme is "<commonEnvPrefix>-<envName>-<concreteEnvName>"
+    # When switching this option to true for an environment, the rollout-service will:
+    # 1) keep the normal argo apps for now
+    # 2) create bracket apps for argo - for a moment the apps will "fight" against each other, but not find any meaningful difference.
+    # 3) delete the normal apps with cascade=false, so that only bracket apps remain.
+    # ROLLBACK:
+    # When switching this option from true to false, the rollout-service will just add the apps, it will not delete the brackets.
```
ccreinartz11, line 382: ❓ Would this not result in the same "fight" as before? Just that it does not stop because the brackets are not deleted? 
If yes, that seems like an problem for a possible emergency rollback

services/cd-service/pkg/service/overview.go
```diff
@@ -543,7 +546,7 @@ func (o *OverviewServiceServer) StreamOverview(in *api.GetOverviewRequest,
 	}
 }
 
-func (o *OverviewServiceServer) StreamChangedApps(in *api.GetChangedAppsRequest,
+func (o *OverviewServiceServer) StreamChangedApps(_ *api.GetChangedAppsRequest,
```
ccreinartz11, line 590: ❓ Why is `_ *api.GetChangedAppsRequest,` passed if it is not required (and was not required before apparently)?

services/cd-service/pkg/service/overview.go
```diff
@@ -636,6 +656,130 @@ func (o *OverviewServiceServer) getAllAppNames(ctx context.Context) ([]types.App
...
+		var latestTime time.Time
+		var latestDeployment *api.Deployment
+		var latestAppDetails *api.GetAppDetailsResponse
+
+		for _, appDetail := range sortedAppDetails {
+			dep := appDetail.Deployments[envName]
+			if dep == nil {
+				versionParts = append(versionParts, "0")
+				continue
+			}
```
ccreinartz11, line 731: ❓ Is 'versionParts=0' our encoding for 'no deployment'?

services/cd-service/pkg/service/overview.go
```diff
@@ -636,6 +656,130 @@ func (o *OverviewServiceServer) getAllAppNames(ctx context.Context) ([]types.App
...
+
+func combineBracketDeployments(sortedAppDetails []*api.GetAppDetailsResponse, bracketEnvs []string) map[string]*api.BracketDeployment {
+	result := make(map[string]*api.BracketDeployment)
+	for _, envName := range bracketEnvs {
+		var versionParts []string
+		var latestTime time.Time
+		var latestDeployment *api.Deployment
+		var latestAppDetails *api.GetAppDetailsResponse
+
+		for _, appDetail := range sortedAppDetails {
```
ccreinartz11: 🤔 I am always a bit worried when we rely on sorting inside a function but handle sorting outside the function.
Seems a bit fragile. 
Do we sort once outside the function and then need the sorted array in multiple places?
If we only need it here, sort inside the function please.

services/cd-service/pkg/service/overview.go
```diff
@@ -636,6 +656,130 @@ func (o *OverviewServiceServer) getAllAppNames(ctx context.Context) ([]types.App
...
+				for _, changedApp := range changedApps {
+					if app == changedApp {
+						if !seen[bracketName] {
+							seen[bracketName] = true
+							affectedBrackets = append(affectedBrackets, bracketName)
+						}
+						break
+					}
+				}
+			}
```
ccreinartz11: 🤔 seems inefficient but I don't have an approach that's clearly better.

But I would suggest switching the order of the apps and changedApps loops, because there should be way fewer changedApps than apps and we don't need to iterate through more apps once all changedApps have been hit. (all assuming that apps do not repeat inside `apps`)

services/rollout-service/pkg/argo/argo.go
```diff
@@ -187,10 +202,89 @@ func (a *ArgoAppProcessor) extractFullyQualifiedEnvironmentName(commonPrefix, en
...
+				zap.String("env", pd.EnvironmentName))
+			if knownApps := a.KnownApps[pd.EnvironmentName]; knownApps != nil {
+				a.deleteAppNoCascade(ctx, knownApps, pd.AppName)
+			}
+		} else {
+			remaining = append(remaining, pd)
+		}
+	}
+	a.pendingDeletions = remaining
+}
```
ccreinartz11, line 270: ❓ needs explanation for me

services/rollout-service/pkg/versions/versions.go
```diff
@@ -89,6 +94,15 @@ func (v *versionClient) GetVersion(ctx context.Context, revision, environment, a
...
 	span.SetTag("Application", app)
+	logging.Info(ctx, "getversion called", zap.String("env", environment), zap.String("app", app))
+
+	if slices.Contains(v.experimentalBracketsClusters, environment) {
+		result, err := v.getBracketVersion(ctx, revision, environment, types.ArgoBracketName(app))
+		if err != nil {
+			return nil, onErr(err)
+		}
+		return result, nil
+	}
```
ccreinartz11, line 107: ❓ It seems strange to me that we return early here when the previous path invoked a db transaction. Is it intentional that we skip that for experimental brackets clusters now?

ccreinartz11: Ah, getBracketVersion includes the db transaction. 
Can you wrap the code below in a function, too then? That way the encapsulation is consistent.

services/rollout-service/pkg/versions/versions.go
```diff
@@ -283,6 +348,61 @@ func (v *versionClient) ConsumeEvents(ctx context.Context, processor VersionEven
...
+						logging.Info(ctx, "bracket in same version",
+							zap.Bool("hasVersion", hasVersion),
+							zap.String("bracketVersion", string(bracketVersion)),
+							zap.String("seenVersion", string(seenVersion)),
+						)
+						continue
+					}
+					seenVersions[bracketKey] = bracketVersion
+
+					var dt time.Time
```
ccreinartz11: :nit: Improve variable name

charts/kuberpult/install-kuberpult-helm.sh
sven-urbanski-freiheit-com: It's not an addition, I moved it from run-kind.sh to here.
This way, it's easier to just run `./install-kuberpult-helm.sh` again.

charts/kuberpult/values.yaml
sven-urbanski-freiheit-com: Yes, they would "fight over nothing" in this case.
To resolve that, all we have to do in this case is delete the bracket apps manually (without cascade).

services/cd-service/pkg/service/overview.go
sven-urbanski-freiheit-com: This implements a grpc endpoint with
`message GetChangedAppsRequest {}`. So there is no input data. I did not change this.

services/cd-service/pkg/service/overview.go
sven-urbanski-freiheit-com: I agree with the general sentiment, but I think the code will in this specific case become more ugly if we do this.  Let's discuss.

services/cd-service/pkg/service/overview.go
sven-urbanski-freiheit-com: Yes it is

services/rollout-service/pkg/versions/versions.go
sven-urbanski-freiheit-com: good point, done

services/rollout-service/pkg/argo/argo.go
sven-urbanski-freiheit-com: I added code comments.

services/rollout-service/pkg/versions/versions.go
sven-urbanski-freiheit-com: done

sven-urbanski-freiheit-com:
> > The cd-service splits up its output to serve some envs as normal, and some with brackets. The rollout-service picks up on that and immediately creates brackets, when an environment is configured for that.
> 
> How does this affect the manifest-export service? SInce that is "behind" the CD service, it would also be affected, I suppose?

It's a different endpoint, it does not affect the manifest-export at all.
The manifest-export queries the DB - the rollout-service relies on the grpc endpoints of the cd-service to work.

charts/kuberpult/install-kuberpult-helm.sh
sven-urbanski-freiheit-com: done

services/cd-service/pkg/service/overview.go
sven-urbanski-freiheit-com: done

ccreinartz11 (approved):

charts/kuberpult/templates/_helpers.tpl
```diff
@@ -14,6 +14,20 @@
 
 # Copyright freiheit.com
 
+{{/*
```
sven-urbanski-freiheit-com, line 17: please resolve this comment once we can merge this.

tests/kind-brackets/bracket_stability_test.go
```diff
@@ -0,0 +1,322 @@
...
+//
+// This means: if a pod's startTime changes between versions, its Deployment was
+// deleted (i.e. the bracket Argo app was cascade-deleted unexpectedly).
+func stableManifest(app, namespace, version string) string {
+	return fmt.Sprintf(`---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: %s-bracket-cfg
+  namespace: %s
```
ccreinartz11, line 75: 🔧 Pass as file

ccreinartz11 (approved):

PR #2875, opened 2026-04-24
fix: make kuberpult-freshdb work on Apple Silicon (arm64)

pschwartau:
## Problem

`make kuberpult-freshdb` failed on Mac Apple Silicon when using **colima** as the Docker runtime. Unlike Docker Desktop, colima only supports `linux/arm64` natively and has no Rosetta/amd64 emulation. Three separate root causes each caused a failure at different stages of the build:

### 1. `frontend-service` binary was amd64 in an arm64 image

`services/frontend-service/Dockerfile` had `GOARCH=amd64` hardcoded in the Go build step:

```dockerfile
RUN GOARCH=amd64 GOOS=linux CGO_ENABLED=0 go build ...
```

On Apple Silicon, Docker builds images for the native platform (`linux/arm64`). The Go compiler was cross-compiling an **amd64 binary**, but the image manifest said `linux/arm64`. When colima tried to start the container it got `exec /main: exec format error`.

### 2. `make -C pkg gen` pulled an amd64 image from the registry

The `prepare-compose` target called `make -C pkg gen` without setting `BUILDER_IMAGE`. The default in `Makefile.variables` points to `builder:main` — the CI-published image, which is `linux/amd64`. colima cannot run amd64 containers, so every `docker run` inside `pkg gen` failed with `exec /bin/sh: no such file or directory`.

### 3. Builder image installed an amd64 helm binary

`infrastructure/docker/builder/Dockerfile` downloaded helm from a hardcoded `linux-amd64` URL. On an arm64 build this placed a non-executable x86_64 binary at `/usr/local/bin/helm`. While helm is not called during the `kuberpult-freshdb` path today, the builder image was silently broken for any future target that invokes it.

## Solution

**`services/frontend-service/Dockerfile`** — remove `GOARCH=amd64` so the binary is compiled for the native platform. On amd64 CI runners `GOARCH` defaults to `amd64`, so behaviour is unchanged there.

**`Makefile` (`prepare-compose`)** — pass `BUILDER_IMAGE=...builder:local` when calling `make -C pkg gen`, so it uses the locally built arm64 image instead of the remote CI image.

**`infrastructure/docker/builder/Dockerfile`** — replace the hardcoded `linux-amd64` helm download with an arch-aware install using Docker BuildKit's `TARGETARCH` arg. Both `amd64` and `arm64` sha256 checksums are verified.

**`Makefile` (`prepare-git`)** — new target that auto-initialises the bare git repository at `services/cd-service/repository_remote` when absent, eliminating the manual setup step that was only documented in `README_DEVELOPERS.md`.

## Test plan

- [ ] Run `make kuberpult-freshdb` on Apple Silicon Mac with colima — all services start, UI available at `localhost:3000`
- [ ] Verify CI passes on amd64 (no regression from removing `GOARCH=amd64`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,28 +86,36 @@ prepare-git:
 		rm -rf services/cd-service/repository_checkedout; \
 	fi
 
-prepare-compose: builder prepare-git
+reset-git:
+	rm -rf services/cd-service/repository_remote
+	git init --bare services/cd-service/repository_remote --initial-branch=master
+	git clone services/cd-service/repository_remote services/cd-service/repository_checkedout
+	git -C services/cd-service/repository_checkedout commit --allow-empty -m 'initial commit'
+	git -C services/cd-service/repository_checkedout push origin master
+	rm -rf services/cd-service/repository_checkedout
+
+prepare-compose: builder
 	BUILDER_IMAGE=$(DOCKER_REGISTRY_URI)/infrastructure/docker/builder:local make -C pkg gen
 	IMAGE_TAG=local make -C services/cd-service docker
 	IMAGE_TAG=local make -C services/manifest-repo-export-service docker
 	IMAGE_TAG=local make -C services/frontend-service docker gen-api
 	make -C infrastructure/docker/ui build-pr
 
-kuberpult-datadog: prepare-compose compose-down
+kuberpult-datadog: prepare-compose prepare-git compose-down
 ifndef DD_ENV
 	$(error "DD_ENV should be set to execute this target. E.g., DD_ENV=example-local-nov7-a make kuberpult-datadog")
 else
 	DD_ENV=$(DD_ENV) docker compose -f docker-compose.datadog.yml -f docker-compose.yml -f docker-compose.persist.yml up
 endif
 
-kuberpult: prepare-compose compose-down
+kuberpult: prepare-compose prepare-git compose-down
 	docker compose -f docker-compose.yml -f docker-compose.persist.yml up
 
 reset-db: compose-down
 	# This deletes the volume of the default db location:
 	docker volume rm kuberpult_pgdata
 
-kuberpult-freshdb: prepare-compose compose-down
+kuberpult-freshdb: prepare-compose reset-git compose-down
 	docker compose up
 
 # Run this before starting the unit tests in your IDE:
--- a/Makefile
+++ b/Makefile
@@ -76,8 +76,18 @@ builder:
 compose-down:
 	docker compose -f docker-compose.yml -f docker-compose.datadog.yml down
 
-prepare-compose: builder
-	make -C pkg gen
+prepare-git:
+	@if [ ! -d "services/cd-service/repository_remote" ]; then \
+		echo "Initializing bare git repository for local development..."; \
+		git init --bare services/cd-service/repository_remote --initial-branch=master && \
+		git clone services/cd-service/repository_remote services/cd-service/repository_checkedout && \
+		git -C services/cd-service/repository_checkedout commit --allow-empty -m 'initial commit' && \
+		git -C services/cd-service/repository_checkedout push origin master && \
+		rm -rf services/cd-service/repository_checkedout; \
+	fi
+
+prepare-compose: builder prepare-git
+	BUILDER_IMAGE=$(DOCKER_REGISTRY_URI)/infrastructure/docker/builder:local make -C pkg gen
 	IMAGE_TAG=local make -C services/cd-service docker
 	IMAGE_TAG=local make -C services/manifest-repo-export-service docker
 	IMAGE_TAG=local make -C services/frontend-service docker gen-api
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -76,8 +76,18 @@ builder:
...
+	@if [ ! -d "services/cd-service/repository_remote" ]; then \
+		echo "Initializing bare git repository for local development..."; \
+		git init --bare services/cd-service/repository_remote --initial-branch=master && \
+		git clone services/cd-service/repository_remote services/cd-service/repository_checkedout && \
+		git -C services/cd-service/repository_checkedout commit --allow-empty -m 'initial commit' && \
+		git -C services/cd-service/repository_checkedout push origin master && \
+		rm -rf services/cd-service/repository_checkedout; \
+	fi
+
+prepare-compose: builder prepare-git
```
pschwartau: Would be good if the repository is always reset on "make kuberpult-freshdb", as the DB and repo should be in sync. For a normal "make kuberpult" the logic "create if not present" is correct.

pschwartau (commented):
Good point — fixed in the second commit. Split into `prepare-git` (create-if-absent, used by `kuberpult` and `kuberpult-datadog`) and `reset-git` (always recreates, used by `kuberpult-freshdb`). Also moved git setup out of `prepare-compose` so each top-level target controls its own behaviour.

Makefile
```diff
@@ -76,28 +76,46 @@ builder:
...
 	docker compose -f docker-compose.yml -f docker-compose.datadog.yml down
 
+prepare-git:
+	@if [ ! -d "services/cd-service/repository_remote" ]; then \
+		echo "Initializing bare git repository for local development..."; \
+		git init --bare services/cd-service/repository_remote --initial-branch=master && \
+		git clone services/cd-service/repository_remote services/cd-service/repository_checkedout && \
+		git -C services/cd-service/repository_checkedout commit --allow-empty -m 'initial commit' && \
+		git -C services/cd-service/repository_checkedout push origin master && \
+		rm -rf services/cd-service/repository_checkedout; \
```
sven-urbanski-freiheit-com: why delete the repo at the end? Keep it for easier debugging

Makefile
```diff
@@ -76,28 +76,46 @@ builder:
...
+	fi
+
+reset-git:
+	rm -rf services/cd-service/repository_remote
+	git init --bare services/cd-service/repository_remote --initial-branch=master
+	git clone services/cd-service/repository_remote services/cd-service/repository_checkedout
+	git -C services/cd-service/repository_checkedout commit --allow-empty -m 'initial commit'
+	git -C services/cd-service/repository_checkedout push origin master
+	rm -rf services/cd-service/repository_checkedout
```
sven-urbanski-freiheit-com, line 90: same here, why delete it at the end? it's useful.

Makefile
```diff
@@ -76,28 +76,46 @@ builder:
...
+		git -C services/cd-service/repository_checkedout push origin master && \
+		rm -rf services/cd-service/repository_checkedout; \
+	fi
+
+reset-git:
+	rm -rf services/cd-service/repository_remote
+	git init --bare services/cd-service/repository_remote --initial-branch=master
+	git clone services/cd-service/repository_remote services/cd-service/repository_checkedout
+	git -C services/cd-service/repository_checkedout commit --allow-empty -m 'initial commit'
+	git -C services/cd-service/repository_checkedout push origin master
```
sven-urbanski-freiheit-com: you added the same code twice, please finde a way to simplify this.

infrastructure/docker/builder/Dockerfile
```diff
@@ -23,12 +23,17 @@ RUN go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
 
 RUN go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@v2.5.1
 
-RUN curl --fail --silent --show-error --location --remote-name "https://get.helm.sh/helm-v3.14.2-linux-amd64.tar.gz"
-RUN echo 0885a501d586c1e949e9b113bf3fb3290b0bbf74db9444a1d8c2723a143006a5 helm-v3.14.2-linux-amd64.tar.gz | sha256sum -c
-
-RUN tar xzf helm-v3.14.2-linux-amd64.tar.gz && rm -rf helm-v3.14.2-linux-amd64.tar.gz
-RUN mv linux-amd64/helm /usr/local/bin/helm
-RUN chmod +x /usr/local/bin/helm
+ARG TARGETARCH
```
sven-urbanski-freiheit-com, line 26: how is the targetarch arg set? It must be set both in the local setup an the ci. 
`make -C infrastructure/docker/builder/ build` has to work.

ccreinartz11: it's set automatically based on the host machines architecture.
based on what I see the changes should work in the CI and locally on amd and arm architectures, because the changes include cases for both of these architecture.

sven-urbanski-freiheit-com (approved):

PR #2882, opened 2026-04-28
fix: extra logging for git push

sven-urbanski-freiheit-com:
Ref: SRX-SFV8OM

### DIFF
--- a/services/manifest-repo-export-service/pkg/repository/repository.go
+++ b/services/manifest-repo-export-service/pkg/repository/repository.go
@@ -354,7 +354,10 @@ func (r *repository) useRemote(callback func(*git.Remote) error) error {
 		// So `callback` may run longer than `useRemote`, and if at that point `Disconnect` was already called, we get a `panic`.
 		defer logging.HandlePanic(true)
 		defer remote.Disconnect()
-		errCh <- callback(remote)
+		start := time.Now()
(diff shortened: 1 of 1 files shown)


services/manifest-repo-export-service/pkg/repository/repository.go
```diff
@@ -852,7 +862,10 @@ func (r *repository) Push(ctx context.Context, pushAction func() error) error {
 		func() error {
 			span, _ := tracer.StartSpanFromContext(ctx, "Push")
 			defer span.Finish()
+			start := time.Now()
 			err := pushAction()
+			elapsed := time.Since(start)
+			logging.Info(ctx, "push attempt duration", zap.Duration("elapsed", elapsed), zap.Error(err))
```
ccreinartz11, line 868: 🤔 🔧 We should have a `clock` abstraction for so that `time.Now()` is not called directly in production code.
Otherwise these functions are very hard to test.
But does not have to be introduced as part of this commit, I suppose.

ccreinartz11 (approved):

services/manifest-repo-export-service/pkg/repository/repository.go
sven-urbanski-freiheit-com: That, or we transform it into a DataDog metric.

PR #2883, opened 2026-04-29
fix: add CLAUDE.md

sven-urbanski-freiheit-com:
Ref: SRX-VRKPNL

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -78,12 +78,8 @@ pnpm lint-fix
 
 ## Code Coverage Thresholds
 
-Coverage is enforced at test time and the build fails if thresholds are not met:
-- **cd-service overall**: 66.1%
-- **cd-service `releasetrain.go`**: 89.9%
-- **cd-service `deployment.go`**: 87.3%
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+## Code Coverage Thresholds
+
+Coverage is enforced at test time and the build fails if thresholds are not met:
+- **cd-service overall**: 66.1%
+- **cd-service `releasetrain.go`**: 89.9%
+- **cd-service `deployment.go`**: 87.3%
+- **frontend-service**: 42.0%
+- **pkg** (default): 99.9%
```
ccreinartz11: Where do these coverage thresholds come from? Should they be added as a reference to the point in our code where they are set  rather than hardcoded values so that claude understands when the ychange?

ccreinartz11 (approved):

CLAUDE.md
sven-urbanski-freiheit-com: done

PR #2885, opened 2026-04-29
feat: add manifestRepoExport.rendering.experimentalRootAppFilter

sven-urbanski-freiheit-com:
in order to control which root apps are rendered.

Ref: SRX-ZVUZ9N

### DIFF
--- a/charts/kuberpult/templates/_helpers.tpl
+++ b/charts/kuberpult/templates/_helpers.tpl
@@ -1,0 +1,24 @@
+# This file is part of kuberpult.
+
+# Kuberpult is free software: you can redistribute it and/or modify
+# it under the terms of the Expat(MIT) License as published by
+# the Free Software Foundation.
+
+# Kuberpult is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# MIT License for more details.
+
+# You should have received a copy of the MIT License
+# along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>.
+
+# Copyright freiheit.com
+
+# Selects all root app filter clusters that are set to true
+{{- define "kuberpult.experimentalRootAppFilter" -}}
+{{- $envs := list -}}
+{{- range $k, $v := .Values.manifestRepoExport.rendering.experimentalRootAppFilter.envs -}}
+{{- if $v -}}{{- $envs = append $envs $k -}}{{- end -}}
+{{- end -}}
+{{- sortAlpha $envs | join "," -}}
+{{- end -}}
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -222,6 +222,13 @@ manifestRepoExport:
       initialDelaySeconds: 5
   rendering:
 
+    experimentalRootAppFilter:
+      # enable these configs - if false render ALL env-root apps normally.
+      enabled: false
+      # key: name of the environment in kuberpult. value:true => render the root app of this environment.
+      # This option has no effect if enabled=false.
+      envs: {}
+
     # True if the individual apps should be "rendered" (written) into the manifest repo at
     # /environments/<env>/applications/<app>/manifests/manifests.yaml
     # Keep the default, unless you already moved to brackets based rendering.
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -232,6 +232,8 @@ manifestRepoExport:
       # envs:
       #   development: true
       #   staging: false
+      #
+      # For active/active environments, you can enable them 1 by 1. The naming scheme is "<commonEnvPrefix>-<envName>-<concreteEnvName>".
       envs: {}
 
     # True if the individual apps should be "rendered" (written) into the manifest repo at
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -227,6 +227,11 @@ manifestRepoExport:
       enabled: false
       # key: name of the environment in kuberpult. value:true => render the root app of this environment.
       # This option has no effect if enabled=false.
+      # For example:
+      #
+      # envs:
+      #   development: true
+      #   staging: false
       envs: {}
 
     # True if the individual apps should be "rendered" (written) into the manifest repo at
--- a/services/manifest-repo-export-service/pkg/argocd/render.go
+++ b/services/manifest-repo-export-service/pkg/argocd/render.go
@@ -35,7 +35,14 @@ type RenderOptions struct {
 	RenderApps      bool // do we render apps?
 	RenderBrackets  bool // do we render brackets?
 	PointToBrackets bool // do we point the root app to brackets?
+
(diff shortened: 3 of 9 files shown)


services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -270,6 +270,17 @@ func Run(ctx context.Context) error {
...
+	renderOptions.RootAppFiltering.Enabled, err = valid.ReadEnvVarBool("KUBERPULT_EXPERIMENTAL_ROOT_APP_FILTER_ENABLED")
+	if err != nil {
+		return err
+	}
+	tmp, err := valid.ReadEnvVarAsList("KUBERPULT_EXPERIMENTAL_ROOT_APP_FILTER_ENVIRONMENTS", ",")
+	if err != nil {
+		return err
+	}
+	renderOptions.RootAppFiltering.EnabledEnvironments = types.StringsToEnvNames(tmp)
+	logging.Info(ctx, "root app filter", zap.Any("filter", renderOptions.RootAppFiltering))
```
bqcuong, line 282: Should we minimize the usages of `logging.Info` here? We only want to add actionable logs.

services/manifest-repo-export-service/pkg/argocd/render.go
```diff
@@ -35,7 +35,14 @@ type RenderOptions struct {
 	RenderApps      bool // do we render apps?
 	RenderBrackets  bool // do we render brackets?
 	PointToBrackets bool // do we point the root app to brackets?
+
+	RootAppFiltering RootAppFiltering
+}
+type RootAppFiltering struct {
+	Enabled             bool // if false render all root apps
```
bqcuong:
```suggestion
	Enabled             bool // if false, render all root apps
```

charts/kuberpult/values.yaml
```diff
@@ -222,6 +222,13 @@ manifestRepoExport:
       initialDelaySeconds: 5
   rendering:
 
+    experimentalRootAppFilter:
+      # enable these configs - if false render ALL env-root apps normally.
+      enabled: false
+      # key: name of the environment in kuberpult. value:true => render the root app of this environment.
```
bqcuong, line 228: Are the types of the `envs` map are (string, bool) or (string, string)? 

What happens if `value = false`, `value = 'neither true nor false'`?

Maybe an config example can be added in the comment for better explanation?
```
# configuration Example:
#
# enabled: true
# envs: {'env1': true, 'env2': false}
#
# => with this example, only the root app of environment env1 is rendered?
```

bqcuong: Or if we want to keep only the list of envs for rendering their root apps.
Should a list be sufficient enough?

charts/kuberpult/values.yaml
```diff
@@ -222,6 +222,13 @@ manifestRepoExport:
       initialDelaySeconds: 5
   rendering:
 
+    experimentalRootAppFilter:
```
bqcuong, line 225: Should we also add a test in `chart_test.go` for this `experimentalRootAppFilter` option?

charts/kuberpult/templates/_helpers.tpl
```diff
@@ -0,0 +1,24 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# MIT License for more details.
+
+# You should have received a copy of the MIT License
+# along with kuberpult. If not, see <https://directory.fsf.org/wiki/License:Expat>.
+
+# Copyright freiheit.com
+
+# Selects all root app filter clusters that are set to true
```
bqcuong, line 17: The name of this file indicates that it contains some generic helper logic.

Can we make the code in this file generic? Something like:

* `_helper.tpl`
```
{{- define "utils.joinEnvs" -}}
  {{- $envs := list -}}
  {{- range $k, $v := . -}}
    {{- if $v -}}
      {{- $envs = append $envs $k -}}
    {{- end -}}
  {{- end -}}
  {{- sortAlpha $envs | join "," -}}
{{- end -}}
```

* Then in `manifest-repo-export-service.yaml`, we can call:
```
name: KUBERPULT_EXPERIMENTAL_ROOT_APP_FILTER_ENVIRONMENTS
value: {{ include "utils.joinEnvs .Values.manifestRepoExport.rendering.experimentalRootAppFilter.envs" . | quote }}
```

services/manifest-repo-export-service/pkg/repository/repository_test.go
```diff
@@ -788,6 +788,116 @@ func TestArgoCDFileGeneration(t *testing.T) {
...
+				RenderBrackets:  false,
+				PointToBrackets: false,
+				RootAppFiltering: argocd.RootAppFiltering{
+					Enabled:             tc.FilterEnabled,
+					EnabledEnvironments: tc.FilteredEnvs,
+				},
+			}
+			ctx := testutilauth.MakeTestContext()
+
+			transformers := []Transformer{
```
bqcuong, line 842: nitpick: I'll keep this setup of transformers in the definition of test cases. This makes sense when we want to add new tests with different setups (e.g., create 3 instead of 2 envs), we won't have to update all the existing tests.

services/manifest-repo-export-service/pkg/cmd/server.go
sven-urbanski-freiheit-com: this is just printing one line on startup. I'd like to keep it for now.

services/manifest-repo-export-service/pkg/repository/repository_test.go
sven-urbanski-freiheit-com: I disagree. The test cases should always be as simple as possible. Only make configurable, what needs to be. Everything else should be static.
It's also easy enough to adapt my code later into your suggestion - the other way around is not always that easy.

charts/kuberpult/templates/_helpers.tpl
sven-urbanski-freiheit-com: I would delay this until we have 2 cases that actually use this util.

charts/kuberpult/values.yaml
```diff
@@ -222,6 +222,18 @@ manifestRepoExport:
...
+      # enable these configs - if false render ALL env-root apps normally.
+      enabled: false
+      # key: name of the environment in kuberpult. value:true => render the root app of this environment.
+      # This option has no effect if enabled=false.
+      # For example:
+      #
+      # envs:
```
sven-urbanski-freiheit-com, line 232: Q: how to configure it for active/active envs?

charts/kuberpult/values.yaml
sven-urbanski-freiheit-com: I added a configuration example and also described how active/active works here.

services/manifest-repo-export-service/pkg/argocd/render.go
sven-urbanski-freiheit-com: done

charts/kuberpult/values.yaml
sven-urbanski-freiheit-com: done

bqcuong (approved):

PR #2887, opened 2026-04-30
fix: allow running multiple kuberpult instances simultaneously

sven-urbanski-freiheit-com:
This change only affects the local setup.

Parameterize all host port bindings in docker-compose.yml via KUBERPULT_PORT_* env vars with defaults matching the current values. Add COMPOSE_PROJECT_NAME support to namespace container names and volumes per instance. Fix reset-db to use the project name instead of a hardcoded volume name. Change the package.json dev proxy to use the service name (frontend-service) instead of the container name so it works regardless of COMPOSE_PROJECT_NAME. Add .env.example documenting a second-instance configuration with all ports shifted by +100.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

Ref: SRX-CHXDQM


docker-compose.yml
```diff
@@ -114,7 +114,7 @@ services:
   frontend-service:
     image: europe-west3-docker.pkg.dev/fdc-public-docker-registry/kuberpult/kuberpult-frontend-service:local
     # Note: this `container_name` needs to be the same as in `package.json`
```
pschwartau: **[warning] k8s · Documentation**

This comment is now stale: `package.json` no longer references the `container_name` value. The proxy was changed to use the docker-compose service name `frontend-service` directly (which is the correct approach for service-to-service DNS in docker-compose networks). The comment should be updated to reflect that, or removed.

docker-compose.yml
```diff
@@ -114,7 +114,7 @@ services:
...
-    container_name: kuberpult-frontend-service
+    container_name: ${COMPOSE_PROJECT_NAME:-kuberpult}-frontend-service
```
pschwartau, line 121: **[suggestion] k8s · Configuration**

The `container_name` is now dynamic (`${COMPOSE_PROJECT_NAME:-kuberpult}-frontend-service`), but `package.json` uses a static `http://frontend-service:8081` (the service name, not the container name). These two mechanisms are independent: docker-compose service DNS always resolves by service name regardless of `container_name`. This is correct behavior, but the relationship between the two settings is worth documenting so future readers understand why `container_name` is dynamic while the proxy URL stays static.

infrastructure/scripts/create-testdata/ports.sh
```diff
@@ -0,0 +1,20 @@
...
+# Source this file to get the actual host ports.
+# Reads the resolved configuration from docker-compose.yml (honoring .env overrides)
+# when docker and jq are available; otherwise falls back to the default ports.
+
+_REPO_ROOT=$(git -C "$(dirname "${BASH_SOURCE[0]}")" rev-parse --show-toplevel 2>/dev/null || true)
+
+if [ -n "$_REPO_ROOT" ] && command -v docker &>/dev/null && command -v jq &>/dev/null; then
+    _DC_JSON=$(docker compose -f "${_REPO_ROOT}/docker-compose.yml" config --format json 2>/dev/null || true)
+    if [ -n "$_DC_JSON" ]; then
+        FRONTEND_PORT=$(printf '%s' "$_DC_JSON" | jq -r '.services["frontend-service"].ports[] | select(.target == 8081) | .published // empty' 2>/dev/null || true)
```
pschwartau, line 14: **[warning] security · Shell Robustness**

The jq query uses `.published // empty` to extract the host port. In `docker compose config --format json` output the `ports` entries are objects with a `published` key that holds either a string or an integer depending on the compose version and whether the port binding specifies an IP. If `docker compose` is not installed or the compose file references images that are not yet pulled, `docker compose config` may still succeed but produce output that differs from expectations. The `|| true` guards prevent script failure, but consider logging a warning when the dynamic port extraction fails, so users know they are running on defaults.

infrastructure/scripts/create-testdata/manifest-setup.sh
```diff
@@ -0,0 +1,35 @@
...
+
+set -eu
+set -o pipefail
+
+SCRIPT_DIR="$(dirname "$0")"
+
+START_VERSION="${1:-100}"
+NUM_VERSIONS="${2:-7}"
+
+"${SCRIPT_DIR}/create-environments.sh" || true
```
pschwartau: **[suggestion] go · Error Handling**

`create-environments.sh || true` silently suppresses all failures. If the frontend service is not yet healthy (e.g., the user runs this before `docker compose up` finishes), the script will continue and the subsequent `create-release.sh` calls will also fail — but with less obvious error messages. Consider replacing `|| true` with a retry loop or at minimum a warning message on failure, e.g.:
```bash
"${SCRIPT_DIR}/create-environments.sh" || echo "Warning: create-environments.sh failed, continuing anyway"
```

pschwartau (commented):
## freiheit.com Code Review

Found 4 issue(s) across 3 file(s) — 0 error(s), 2 warning(s), 2 suggestion(s).

infrastructure/scripts/create-testdata/manifest-setup.sh
```diff
@@ -0,0 +1,35 @@
...
+
+TEAM="sre"
+
+for v in $(seq 1 "$NUM_VERSIONS"); do
+    echo "$v"
+    REVISION=$v RELEASE_VERSION=$((START_VERSION + v)) \
+        "${SCRIPT_DIR}/create-release.sh" "${APP_NAME}" "${TEAM}" "" "${ARGO_BRACKET}"
+done
+
+"${SCRIPT_DIR}/run-releasetrain.sh" staging "${TEAM}"
```
pschwartau: **[error] shell · Incorrect Script Invocation**

`run-releasetrain.sh` uses its second positional argument (`$2`) as a raw URL query string — the URL becomes `/releasetrain?sre`, which is not a valid API call. The team name `sre` is not a recognised query parameter for the release-train endpoint. Either drop the second argument (to run the train for all teams) or construct the correct query string, e.g. `"team=${TEAM}"`.

infrastructure/scripts/create-testdata/create-many-releases.sh
```diff
@@ -43,7 +46,7 @@ for appId in $(seq -w "${MIN}" "${MAX}" ); do
   }
 ]
 }' | \
-    evans --header author-name=YXV0aG9y --header author-email=YXV0aG9yQGF1dGhvcg== --host localhost --port 8443 -r cli call api.v1.BatchService.ProcessBatch \
+    evans --header author-name=YXV0aG9y --header author-email=YXV0aG9yQGF1dGhvcg== --host localhost --port ${CD_GRPC_PORT} -r cli call api.v1.BatchService.ProcessBatch \
     && echo "app ${appId} deletion success" \
```
pschwartau, line 50: **[warning] shell · Shell Quoting**

`${CD_GRPC_PORT}` is passed unquoted to `evans --port`. Although port values are numeric and safe in practice, quoting all variable expansions is a shell best-practice and keeps shellcheck clean. Use `"${CD_GRPC_PORT}"` (or just `"$CD_GRPC_PORT"`).

.gitignore
```diff
@@ -39,4 +39,5 @@ docker.env
 
 # local setup
 .direnv
-.envrc.local
\ No newline at end of file
+.envrc.local
+.env
```
pschwartau: **[suggestion] general · Missing Newline at End of File**

The file still lacks a trailing newline (`\ No newline at end of file`). Most editors and POSIX tools expect a final newline; many linters and diff tools flag its absence. Please add a newline after `.env`.

pschwartau (dismissed):
## freiheit.com Code Review

Found 3 issue(s) across 3 file(s) — 1 error(s), 1 warning(s), 1 suggestion(s).

infrastructure/scripts/create-testdata/manifest-setup.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
ccreinartz11, line 1: 🤔 Seeing the benefits in the current client project we should consider to start using typescript. 
I will leave the decision whether you want to introduce it now up to you.
But give the idea some thought to convert this script already to typescript. It's "complicated" enough to be justified.

ccreinartz11, line 1: Then again, it also calls a bunch of other shellscripts already.
Maybe create a story. 
"Convert "create-release.sh" + helpers to typescript.

services/frontend-service/package.json
```diff
@@ -104,5 +104,5 @@
       "last 1 safari version"
     ]
   },
-  "proxy": "http://kuberpult-frontend-service:8081"
+  "proxy": "http://frontend-service:8081"
```
ccreinartz11, line 107: 🤔 I don't understand why the proxy changes here.

.env.example
```diff
@@ -0,0 +1,20 @@
...
+COMPOSE_PROJECT_NAME=kuberpult2
+
+KUBERPULT_PORT_POSTGRES=5533
+KUBERPULT_PORT_CD_HTTP=8180
+KUBERPULT_PORT_CD_GRPC=8543
+KUBERPULT_PORT_MANIFEST_HTTP=8190
+KUBERPULT_PORT_MANIFEST_GRPC=8544
+KUBERPULT_PORT_FRONTEND_HTTP=8181
+KUBERPULT_PORT_FRONTEND_GRPC=5655
+KUBERPULT_PORT_UI=3100
```
ccreinartz11: 🤔 Can we simplfy the "set env vars + spin up kuberpult" even more by creating a new make target that you can e.g. call with:
`make kuberpult 3` or something? (3 if it's the 3rd parallel version)

.gitignore
```diff
@@ -40,4 +40,5 @@ docker.env
...
```
ccreinartz11: Usually, `.env` is exposed to be able to share configurations and `.env.local` is ignored.
At least that's what I'm used to.

ccreinartz11 (approved):

services/frontend-service/package.json
sven-urbanski-freiheit-com: Both the service name and the container name are valid options.
But in this change, we make the container name dynamic (in docker-compose), so we can't use that here easily.

.gitignore
sven-urbanski-freiheit-com: good point I will change that

infrastructure/scripts/create-testdata/manifest-setup.sh
sven-urbanski-freiheit-com: I created a story for now

.env.example
sven-urbanski-freiheit-com: that would be an option.
For now I changed the ports so that they are easier to change, e.g.
KUBERPULT_PORT_POSTGRES=4000
KUBERPULT_PORT_CD_HTTP=4001
etc.

PR #2890, opened 2026-05-07
fix: improve claude.md about testing, etc

sven-urbanski-freiheit-com:
Ref: SRX-5TDEDD

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,6 +60,17 @@ pnpm test              # Watch mode
 pnpm test-ci           # CI mode (no watch)
 ```
 
+
+## How to write unit tests
+Even for the simplest test, immediately create a "table" (go slice), so that testing different variations becomes easier in the future.
+
+In a table driven test, only put the really relevant parts into the table data.
+Data that is identical for all cases, should not be part of the table.
+
+Avoid function parameters in table tests.
+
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -60,6 +60,17 @@ pnpm test              # Watch mode
...
+## How to write unit tests
+Even for the simplest test, immediately create a "table" (go slice), so that testing different variations becomes easier in the future.
+
+In a table driven test, only put the really relevant parts into the table data.
+Data that is identical for all cases, should not be part of the table.
+
+Avoid function parameters in table tests.
+
+Omit the line "tc := tc" at the beginning of test loops, it is outdated.
+
```
ccreinartz11, line 73: 🔧 Apparently claude works better if a small code snippet is provided as part of the instructions. (1-2 examples). Here, a short table test structure snippet would be sufficnet I guess.
I don't know yet what the correct place for such snippets would be. I'd suggest to just put it in the claude.md for now.

ccreinartz11 (approved):

CLAUDE.md
sven-urbanski-freiheit-com: We already have a snippet for tests in this file, but it didn't seem sufficient. That's why I'm now trying to describe it verbally.

PR #2893, opened 2026-05-07
feat: Store new manifest-lock in DB

sven-urbanski-freiheit-com:
This creates a new table, and read-only interaction for a new kind of lock: Manifest-locks.
It also displays manifest-locks in the UI.

Manifest-locks are intended to stop the manifest-export from overwriting a specific file in an emergency situation.

Interactions like adding and removing locks in the UI will be added in another PR.

Ref: SRX-1V3OAY

### DIFF
--- a/services/cd-service/pkg/service/overview.go
+++ b/services/cd-service/pkg/service/overview.go
@@ -413,6 +413,41 @@ func (o *OverviewServiceServer) GetAllEnvTeamLocks(ctx context.Context,
 	})
 }
 
+func (o *OverviewServiceServer) GetAllManifestLocks(ctx context.Context,
+	in *api.GetAllManifestLocksRequest) (*api.GetAllManifestLocksResponse, error) {
+
+	span, ctx := tracer.StartSpanFromContext(ctx, "GetAllManifestLocks")
+	defer span.Finish()
+
+	return db.WithTransactionT(o.DBHandler, ctx, 2, true, func(ctx context.Context, transaction *sql.Tx) (*api.GetAllManifestLocksResponse, error) {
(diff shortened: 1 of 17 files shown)


services/cd-service/pkg/service/overview.go
```diff
@@ -413,6 +413,41 @@ func (o *OverviewServiceServer) GetAllEnvTeamLocks(ctx context.Context,
...
 }
 
+func (o *OverviewServiceServer) GetAllManifestLocks(ctx context.Context,
+	in *api.GetAllManifestLocksRequest) (*api.GetAllManifestLocksResponse, error) {
+
+	span, ctx := tracer.StartSpanFromContext(ctx, "GetAllManifestLocks")
+	defer span.Finish()
+
+	return db.WithTransactionT(o.DBHandler, ctx, 2, true, func(ctx context.Context, transaction *sql.Tx) (*api.GetAllManifestLocksResponse, error) {
+		locks, err := o.DBHandler.DBSelectAllActiveManifestLocks(ctx, transaction)
```
bqcuong, line 423: It seems that `locks` and `err` might be `nil` at the same time (returned from `processManifestLockRows`).

bqcuong: Please also check for the usage of `DBSelectAllActiveManifestLocks` in other places.

bqcuong (approved):

services/cd-service/pkg/service/overview.go
sven-urbanski-freiheit-com: both len(nil) and range nil should be safe. But you are right anyway, it's nicer to check this explicitely. I'll add it here.
The function is only called from here.

PR #2894, opened 2026-05-07
fix(docs): give every table a description

sven-urbanski-freiheit-com:
Ref: SRX-5TDEDD

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -270,6 +270,16 @@ const (
 	EvtDeleteAAEnvironmentConfig        EventType = "EvtDeleteAAEnvironmentConfig"
 )
 
+/*
+The event_sourcing_light (ESL) table is the master event log for all state changes in kuberpult.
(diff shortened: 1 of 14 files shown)


pkg/db/db.go
```diff
@@ -525,8 +535,16 @@ func (h *DBHandler) DBCountEslEventsNewer(ctx context.Context, tx *sql.Tx, eslVe
 	return count, nil
 }
 
+/*
+commit_events (renamed from "events") links git commits to the kuberpult transformer that produced them.
```
ccreinartz11: ❓ Does the renaming really matter?

ccreinartz11 (approved):

pkg/db/db.go
sven-urbanski-freiheit-com:
```suggestion
commit_events links git commits to the kuberpult transformer that produced them.
```

sven-urbanski-freiheit-com: good point, I removed it

PR #2902, opened 2026-05-12
fix: refactor endpoint GetCommitInfo to fetch data from database

bqcuong:
Ref: [SRX-X5PFER](https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/STORY/-Ooxxgmax0jG4koPidn-/)

The endpoint `GetCommitInfo` is now fetching data from database instead the manifest repo.
This is the final step for removing the minimizeGitData feature.

image
image
image


docs/operators/endpoint-releasetrain.md
```diff
@@ -13,8 +13,6 @@ or
 but in the state given by this git tag. If the tag does not exist, the endpoint will fail.
 * `gitTag=${myNonExistingManifestRepoGitTag}` is an optional parameter. If set, the manifest-export will create the git tag on the manifest repo after successfully pushing the commit.
 What happens if the creation of the git tag fails depends on the option `manifestRepoExport.failOnErrorWithGitPushTags` (see `charts/kuberpult/values.yaml`).
-If the option `git.minimizeExportedData` is `true`, and the release train changes nothing (meaning all deployments are already in the desired version),
```
sven-urbanski-freiheit-com, line 16: this should be rephrased rather then deleted entirely.

"If the release train changes nothing..."

pkg/db/db_releases.go
```diff
@@ -35,14 +35,15 @@ import (
...
-	SourceCommitId  string
-	SourceMessage   string
-	DisplayVersion  string
-	UndeployVersion bool
-	IsMinor         bool
-	CiLink          string
-	IsPrepublish    bool
+	SourceAuthor     string
+	SourceCommitId   string
+	PreviousCommitId string
```
sven-urbanski-freiheit-com: I'm a bit skeptical if this is the right place to store the `PreviousCommitId`. The commit table seems to be the more appropriate place.

bqcuong: I would argue this for two reasons:
1. `PreviousCommitId` is currently passed through the `CreateApplicationVersion` transformer. One `release` can have only one `PreviousCommitId`. So it makes sense that the `PreviousCommitId` should be stored in the `releases` table, similar to `SourceCommitId`, `SourceMessage`, etc.

2. Is the "commit table" you mentioned the `commit_events` table? If yes, there we store different events (not only `new-release` event type when we execute `CreateApplicationVersion`). Please see the screenshots below, and check the other places where the function `WriteEvent()` is used. Besides, we already implemented these `Commit Events` in a separated table in our UI page (and I don't see Previous Commit buttons for each of these events, but one for the whole page).

image

image

docs/operators/endpoint-releasetrain.md
bqcuong: rephrased

services/manifest-repo-export-service/pkg/service/git.go
```diff
@@ -81,86 +77,71 @@ func (s *GitServer) GetGitTags(ctx context.Context, _ *api.GetGitTagsRequest) (*
...
-	if dat, err := util.ReadFile(fs, sourceMessagePath); err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			return nil, status.Error(codes.NotFound, "commit info does not exist")
+		commitID := releases[0].Metadata.SourceCommitId
+		sourceMessage := releases[0].Metadata.SourceMessage
+		prevCommitID := releases[0].Metadata.PreviousCommitId
+
+		var touchedApps []string
+		for _, release := range releases {
+			touchedApps = append(touchedApps, string(release.App))
```
sven-urbanski-freiheit-com, line 102: please sort the `touchedApps` by name, and remove duplicates.

services/manifest-repo-export-service/pkg/service/git.go
```diff
@@ -189,64 +170,17 @@ func (s *GitServer) ReadEvent(_ context.Context, fs billy.Filesystem, eventPath
...
-		commitPath := fs.Join("commits", commitPrefix[:2], commitPrefix[2:])
-
-		if _, err := fs.Stat(commitPath); err != nil {
-			return "", grpcErrors.NotFoundError(ctx,
-				fmt.Errorf("commit %s was not found in the manifest repo", commitPrefix))
-		}
 
-		return commitPrefix, nil
-	}
+	commitPrefix = strings.ToLower(commitPrefix)
```
sven-urbanski-freiheit-com: this "toLower" call should be done before the validation, so that we also use the lowercase version for db queries

services/manifest-repo-export-service/pkg/repository/transformer_test.go
```diff
@@ -1543,87 +667,13 @@ func TestDeleteEnvironment(t *testing.T) {
 				},
 				{
 					path:     "/argocd/v1alpha1/acceptance.yaml",
-					fileData: []byte(authorEmail),
-				},
-			},
-			expectedMessage: "delete environment \"acceptance\"",
-		},
-		{
-			Name: "create two environments and delete one of them",
```
sven-urbanski-freiheit-com, line 1552: why was this deleted? Sounds independent of GetCommitInfo

services/manifest-repo-export-service/pkg/repository/transformer_test.go
```diff
@@ -2135,276 +1217,6 @@ spec:
...
-					Team:                  "team-123",
-					TransformerMetadata: TransformerMetadata{
-						AuthorName:  authorName,
-						AuthorEmail: authorEmail,
-					},
-				},
-			},
-			ExpectedFile: []*FilenameAndData{
-				{
-					path:     "applications/" + appName + "/releases/1.0/environments/production/manifests.yaml",
```
sven-urbanski-freiheit-com, line 2215: this test looks like it should be adapted, not deleted.

Just change the paths to point to the actual manifest.

services/manifest-repo-export-service/pkg/service/git_test.go
```diff
@@ -644,39 +644,21 @@ func TestGetCommitInfo(t *testing.T) {
...
 			},
 			request: &api.GetCommitInfoRequest{
 				CommitHash: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
 			},
-			allowReadingCommitData: false, // … but do not return it
-			expectedError:          status.Error(codes.NotFound, "error: commit aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa was not found in the manifest repo"),
-			expectedResponse:       nil,
-		},
-		{
-			name: "no commit info written if toggle not set",
```
sven-urbanski-freiheit-com, line 659: why was this removed?
We still want GetCommitinfo to return an error if the commit was not found, so this test should stay.

pkg/db/db_releases.go
```diff
@@ -357,6 +359,39 @@ func (h *DBHandler) DBSelectLatestReleaseOfApp(ctx context.Context, tx *sql.Tx,
 	return h.processReleaseRow(ctx, err, rows, ignorePrepublishes, false)
 }
 
+func (h *DBHandler) DBSelectAllReleasesBySourceCommit(ctx context.Context, tx *sql.Tx, sourceCommitID string, ignorePrepublishes bool) (_ []*DBReleaseWithMetaData, err error) {
+	selectQuery := h.AdaptQuery(`
+		SELECT created, appName, metadata, releaseVersion, environments, revision
+		FROM ` + releasesTable + `
+		WHERE (metadata::jsonb)->>'SourceCommitId' LIKE ?
+		ORDER BY releaseversion DESC, revision DESC;
+	`)
```
sven-urbanski-freiheit-com, line 366: This query needs a limit (or at least a comment explaining why it's not needed).

pkg/db/db_releases.go
```diff
@@ -357,6 +359,39 @@ func (h *DBHandler) DBSelectLatestReleaseOfApp(ctx context.Context, tx *sql.Tx,
 	return h.processReleaseRow(ctx, err, rows, ignorePrepublishes, false)
 }
 
+func (h *DBHandler) DBSelectAllReleasesBySourceCommit(ctx context.Context, tx *sql.Tx, sourceCommitID string, ignorePrepublishes bool) (_ []*DBReleaseWithMetaData, err error) {
+	selectQuery := h.AdaptQuery(`
+		SELECT created, appName, metadata, releaseVersion, environments, revision
+		FROM ` + releasesTable + `
+		WHERE (metadata::jsonb)->>'SourceCommitId' LIKE ?
```
sven-urbanski-freiheit-com: This json parsing is not ideal.
I think we need a new table just for commits.

services/manifest-repo-export-service/pkg/service/git_test.go
bqcuong: This test is duplicated with the test "no commit events written if WriteCommitData is set to false"

services/manifest-repo-export-service/pkg/service/git.go
```diff
@@ -81,86 +77,71 @@ func (s *GitServer) GetGitTags(ctx context.Context, _ *api.GetGitTagsRequest) (*
...
-	}
+		err := validateCommitPrefix(commitPrefix)
+		if err != nil {
+			return nil, err
+		}
 
-	commitPath := fs.Join("commits", commitID[:2], commitID[2:])
+		releases, err := dbHandler.DBSelectAllReleasesBySourceCommit(ctx, transaction, commitPrefix, true)
+		if err != nil {
+			return nil, fmt.Errorf("could not read release with SourceCommitId prefix %s from DB: %v", commitPrefix, err)
```
sven-urbanski-freiheit-com: this is a pretty big query for what you want to get.
It has to scan all releases, even though for the most part you just use the first release.
The other part is the `touchedApps`. For this, getting the releases is a bit more practical - but even then, we can't define an upper bound on how many apps there are. So the query still returns potentially too much.
=> We need to find a better DB schema to tackle this.

pkg/db/db_releases.go
bqcuong: We want to get all the releases with this SourceCommitId so I won't put a limit here. This is for obtaining the unique set of the app names.
I think I can improve the query a bit by using DISTINCT (app)

services/manifest-repo-export-service/pkg/repository/transformer_test.go
bqcuong: Adapted these tests to use the paths `environments/production/applications/...`.
Note that only the manifests of the latest versions are stored.

services/manifest-repo-export-service/pkg/repository/transformer_test.go
bqcuong: This test is not deleted, GitHub may show something stupid.
Please check the line 676, on the new version of `transformer_test.go`, you'll see this test still exists: "services/manifest-repo-export-service/pkg/repository/transformer_test.go:676"

pkg/db/db_releases.go
bqcuong: fixed

services/manifest-repo-export-service/pkg/service/git.go
bqcuong: fixed by introducing new table `commits_history` and new column `releases.commitHash`

pkg/db/db_releases.go
bqcuong: fixed

pkg/db/db.go
```diff
@@ -535,6 +535,86 @@ func (h *DBHandler) DBCountEslEventsNewer(ctx context.Context, tx *sql.Tx, eslVe
 	return count, nil
 }
 
+const commitHistoryTable = "commits_history"
```
sven-urbanski-freiheit-com, line 541:
```suggestion
/*
The commit history stores all commits and how they are connected.
*/
const commitHistoryTable = "commits_history"
```

pkg/db/db.go
```diff
@@ -535,6 +535,86 @@ func (h *DBHandler) DBCountEslEventsNewer(ctx context.Context, tx *sql.Tx, eslVe
...
+		previousCommitHash,
+	)
+	if err != nil {
+		return fmt.Errorf("error inserting commit history row into DB. Error: %w", err)
+	}
+	return nil
+}
+
+func (h *DBHandler) DBGetNextCommit(ctx context.Context, transaction *sql.Tx, commitHash string) (_ *string, err error) {
+	selectQuery := h.AdaptQuery("SELECT commitHash FROM " + commitHistoryTable + " WHERE previousCommitHash = ?;")
```
sven-urbanski-freiheit-com: please keep the pattern of using ` and a newline for each main sql keyword:

```suggestion
	selectQuery := h.AdaptQuery(`
	SELECT commitHash
	FROM " + commitHistoryTable + 
	" WHERE previousCommitHash = ?;
	`)
```

pkg/db/db.go
```diff
@@ -535,6 +535,86 @@ func (h *DBHandler) DBCountEslEventsNewer(ctx context.Context, tx *sql.Tx, eslVe
...
+		commitHash,
+		previousCommitHash,
+	)
+	if err != nil {
+		return fmt.Errorf("error inserting commit history row into DB. Error: %w", err)
+	}
+	return nil
+}
+
+func (h *DBHandler) DBGetNextCommit(ctx context.Context, transaction *sql.Tx, commitHash string) (_ *string, err error) {
```
sven-urbanski-freiheit-com: it's customary in go to return "" for not existing string.
Since "" is not a valid commitHash, I would keep this standard.
```suggestion
func (h *DBHandler) DBGetNextCommit(ctx context.Context, transaction *sql.Tx, commitHash string) (_ string, err error) {
```

pkg/db/db.go
```diff
@@ -535,6 +535,86 @@ func (h *DBHandler) DBCountEslEventsNewer(ctx context.Context, tx *sql.Tx, eslVe
 	return count, nil
 }
 
+const commitHistoryTable = "commits_history"
+
+func (h *DBHandler) DBWriteCommitHistoryRow(ctx context.Context, transaction *sql.Tx, commitHash, previousCommitHash string) (err error) {
+	insertQuery := h.AdaptQuery(`
```
sven-urbanski-freiheit-com, line 544: not sure if this is the best place to check it - but I would only insert hashes that are complete (40 chars).

Then we could also add a constraint in the DB, something like this:

ALTER TABLE your_table_name
ADD CONSTRAINT check_valid_code 
CHECK (length(your_column) = 40);

pkg/db/db_releases.go
```diff
@@ -357,6 +357,22 @@ func (h *DBHandler) DBSelectLatestReleaseOfApp(ctx context.Context, tx *sql.Tx,
 	return h.processReleaseRow(ctx, err, rows, ignorePrepublishes, false)
 }
 
+func (h *DBHandler) DBSelectAllLatestReleasesByCommitHash(ctx context.Context, tx *sql.Tx, commitHash string, ignorePrepublishes bool) (_ []*DBReleaseWithMetaData, err error) {
+	selectQuery := h.AdaptQuery(`
+		SELECT DISTINCT ON (appName) created, appName, metadata, releaseVersion, environments, revision
+		FROM ` + releasesTable + `
+		WHERE commitHash LIKE ?
```
sven-urbanski-freiheit-com: why use LIKE here? We should use `=`

If you need to find a commitHash based on a commitHashPrefix, please first query the commit table to get the full hash, and then query releases.

services/manifest-repo-export-service/pkg/repository/repository_test.go
```diff
@@ -2433,108 +2429,6 @@ func setupRepositoryBenchmarkWithPath(t *testing.B) (Repository, string) {
 	return repo, remoteDir
 }
 
-func BenchmarkApplyQueue(t *testing.B) {
```
sven-urbanski-freiheit-com, line 2436: why was this removed? I think it should stay

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -792,93 +762,20 @@ func (c *CreateApplicationVersion) Transform(
...
-	if c.WriteCommitData && tCtx.ShouldMaximizeGitData() {
-		ev, err := state.DBHandler.DBSelectAllCommitEventsForTransformer(ctx, transaction, c.TransformerEslVersion, event.EventTypeNewRelease, 1)
-		if err != nil {
-			return "", GetCreateReleaseGeneralFailure(err)
-		}
-		if len(ev) == 0 {
-			return "", fmt.Errorf("no new release event to read from database for application '%s'", c.Application)
-		}
-
-		err = writeCommitData(ctx, c.SourceCommitId, c.SourceMessage, c.Application, c.PreviousCommit, state)
```
sven-urbanski-freiheit-com, line 855: you still have the function writeCommitdata, shouldn't it be called here?

services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -904,82 +801,13 @@ func (c *CreateApplicationVersion) Transform(
...
-	if tCtx.ShouldMinimizeGitData() && len(deploymentsMap) == 0 {
+	if len(deploymentsMap) == 0 {
 		return GetNoOpMessage(c)
 	}
 
 	return fmt.Sprintf("created version %v of %q", version, c.Application), nil
 }
 
-func writeCommitData(ctx context.Context, sourceCommitId string, sourceMessage string, app string, previousCommitId string, state *State) error {
-	fs := state.Filesystem
```
sven-urbanski-freiheit-com, line 915: I expected that in these functions you would now write the commit to the db

services/manifest-repo-export-service/Makefile
```diff
@@ -14,7 +14,7 @@
 
 # Copyright freiheit.com
 
-MIN_COVERAGE=45.8
+MIN_COVERAGE=42.6
```
sven-urbanski-freiheit-com: you reactivated some tests, can we increase this now?

database/migrations/postgres/1778597687238527_commits_history.up.sql
```diff
@@ -0,0 +1,12 @@
...
+    IF NOT EXISTS (SELECT 1 FROM information_schema.columns WHERE table_name = 'releases' AND column_name='commitHash') THEN
+    ALTER TABLE IF EXISTS releases ADD COLUMN commitHash VARCHAR(64);
+END IF;
+END $$;
+
+CREATE TABLE IF NOT EXISTS commits_history
+(
+    commitHash VARCHAR(64) PRIMARY KEY NOT NULL,
+    previousCommitHash VARCHAR(64)
+);
```
sven-urbanski-freiheit-com, line 12: please also create an index on `releases` table and `commitHash` column, since we need to filter for that.

bqcuong: Actually `commitHash` in the `releases` are not unique.
The combination of (app, commitHash) is unique.
Therefore, I didn't add the index for this column in the releases table.

pkg/db/db.go
bqcuong: This is already checked in the CreateApplicationVersion transformer. We only call DBWriteCommitHistoryRow from there.

services/manifest-repo-export-service/pkg/repository/transformer.go
bqcuong: No we dont have writeCommitdata for manifest-repo-export-service anymore

services/manifest-repo-export-service/pkg/repository/transformer.go
bqcuong: The function with the same name in cd-service has been used for doing this job for a while.

services/manifest-repo-export-service/pkg/repository/repository_test.go
bqcuong: This test fetched data from manifest repo (`repo.State().GetAllApplicationReleasesFromManifest("foo")`), which is not available anymore.

There is exactly the same test in cd-service, does the exact job but fetches data from the database.

I believe this is enough.

services/manifest-repo-export-service/Makefile
bqcuong: No, actually it is lower now: `42.5`

sven-urbanski-freiheit-com (approved):

PR #2906, opened 2026-05-13
fix: rollout-service edge cases

sven-urbanski-freiheit-com:
fixes several issues in the rollout-service:
* the option manageArgoApplications in helm is not compatible with bracket mode => not needed so it is forbidden in helm now.
* the GetVersion implementation was unstable because it derived the bracket mode from the env var => not it uses the request data only, and to make that easier the request now says "1:" instead of "1" so that we can just search for a ":" to know if bracket mode is enabled. ArgoCd is the reference here, we need to answer what it asks, independent of our configuration.
* a failing "push" to argocd was silently ignore, this is now handled
* several new tests were added


services/rollout-service/pkg/argo/argo.go
```diff
@@ -110,12 +110,12 @@ func (a *ArgoAppProcessor) Push(ctx context.Context, last *ArgoOverview) error {
...
 	select {
-	case a.trigger <- a.lastOverview:
+	case a.trigger <- last:
 		l.Info("argocd.pushed")
 		a.GaugeKuberpultEventsQueueFillRate(ctx)
 		return nil
-	default:
-		return fmt.Errorf("failed to push to argo app processor: channel full")
+	case <-ctx.Done():
+		return ctx.Err()
```
ccreinartz11, line 118: ❓ why don't we keep the default?

services/rollout-service/pkg/versions/versions.go
```diff
@@ -436,11 +436,10 @@ func (v *versionClient) ConsumeEvents(ctx context.Context, processor VersionEven
 
 			overview.AppDetails = appsToChange
 			if err := v.ArgoProcessor.Push(ctx, &overview); err != nil {
-				l.Sugar().Warnf("version.push failed: %v", err)
-			} else {
-				l.Info("version.push")
-				appsToChange = make(map[string]*api.GetAppDetailsResponse)
+				return err
```
ccreinartz11: 🔧 wrap with proper error message

ccreinartz11 (approved):

services/rollout-service/pkg/argo/argo.go
sven-urbanski-freiheit-com: before, we kept the default, and just logged it.
Now, by not keeping the default, we wait until we can put the element into the queue (`a.trigger<-last`) - or until the context is cancelled. This way we do not lose events.

PR #2910, opened 2026-05-18
feat: create and delete manifest-locks 

sven-urbanski-freiheit-com:
This handles creation and deletion of manifest locks, as well as documentation.

Ref: SRX-97UP2T


docs/users/6_locks.md
```diff
@@ -60,6 +60,58 @@ These work the same as environment locks, but are specific to one team alone.
 3) Click on the trash icon to delete it.
 4) Submit planned actions.
 
+## Manifest Locks
```
ccreinartz11, line 63: 🤔 Overall the section is really good. 
I think it should also contain a short section at the top that explains that manifest-locks are meant as an emergency measure and not as an everyday mechanism. You're hinting at it but it's not completely clear. 

Also, there should be a section (somewhere) that describes what behaviour is to be expected / what measures should be taken when the manifest lock is removed.

pkg/api/v1/api.proto
```diff
@@ -255,6 +257,18 @@ message DeleteEnvironmentTeamLockRequest {
   string lock_id = 3;
 }
 
+message CreateManifestLockRequest {
+  string app = 1;
+  string env = 2;
+  string message = 3;
+  optional string suggested_life_time = 4;
```
ccreinartz11, line 264: ❓ Do we really need it? It already seems unnecessary for the regular locks.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -1995,6 +1995,109 @@ func (c *DeleteEnvironmentTeamLock) Transform(
...
+	if !valid.EnvironmentName(c.Env) {
+		return "", grpc.InvalidArgument(ctx, fmt.Errorf("cannot create manifest lock: invalid environment name: '%s'", c.Env))
+	}
+	hasLock, err := state.DBHandler.DBHasActiveManifestLock(ctx, transaction, c.App, c.Env)
+	if err != nil {
+		return "", fmt.Errorf("CreateManifestLock: could not check for existing lock: %w", err)
+	}
+	if hasLock {
+		return "", grpc.FailedPrecondition(ctx, fmt.Errorf("manifest lock already exists for app %q on environment %q", c.App, c.Env))
+	}
```
ccreinartz11, line 2037: 🤔 Is that the expected behaviour? I would this would just be a noop without error. Or at least that this returns a special error type that we can filter for in the calling function. Because usually this is not an error  here. It could be an error in the FE or another calling function though.

services/cd-service/pkg/repository/transformer.go
```diff
@@ -1995,6 +1995,109 @@ func (c *DeleteEnvironmentTeamLock) Transform(
...
+	t TransformerContext,
+	transaction *sql.Tx,
+) (string, error) {
+	if !valid.ApplicationName(c.App) {
+		return "", grpc.InvalidArgument(ctx, fmt.Errorf("cannot delete manifest lock: invalid application name: '%s'", c.App))
+	}
+	if !valid.EnvironmentName(c.Env) {
+		return "", grpc.InvalidArgument(ctx, fmt.Errorf("cannot delete manifest lock: invalid environment name: '%s'", c.Env))
+	}
+	if err := state.DBHandler.DBDeleteManifestLock(ctx, transaction, c.App, c.Env); err != nil {
```
ccreinartz11, line 2095: can `state.DBHandler` be nil? In this case, add nil-check. I'd actually do that always to avoid nil pointer exceptions, but that should be in line with the rest of the codebase.

sven-urbanski-freiheit-com: We don't do that in the codebase.
We always have exactly 1 `DBHandler`, so it can't ever be `nil` in production code. If it was nil, nothing would work, and having a panic for that would be appropriate.
It saves us lots of nil checks overall.

pkg/api/v1/api.proto
sven-urbanski-freiheit-com: I think it's easier to leave it in than to make it different from the other locks.

services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: This can't be a no-op because creating a lock also requires a message and audit-data (author).
The UI already checks for this, and doesn't allow the user to create multiple.
So this check here is essentially just a fail-safe for the UI. I would like both UI and server-side to be consistent in their error checks.

docs/users/6_locks.md
sven-urbanski-freiheit-com: I added more explanations to both topics.

ccreinartz11 (approved):

PR #2914, opened 2026-05-26
fix: scripts for kind tests and general test setup

sven-urbanski-freiheit-com:
* store logs of tests runs into a file
* run-kind.sh is now re-entrant: it doesn't need to recreate the cluster all the time
* create-testdata: staging has upstream development and not development2.

Ref: SRX-VRKPNL


ccreinartz11 (approved):

charts/kuberpult/.gitignore
```diff
@@ -7,4 +7,5 @@ argocd-values.yml
 argocd_adminpw.txt
 Chart.lock
 emptyfile
-charts/*
\ No newline at end of file
+charts/*
+kind-test-logs/
```
sven-urbanski-freiheit-com:
```suggestion
kind-test-logs/

```

PR #2915, opened 2026-05-26
fix: improve kind-brackets tests

sven-urbanski-freiheit-com:
These are integration-level tests, that focus on the bracket migration.

The tests mostly work like this:

0) assume kind cluster is set up with kuberpult and argocd.
1) clean up cluster by truncating all tables, and scaling kuberpult to 0
2) install kuberpult helm chart with specific values (e.g. brackets disabled on dev)
3) deploy something and record deployment creation times
4) install kuberpult helm chart with other values (e.g. brackets enabled on dev)
5) wait for argo to reconcile
6) check that the recording deployment times (step 3) are still correct (deployment was not interrupted)

All tests come with a log setup, e.g. the test logger always prints the timestamp, and we dump a lot of pod logs (kuberpult/argocd) when a test fails.
portforward_test.go takes care that dropped port-forwards are quickly re-established.

Ref: SRX-VRKPNL

### DIFF
--- a/tests/kind-brackets/bracket_migration_test.go
+++ b/tests/kind-brackets/bracket_migration_test.go
@@ -701,13 +701,7 @@ func TestBracketMoveBetweenBrackets(t *testing.T) {
 
 // TestBracketMoveAndBack is the regression test for the scenario where an app
 // moves from bracket1 to bracket2 and then back to bracket1.
-//
-// The first move (b1→b2) may work due to lucky ArgoCD finalizer timing.
-// The second move (b2→b1) is more likely to fail because bracket1 must be
-// RECREATED (it was deleted after move 1), and bracket2's cascade-delete
-// finalizer may race against bracket1's re-creation and prune the pod.
-//
-// Both moves are checked independently so it is clear which transition fails.
+// Previously this had lead to deployment deletions.
 func TestBracketMoveAndBack(t *testing.T) {
 	cleanupCluster(t)
 	tLogf(t, "runSuffix: %s", runSuffix)
--- a/tests/kind-brackets/bracket_migration_test.go
+++ b/tests/kind-brackets/bracket_migration_test.go
@@ -18,26 +18,26 @@ package kindbracketstest
 
 import (
 	"context"
+	"fmt"
 	"os/exec"
 	"strings"
 	"testing"
 	"time"
 
-	auth "github.com/freiheit-com/kuberpult/pkg/auth"
 	api "github.com/freiheit-com/kuberpult/pkg/api/v1"
+	auth "github.com/freiheit-com/kuberpult/pkg/auth"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
 )
 
-const cdServiceGrpcAddr = "localhost:8083"
+const cdServiceGrpcAddr = "localhost:5004"
+const argoNamespace = "default"
 
-// helmUpgrade calls helm upgrade with the bracket staging cluster override and waits
-// for the rollout-service deployment to finish rolling out.
-// It calls helm directly (not install-kuberpult-helm.sh) to avoid port-forward side
-// effects and dependency rebuild overhead.
-func helmUpgrade(t *testing.T, stagingEnabled bool) {
+// helmUpgrade calls helm upgrade with the given bracket configuration and waits
+// for all services to finish rolling out.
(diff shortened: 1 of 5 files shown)


tests/kind-brackets/portforward_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"os/exec"
+	"sync"
+	"time"
+)
+
+// pfManager keeps a kubectl port-forward process alive. When the process exits
+// (pod replaced, network blip) the goroutine restarts it after a short guard
+// delay. Calling restart() kills the current process so the loop immediately
+// connects to the newly rolled-out pod.
+type pfManager struct {
```
ccreinartz11: what does `pf` stand for? For the struct itself I would recommend a type without abbreviations, unless the abbreviation is completely obvious.

ccreinartz11: ah, I saw further down. port-forward manager? In this case `portForwardManager` would the the name I'd go for.

tests/kind-brackets/bracket_migration_test.go
```diff
@@ -435,59 +415,51 @@ func TestBracketAddAppToExistingBracket(t *testing.T) {
...
+	cleanupCluster(t)
+	tLogf(t, "runSuffix: %s", runSuffix)
 	app1 := "bpu-app1-" + runSuffix
 	app2 := "bpu-app2-" + runSuffix
 	bracket := "bpu-bracket-" + runSuffix
 
-	t.Log("step 1: upgrade to staging=true (bracket mode)")
-	helmUpgrade(t, true)
+	tLog(t, "step 1: upgrade to staging=true (bracket mode)")
+	helmUpgrade(t, true, false, true, 50)
```
ccreinartz11: :nit: would be much easier to read if this function received a `helmUpgradeConfig` struct.

tests/kind-brackets/bracket_migration_test.go
```diff
@@ -630,91 +540,231 @@ func helmUpgradeWithBracketsConfig(t *testing.T, bracketsEnabled, developmentEna
...
+	})
+
+	tLog(t, "step 8: run staging release train (promotes v2, now in bracket2)")
+	releaseTrain(t, stagingNamespace)
+
+	tLog(t, "step 9: wait for bracket2 Argo app to appear on staging")
+	waitForArgoApp(t, "staging-"+bracket2)
+
+	tLog(t, "step 10: wait for v2 synced in staging")
+	waitForDeploymentAnnotation(t, stagingNamespace, app+"-bracket-dep", "2")
```
ccreinartz11, line 696: ❓ How do you ensure that these happen in the correct order?

tests/kind-brackets/bracket_migration_test.go
```diff
@@ -630,91 +540,231 @@ func helmUpgradeWithBracketsConfig(t *testing.T, bracketsEnabled, developmentEna
...
+	assertDeploymentCreationTimesStable(t, creationTimes, "move bracket1→bracket2")
+}
+
+// TestBracketMoveAndBack is the regression test for the scenario where an app
+// moves from bracket1 to bracket2 and then back to bracket1.
+//
+// The first move (b1→b2) may work due to lucky ArgoCD finalizer timing.
+// The second move (b2→b1) is more likely to fail because bracket1 must be
+// RECREATED (it was deleted after move 1), and bracket2's cascade-delete
+// finalizer may race against bracket1's re-creation and prune the pod.
```
ccreinartz11: 🤔 This sounds like this issue still isn't fixed.

sven-urbanski-freiheit-com: Yes, but it's just an outdated comment. I will rephrase

tests/kind-brackets/bracket_migration_test.go
sven-urbanski-freiheit-com: Not sure I understand the question.
We first check the argo app (waitForArgoApp) to appear, then we wait for the deployment annotation.
We look into the deployment to see if we find the correct release_version in there.

tests/kind-brackets/portforward_test.go
sven-urbanski-freiheit-com: done

tests/kind-brackets/bracket_migration_test.go
sven-urbanski-freiheit-com: done

tests/kind-brackets/bracket_migration_test.go
ccreinartz11: I thought it was important that the app appears before the deployment annotation?
Even if you wait in this order, that does not ensure that the resources appear in that order.

ccreinartz11 (approved):

PR #2917, opened 2026-05-28
fix: delayed bracket deletion between reposerver and argo

sven-urbanski-freiheit-com:
This changes the way we encode the ApplicationSource.Path in argo.
We now add "@<eslId>" so that we can easily reference the eslId in the reposerver-service. In that service, we must return the version that argo is asking from us, we can't just always return the newest version (e.g. of a bracket).
To achieve this, we add the transformerId to the bracket_history table.

When the rollout-service creates apps, it will use the latest bracket_history table to get the transformerId and write that into the argo app (ApplicationSource.Path).
When the reposerver-service is asked to return a bracket, it will determine the version based on the added transformerId in the ApplicationSource.Path, instead of just assuming that the latest version is ok.

The latest version is specifically not ok when we move apps between brackets. Without this change, it could happen that argo asks the reposerver very early what version is deployed, and it would therefore delete a bracket that was still in use.

Also added a kind-test "TestBracketDeploymentAndBracketMoveSimultaneous", and lots of documentation to argo.go.

Ref: SRX-VRKPNL

### DIFF
--- a/services/reposerver-service/pkg/reposerver/reposerver.go
+++ b/services/reposerver-service/pkg/reposerver/reposerver.go
@@ -123,17 +123,57 @@ type bracketResult struct {
 	revision  string
 }
 
+// parseBracketSegment splits the 4th path component into a bracket name and an
+// optional brackets_history.source_transformer_esl_id. The rollout-service appends
+// "@<esl_id>" so the reposerver can read the exact snapshot the Argo CD app was
+// last spec-updated against, even if a newer brackets_history row has already
+// arrived but the rollout-service has not yet refreshed the Argo CD app spec.
+//
+// Returns isLegacyFormat=true when the segment carries no "@<esl_id>" suffix.
+// In that case the caller must fall back to DBSelectBracketHistoryLatest and the
+// returned eslId is meaningless (zero). This handles pre-upgrade Argo CD apps
+// that were created before the rollout-service learned to embed the esl_id;
+// they keep working until the rollout-service refreshes their spec.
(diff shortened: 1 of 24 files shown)


services/reposerver-service/pkg/reposerver/reposerver.go
```diff
@@ -123,17 +123,57 @@ type bracketResult struct {
...
+		return types.ArgoBracketName(segment), 0, true, nil
+	}
+	name := segment[:at]
+	idStr := segment[at+1:]
+	parsed, err := strconv.ParseInt(idStr, 10, 64)
+	if err != nil {
+		return "", 0, false, fmt.Errorf("generateBracketManifest: malformed esl_id in path segment '%s': %w", segment, err)
+	}
+	return types.ArgoBracketName(name), db.TransformerID(parsed), false, nil
+}
```
ccreinartz11, line 149: 🤔 Can we force a spec-update on all services so that we don't have to worry about edge cases for "old" entries in the table?

ccreinartz11 (approved):

services/reposerver-service/pkg/reposerver/reposerver.go
sven-urbanski-freiheit-com: I don't think there's an easy way to do this. I mean, we could spec-update all apps on a rollout-service restart, but during that time the reposerver services (reposerver) would already receive requests.

But with the new "@" character there is a clear separator between legacy and new, I think that is solid.

PR #2918, opened 2026-05-28
fix: add rollout_should_undeploy_cascade table

sven-urbanski-freiheit-com:
Ref: SRX-VRKPNL

### DIFF
--- a/pkg/db/db_brackets_history.go
+++ b/pkg/db/db_brackets_history.go
@@ -209,6 +209,26 @@ func executeSelectQuery(ctx context.Context, tx *sql.Tx, selectQuery string, arg
 	return result, nil
 }
 
+// DBSelectBracketHistoryPrevious returns the bracket snapshot immediately before
+// the latest one (the second-newest row). It is used to detect which bracket an
+// app just left: comparing the previous snapshot with the latest reveals brackets
+// that became empty, so the rollout-service can delete their now-orphaned Argo app.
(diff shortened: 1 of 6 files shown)


pkg/db/db_brackets_history.go
```diff
@@ -209,6 +209,26 @@ func executeSelectQuery(ctx context.Context, tx *sql.Tx, selectQuery string, arg
 	return result, nil
 }
 
+// DBSelectBracketHistoryPrevious returns the bracket snapshot immediately before
+// the latest one (the second-newest row). It is used to detect which bracket an
+// app just left: comparing the previous snapshot with the latest reveals brackets
+// that became empty, so the rollout-service can delete their now-orphaned Argo app.
+func DBSelectBracketHistoryPrevious(ctx context.Context, h *DBHandler, tx *sql.Tx) (result *BracketRow, err error) {
```
ccreinartz11, line 216: ❓ What happens if I change the bracket twice in quick succession?
Could this happen:

Initial condition: App in Bracket 1 (short: 'b1')
Switch to 'b2' then very quickly
Switch to 'b3'.

Now, the asynchronous cleanup starts and sees: 'current b3, previous b2'. 'b1' is not cleaned up.

ccreinartz11: probably works fine but maybe add a test for it just in case.

ccreinartz11 (approved):

PR #2920, opened 2026-05-29
fix(rollout): undeploy process as background job

sven-urbanski-freiheit-com:
Ref: SRX-VRKPNL

### DIFF
--- a/services/rollout-service/pkg/argo/argo.go
+++ b/services/rollout-service/pkg/argo/argo.go
@@ -38,6 +38,7 @@ import (
 	"github.com/freiheit-com/kuberpult/pkg/conversion"
 	"github.com/freiheit-com/kuberpult/pkg/logger"
 	"github.com/freiheit-com/kuberpult/pkg/setup"
+	"github.com/freiheit-com/kuberpult/pkg/sorting"
 )
 
 // this is a simpler version of ApplicationServiceClient from the application package
@@ -191,7 +192,8 @@ type ArgoOverview struct {
 
 func (a *ArgoAppProcessor) ProcessArgoOverview(ctx context.Context, l *zap.Logger, argoOv *ArgoOverview) {
 	overview := argoOv.Overview
-	for currentApp, currentAppDetails := range argoOv.AppDetails {
+	for _, currentApp := range sorting.SortKeys(argoOv.AppDetails) {
+		currentAppDetails := argoOv.AppDetails[currentApp]
 		span, ctx := tracer.StartSpanFromContext(ctx, "ProcessChangedApp")
 		defer span.Finish()
 		span.SetTag("kuberpult-app", currentApp)
@@ -211,7 +213,7 @@ func (a *ArgoAppProcessor) ProcessArgoOverview(ctx context.Context, l *zap.Logge
 							ArgoEnvironmentConfiguration: cfg,
 							IsBracket:                    isBracket,
 						}
-						a.ProcessAppChange(ctx, appInfo, currentAppDetails, overview)
+						a.ProcessAppChange(ctx, appInfo, currentAppDetails, overview, argoOv.AppDetails)
 					}
 				} else {
 					appInfo := &AppInfo{
@@ -222,7 +224,7 @@ func (a *ArgoAppProcessor) ProcessArgoOverview(ctx context.Context, l *zap.Logge
 						ArgoEnvironmentConfiguration: parentEnvironment.Config.Argocd,
 						IsBracket:                    isBracket,
 					}
-					a.ProcessAppChange(ctx, appInfo, currentAppDetails, overview)
+					a.ProcessAppChange(ctx, appInfo, currentAppDetails, overview, argoOv.AppDetails)
 				}
 
 			}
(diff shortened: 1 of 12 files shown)


services/rollout-service/pkg/argo/argo.go
```diff
@@ -82,22 +94,38 @@ type ArgoAppProcessor struct {
...
+// MaxProcessedTransformerEslId returns a pointer to the atomic that tracks the
+// highest transformer ESL ID for which ProcessArgoOverview has fully completed.
+// The cascade consumer reads this to ensure it never cascade-deletes a bracket
+// before the corresponding new bracket Argo Application has been created.
+func (a *ArgoAppProcessor) MaxProcessedTransformerEslId() *atomic.Int64 {
+	return a.maxProcessedTransformerEslId
+}
+
+// updateMaxProcessedEslId advances the max-processed ESL ID if eslId is larger.
+// Only called from Consume (single writer), so a plain Load+Store is safe.
```
ccreinartz11, line 124: ❓ The `rollout-service` always runs with a single replica, correct? Otherwise this would not be safe.

services/rollout-service/pkg/argo/argo.go
```diff
@@ -259,12 +282,20 @@ func (a *ArgoAppProcessor) deleteAppNoCascade(ctx context.Context, knownApps map
...
+// not yet been received (KnownApps cache is stale after rollout-service restart).
+func (a *ArgoAppProcessor) deleteAppNoCascadeByName(ctx context.Context, argoAppName string) error {
+	logger.FromContext(ctx).Info("bracket.delete.no-cascade.by-name",
+		zap.String("argo.app", argoAppName))
+	err := DeleteApplication(ctx, a.ApplicationClient, argoAppName, false)
+	if err != nil && status.Code(err) != codes.NotFound {
+		return err
+	}
+	return nil
 }
```
ccreinartz11, line 302: ❓ Clarify

services/rollout-service/pkg/undeploy/undeploy.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+		batch = rows
+		return nil
+	})
+	if err != nil {
+		return 0, fmt.Errorf("undeploy: read batch: %w", err)
+	}
+	maxProcessed := maxProcessedTransformerEslId.Load()
+	for _, row := range batch {
+		if int64(row.NotBeforeTransformerEslId) > maxProcessed {
```
ccreinartz11: 🤔 The naming in this line is very confusing to me. Let's discuss.

ccreinartz11: also, shouldn't this be >=?

services/rollout-service/pkg/undeploy/undeploy.go
```diff
@@ -0,0 +1,203 @@
...
+			// pick it up.
+			continue
+		}
+		// Safety check: skip if the app/bracket has been redeployed since this row
+		// was written (e.g. the rollout-service was offline, the app was undeployed
+		// AND redeployed, and now the stale row would destroy the live workload).
+		stalePtr, staleErr := db.WithTransactionT(dbHandler, ctx, 1, true, func(ctx context.Context, tx *sql.Tx) (*bool, error) {
+			s, err := dbHandler.IsCascadeRowStale(ctx, tx, row.ArgoApp, row.Env, row.IsBracket)
+			return &s, err
+		})
```
ccreinartz11, line 130: 🤔 Let's discuss edge cases
What if this is an outdated bracked that was redeployed? Would it then exist forever?

services/rollout-service/pkg/undeploy/undeploy.go
```diff
@@ -0,0 +1,203 @@
...
+			l.Error("rollout.undeploy.cascade.row-delete-failed", zap.Error(dbErr))
+		}
+	case status.Code(err) == codes.NotFound:
+		// Argo Application is already gone — idempotent success.
+		l.Info("rollout.undeploy.cascade.already-gone")
+		if dbErr := deleteRow(ctx, dbHandler, row); dbErr != nil {
+			l.Error("rollout.undeploy.cascade.row-delete-failed", zap.Error(dbErr))
+		}
+	default:
+		// Transient. Bump attempts; drop the row once the budget is exhausted.
```
ccreinartz11, line 175: ❓ Is this safe?

services/rollout-service/pkg/argo/argo.go
sven-urbanski-freiheit-com: we'll try with maxSurge=0, or make it a statefulset

services/rollout-service/pkg/undeploy/undeploy.go
ccreinartz11: Suggestion: `row.MaxProcessedTransformerEslId`

sven-urbanski-freiheit-com: I went for another name ("gating_transformer_esl_id" because it acts as a gate), and added comments e.g. in the sql file.

services/rollout-service/pkg/argo/argo.go
sven-urbanski-freiheit-com: Actually, I don't see any scenario where this is an actual problem.
But just in case, I changed the deployment strategy to maxSurge=0.

services/rollout-service/pkg/undeploy/undeploy.go
sven-urbanski-freiheit-com: it should be `>` because this is the skip-path. So if they are equal we do process the event.
This is safe, because when receiving the event, the rollout-service first does the create/update operations, and then sets the atomic maxProcessedTransformerEslId.

ccreinartz11 (approved):

PR #2921, opened 2026-06-01
fix: Re-render root-app after deleteEnvFromApp

sven-urbanski-freiheit-com:
This fixes 3 issues related to rendering after deleting an app from an env:

1) Render only brackets that have at least one app with deployment in it.
2) The deployment-history was not appended when deleting an env from app (db_deployments).
3) The DeleteEnvFromApp Transformer falsly exited with an error when a specific directory did not exist.

Ref: SRX-PI9NJN

### DIFF
--- a/pkg/db/db_deployments.go
+++ b/pkg/db/db_deployments.go
@@ -745,6 +745,32 @@ func (h *DBHandler) DBDeleteDeployment(ctx context.Context, tx *sql.Tx, appName 
 	return nil
 }
 
+// DBDeleteDeploymentWithHistory removes the current deployment for the given app/env and records
+// the un-deployment in the append-only deployments_history table (a row with a nil release
+// version). Callers that delete a deployment should use this instead of DBDeleteDeployment, so
+// timestamped history queries (e.g. DBSelectAppsWithDeploymentInEnvAtTimestamp) reflect the
+// deletion at the transformer's timestamp. The metadata records who performed the deletion.
+func (h *DBHandler) DBDeleteDeploymentWithHistory(ctx context.Context, tx *sql.Tx, appName types.AppName, envName types.EnvName, transformerID TransformerID, metadata DeploymentMetadata) (err error) {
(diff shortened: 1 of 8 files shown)


pkg/db/db_deployments.go
```diff
@@ -745,6 +745,32 @@ func (h *DBHandler) DBDeleteDeployment(ctx context.Context, tx *sql.Tx, appName
 	return nil
 }
 
+// DBDeleteDeploymentWithHistory removes the current deployment for the given app/env and records
+// the un-deployment in the append-only deployments_history table (a row with a nil release
+// version). Callers that delete a deployment should use this instead of DBDeleteDeployment, so
+// timestamped history queries (e.g. DBSelectAppsWithDeploymentInEnvAtTimestamp) reflect the
+// deletion at the transformer's timestamp. The metadata records who performed the deletion.
```
ccreinartz11, line 752: Makes sense. Do you already adjust all callers (I can see some, just curious if that's all of them)?

pkg/db/db_deployments.go
```diff
@@ -745,6 +745,32 @@ func (h *DBHandler) DBDeleteDeployment(ctx context.Context, tx *sql.Tx, appName
...
+		Created: time.Time{},
+		App:     appName,
+		Env:     envName,
+		ReleaseNumbers: types.ReleaseNumbers{
+			Version:  nil,
+			Revision: 0,
+		},
+		Metadata:      metadata,
+		TransformerID: transformerID,
+	})
```
ccreinartz11, line 771: ❓ Shouldn't these two be in a transaction?

pkg/db/db_deployments.go
sven-urbanski-freiheit-com: yes I adjusted all callers, there were not that many.

pkg/db/db_deployments.go
sven-urbanski-freiheit-com: They are, see parameter "tx"

ccreinartz11 (approved):

PR #2925, opened 2026-06-08
fix(tests): more kind tests

sven-urbanski-freiheit-com:
Adds more kind tests about first installing an old helm-chart to ensure the migration works.

Ref: SRX-PQRQ5V

### DIFF
--- a/tests/kind-brackets/bracket_migration_test.go
+++ b/tests/kind-brackets/bracket_migration_test.go
@@ -40,6 +40,7 @@ type helmUpgradeParams struct {
 	developmentEnabled bool
 	stagingEnabled     bool
 	channelSize        int
+	oldVersion         string // if non-empty, installs this specific chart version
 }
 
 // helmUpgrade restarts the port-forward, then delegates to the shared HelmUpgrade.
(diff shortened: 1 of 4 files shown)


tests/kind-brackets/bracket_migration_test.go
```diff
@@ -40,6 +40,7 @@ type helmUpgradeParams struct {
 	developmentEnabled bool
 	stagingEnabled     bool
 	channelSize        int
+	oldVersion         string // if non-empty, installs this specific chart version
```
ccreinartz11, line 43: ❓ The name does not tell me why this variable is needed / how I would configure it.

ccreinartz11: maybe `remoteVersion` or similar?

ccreinartz11 (approved):

tests/kind-brackets/bracket_migration_test.go
sven-urbanski-freiheit-com: I'll fix that in a later PR

PR #2927, opened 2026-06-08
fix(rollout): movement between brackets

sven-urbanski-freiheit-com:
The cd-service now also sends lost bracket members to the rollout.

The rollout-service uses that to
1) add new bracket members first
2) wait until argo reports they are applied
3) then remove old bracket members

This way there is no downtime in between.

Ref: SRX-PQRQ5V


pkg/api/v1/api.proto
```diff
@@ -487,6 +487,11 @@ message BracketDeployment {
 message GetBracketDetailsResponse {
   string bracket_name = 1;
   map<string, BracketDeployment> deployments = 2;
+  // Names of brackets that gained some members that this bracket lost in this change.
+  // The rollout-service must not refresh this bracket's Argo app spec (which
+  // would prune the moved apps' resources) until each named bracket has
+  // adopted them (reported Synced by Argo CD).
+  repeated string lost_members_to = 3;
```
ccreinartz11, line 494: 🤔 This is getting so complicated. 
At the very least this needs an explicit example like "Bracked A had App abc before the change and then..."

ccreinartz11 (approved):

pkg/api/v1/api.proto
sven-urbanski-freiheit-com: Agreed. I will add that in the next PR.

PR #2941, opened 2026-06-26
fix: ignore deprecated files and fix deletion of files

sven-urbanski-freiheit-com:
This fixes several issues:
* TestUndeployLogic only checked the error diff, if err!=nil => now this happens always
* UndeployApplication had outdated logic that failed on certain old files like "version" => Now it just deletes all files without condition (app & bracket)
* DeleteEnvFromApp now deletes the bracket
* deleteBracketOfAppEnv now uses the given transaction, instead of opening a new one
* (style issue) Replaced some app variables of type "string" with types.AppName

Ref: SRX-FMUCRQ

### DIFF
--- a/services/manifest-repo-export-service/pkg/repository/transformer.go
+++ b/services/manifest-repo-export-service/pkg/repository/transformer.go
@@ -86,8 +86,8 @@ func applicationDirectory(fs billy.Filesystem, application string) string {
 }
 
 // environmentApplicationDirectory returns environments/<env>/applications/<app>/
-func environmentApplicationDirectory(fs billy.Filesystem, environment types.EnvName, application string) string {
-	return fs.Join("environments", string(environment), "applications", application)
+func environmentApplicationDirectory(fs billy.Filesystem, environment types.EnvName, application types.AppName) string {
+	return fs.Join("environments", string(environment), "applications", string(application))
 }
 
 // writeManifests writes the manifests.yaml file to the given parent path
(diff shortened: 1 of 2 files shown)


services/manifest-repo-export-service/pkg/repository/transformer.go
```diff
@@ -1610,11 +1610,39 @@ func (c *DeleteEnvFromApp) Transform(
...
+			zap.String("bracketPath", dir.BracketPath),
+			zap.String("bracketDir", dir.BracketDirectory),
+		)
+		err = fs.Remove(dir.BracketPath)
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			return fmt.Errorf("deleteBracketOfAppEnv: error removing bracket path %s: %v", dir.BracketPath, err)
+		}
+	}
+	return nil
+}
```
ccreinartz11, line 1644: This removes the app manifest from the bracket path. Do we also remove the app from the bracket-manifest (the one that lists all of the contained apps)?

ccreinartz11 (approved):

services/manifest-repo-export-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: I think you're referring to the directory argo/v1alpha/<env>.yaml.
This file would also be written in this context, but that is taken care of not inside a transformer, but in the function `afterTransform` (not changed in this PR).

The brackets themselves are not files, but directories, e.g. this one: https://github.com/freiheit-com/fdc-standard-setup-dev-env-manifest/tree/main/environments/development/brackets
So there is not directly a "bracket manifest".

PR #2944, opened 2026-07-01
fix(ci): add helm-test to PR pipeline

sven-urbanski-freiheit-com:
Previously the helm-test was only run in the release-pipeline, so only after merging.

Ref: SRX-UREFYC


.github/workflows/run-helm-test.yml
```diff
@@ -0,0 +1,15 @@
...
+  run-helm-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Run helm chart tests
+        run: |
+          IMAGE_TAG_KUBERPULT="$(make pr-version)" make -C charts/kuberpult test-helm
```
bqcuong: Just a question, do we also run test-helm for rebuild.yml, and release.yml? Can we make use of this new run-helm-test.yml there?

bqcuong (approved):

.github/workflows/run-helm-test.yml
sven-urbanski-freiheit-com: done

PR #2951, opened 2026-07-06
Link to Brackets in ArgoUI when bracket rendering is enabled

ccreinartz11:


charts/kuberpult/templates/frontend-service.yaml
```diff
@@ -99,6 +99,8 @@ spec:
           value: {{ .Values.git.author.email | quote }}
         - name: KUBERPULT_CDSERVER
           value: kuberpult-cd-service:8443
+        - name: KUBERPULT_ROOT_APPS_POINT_TO_BRACKETS
```
bqcuong, line 102: Can we also have this for local dev? i.e., in docker-compose.yml?

bqcuong (approved):

PR #2953, opened 2026-07-06
fix: add CI reliability test for integration-test

bqcuong:
Ref: [SRX-S66VRR](https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/STORY/-OnVvaSOsN2_srLWsRr5/)

This reliability test intentionally modifies the Postgres port during integration testing. As a result, all integration tests are expected to fail or timeout, which should ultimately cause the entire CI pipeline to fail.


.github/workflows/ci-reliability.yml
```diff
@@ -79,6 +83,22 @@ jobs:
 
             modify_readme: false
             inject_linting_error: true
+            inject_integration_tests_error: false
+
+            check_new_images: false
+            expect_cd: false
+            expect_fe: false
+            expect_export: false
```
ccreinartz11, line 91: ❓ What are these changes to id: "5"? Why are they part of this PR?

.github/workflows/integration-tests.yml
```diff
@@ -7,6 +7,11 @@ on:
         required: true
         type: string
         description: 'The trigger which calls the workflow, one of [pull-request, main, tag]'
+      ref:
+        required: false
+        default: ""
+        type: string
+        description: "The ref to use for the checkout"
```
ccreinartz11, line 14: ❓ Why does this change require that you pass a 'ref' to a specific version?

.github/workflows/ci-reliability.yml
bqcuong: These lines are parts of test-id: 7, git diff just shows them wrong.
See new lines: 103-106, they are the lines that I actually added.

.github/workflows/integration-tests.yml
bqcuong: In this case of reliability test, `ref` should to point to https://github.com/freiheit-com/kuberpult/tree/automated/ci-reliability/ci-test-7, on which we injects the error for integration tests.

With out this `ref`, the integration-test pipeline will run on either current PR or main branch. See line 37 below.

ccreinartz11 (approved):

PR #2962, opened 2026-07-08
fix: remove the max batch limit

pedrogameiro:
After some practical experience with using the batch feature, it became clear that the overhead of batching is smaller than I expected. With no measurable increase of latency even for big values.

### DIFF
--- a/services/manifest-repo-export-service/pkg/cmd/server.go
+++ b/services/manifest-repo-export-service/pkg/cmd/server.go
@@ -52,9 +52,6 @@ const (
 	maxReleaseVersionsLimit = 30
 
 	maxEslProcessingTimeSeconds int64 = 600 // see eslProcessingIdleTimeSeconds in values.yaml
-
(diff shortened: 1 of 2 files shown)


services/manifest-repo-export-service/pkg/cmd/server.go
```diff
@@ -52,9 +52,6 @@ const (
 	maxReleaseVersionsLimit = 30
 
 	maxEslProcessingTimeSeconds int64 = 600 // see eslProcessingIdleTimeSeconds in values.yaml
-
-	// maxExportBatchSizeLimit is a hard upper bound on KUBERPULT_MAX_EXPORT_BATCH_SIZE.
-	maxExportBatchSizeLimit = 100
```
sven-urbanski-freiheit-com, line 57: I would prefer to increase the maximum (.e.g 1k) threshold instead of removing it. That way we still have a sanity check, and 1k should be enough for everyone.

PR #2963, opened 2026-07-08
feat: Add 'allowBracketsMove' helm parameter and wire it through services

ccreinartz11:
[SRX-I5P0IZ](https://revolution.dev/app/-JqFGExX46gs9mH7vxR5/-M37cZx7XYoOvQ5Mmcir/STORY/-OwNSIV2KqRjysNp8sSj/)

### DIFF
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -268,7 +268,7 @@ manifestRepoExport:
   #   - Brackets are rendered with prune=false, so Argo CD does NOT automatically delete resources.
   #
   # NOTE: When allowBracketMove==true, following a bracket move you must manually trigger a
-  # "re-render everything" after resetting to allowBracketMove==false for the affected environment 
+  # "re-render everything" after resetting to allowBracketMove==false for all affected environments 
   # to remove the app from its old bracket. Argo CD will not prune it automatically.
   allowBracketMove: false
 
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -250,6 +250,26 @@ manifestRepoExport:
     # If true, `experimentalRenderBrackets` must be true.
     experimentalRootAppsPointToBrackets: false
 
+  # Controls whether apps are allowed to move between brackets.
+  # Moving an app between brackets is a complicated operation: the app's manifest has to be removed
+  # from its old bracket and added to the new one and it has to be ensured that the app is managed by at 
+  # least one of the brackets at all times.
+  # We deliberately do not automate the full bracket-move logic, instead relying on a manual step during
+  # bracket moves to ensure that the move logic remains simple and reliable.
+  #
+  # allowBracketMove==false (default):
+  #   - The /api/release endpoint disallows bracket changes (returns HTTP 422). Assigning a bracket
+  #     to an app that has none yet, and creating a new app in a new or existing bracket, are always allowed.
+  #   - Brackets are rendered with prune=true.
+  # allowBracketMove==true:
+  #   - The /api/release bracket-change check is skipped; all bracket changes are allowed.
+  #   - Brackets are rendered with prune=false, so Argo CD does NOT automatically delete resources.
+  #
+  # NOTE: When allowBracketMove==true, following a bracket move you must manually trigger a
+  # "re-render everything" after resetting to allowBracketMove==false for the affected environment 
+  # to remove the app from its old bracket. Argo CD will not prune it automatically.
+  allowBracketMove: false
+
   # The initial time in seconds that this service waits when it does not find esl events to process, before trying again.
   # Time increases exponentially (factor 2.0) if errors occur, depending on the type of error.
   # Maximum time to wait is 600sec (cannot be changed).
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -258,8 +258,10 @@ manifestRepoExport:
   # bracket moves to ensure that the move logic remains simple and reliable.
   #
   # allowBracketMove==false (default):
-  #   - The /api/release endpoint disallows bracket changes (returns HTTP 422). Assigning a bracket
-  #     to an app that has none yet, and creating a new app in a new or existing bracket, are always allowed.
+  #   - The /api/release endpoint disallows changing an existing app's bracket (returns HTTP 422).
+  #     Every app already has a bracket (its own name if none is set explicitly), so assigning a real
+  #     bracket to an app that used its default bracket is also a move and is rejected. Only creating a
+  #     new app and putting it into a new or existing bracket is always allowed.
   #   - Brackets are rendered with prune=true.
   # allowBracketMove==true:
   #   - The /api/release bracket-change check is skipped; all bracket changes are allowed.
--- a/charts/kuberpult/values.yaml
+++ b/charts/kuberpult/values.yaml
@@ -257,20 +257,20 @@ manifestRepoExport:
   # We deliberately do not automate the full bracket-move logic, instead relying on a manual step during
   # bracket moves to ensure that the move logic remains simple and reliable.
   #
-  # allowBracketMove==false (default):
+  # allowBracketMoves==false (default):
   #   - The /api/release endpoint disallows changing an existing app's bracket (returns HTTP 422).
   #     Every app already has a bracket (its own name if none is set explicitly), so assigning a real
   #     bracket to an app that used its default bracket is also a move and is rejected. Only creating a
   #     new app and putting it into a new or existing bracket is always allowed.
   #   - Brackets are rendered with prune=true.
-  # allowBracketMove==true:
+  # allowBracketMoves==true:
   #   - The /api/release bracket-change check is skipped; all bracket changes are allowed.
   #   - Brackets are rendered with prune=false, so Argo CD does NOT automatically delete resources.
   #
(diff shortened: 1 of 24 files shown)


docs/operators/endpoint-release.md
```diff
@@ -96,3 +96,24 @@ Only has an effect when using the manifest-repo-export-service.
...
+  * Assigning a bracket to an app that has none yet, and creating a new app in a new or existing bracket,
+    are always allowed.
+  * Brackets are rendered with `prune: true`.
+* `allowBracketMove: true`:
+  * The bracket-change check is skipped; all bracket changes are allowed.
+  * Brackets are rendered with `prune: false`, so Argo CD does **not** automatically delete the resources
+    left behind by a move.
+  * **Operator action required:** after a bracket move you must manually trigger a "re-render everything"
+    for the affected environment to remove the app from its old bracket. Argo CD will not prune it
+    automatically.
```
ccreinartz11: 🤔 Not really sure that this belongs here, since it's a helm setting. 
But it also affects the behavior of the release endpoint...

Maybe a shorter description that only explain how the endpoint behaves based on the setting?

pkg/db/db_brackets_history.go
```diff
@@ -70,16 +70,23 @@ func fromJson(data []byte) (BracketJsonBlob, error) {
...
+func ResolveBracketName(app types.AppName, bracketName types.ArgoBracketName) types.ArgoBracketName {
+	if bracketName == "" {
+		return types.ArgoBracketName(app)
 	}
+	return bracketName
+}
+
+// HandleBracketsUpdate assigns app to newBracketName in the bracket history, removing it from any
+// previous bracket. newBracketName must already be resolved (see ResolveBracketName).
+func HandleBracketsUpdate(ctx context.Context, h *DBHandler, tx *sql.Tx, app types.AppName, newBracketName types.ArgoBracketName, now time.Time, transformerEslId TransformerID) error {
```
ccreinartz11: 🤔 This function does not really handle the full brackets update, because it does not update the bracket column on the 'current' app.

Renaming would be the easy improvement.
Alternatively, moving the update of the current app bracket column in here would be an option that keeps the bracket update atomic (ish)

services/cd-service/pkg/repository/transformer.go
ccreinartz11, line 578: We are duplicating checks here with `HandleBracketsUpdate`. 
It's functionally correct but seems overly complicated (multiple if-branches that are on the same 'happy path').

charts/kuberpult/values.yaml
```diff
@@ -250,6 +250,28 @@ manifestRepoExport:
...
+  #     Every app already has a bracket (its own name if none is set explicitly), so assigning a real
+  #     bracket to an app that used its default bracket is also a move and is rejected. Only creating a
+  #     new app and putting it into a new or existing bracket is always allowed.
+  #   - Brackets are rendered with prune=true.
+  # allowBracketMove==true:
+  #   - The /api/release bracket-change check is skipped; all bracket changes are allowed.
+  #   - Brackets are rendered with prune=false, so Argo CD does NOT automatically delete resources.
+  #
+  # NOTE: When allowBracketMove==true, following a bracket move you must manually trigger a
+  # "re-render everything" after resetting to allowBracketMove==false for the affected environment
```
sven-urbanski-freiheit-com:
```suggestion
  # "re-render everything" after resetting to allowBracketMove==false for all affected environments 
```

charts/kuberpult/values.yaml
```diff
@@ -250,6 +250,28 @@ manifestRepoExport:
...
+  #
+  # allowBracketMove==false (default):
+  #   - The /api/release endpoint disallows changing an existing app's bracket (returns HTTP 422).
+  #     Every app already has a bracket (its own name if none is set explicitly), so assigning a real
+  #     bracket to an app that used its default bracket is also a move and is rejected. Only creating a
+  #     new app and putting it into a new or existing bracket is always allowed.
+  #   - Brackets are rendered with prune=true.
+  # allowBracketMove==true:
+  #   - The /api/release bracket-change check is skipped; all bracket changes are allowed.
+  #   - Brackets are rendered with prune=false, so Argo CD does NOT automatically delete resources.
```
sven-urbanski-freiheit-com, line 268:
```suggestion
  #   - Brackets are rendered with prune=false, so Argo CD does NOT automatically delete resources - but still syncs changes that are not deletions.
```

services/frontend-service/pkg/handler/release.go
```diff
@@ -605,6 +605,11 @@ func writeCorrespondingResponse(ctx context.Context, w http.ResponseWriter, r *h
 			jsonBlob, err := json.Marshal(firstResponse)
 			writeReleaseResponse(w, r, jsonBlob, err, http.StatusBadRequest)
 		}
+	case *api.CreateReleaseResponse_BracketMoveNotAllowed:
+		{
+			jsonBlob, err := json.Marshal(firstResponse)
+			writeReleaseResponse(w, r, jsonBlob, err, http.StatusUnprocessableEntity)
```
sven-urbanski-freiheit-com, line 611: Please add a bit more details here, StatusUnprocessableEntity may be the right code, but it doesn't really explain what's happening.
I would expect that the response mentions in a message that this operation is not allowed when the new helm option is on.

docs/operators/endpoint-release.md
```diff
@@ -96,3 +96,26 @@ Only has an effect when using the manifest-repo-export-service.
...
+    name is used. Assigning a real bracket to an app that previously used its default (app-name) 
+    bracket is therefore also a move and is rejected.
+  * Creating a **new** app and putting it into a new or existing bracket is always allowed.
+  * Brackets are rendered with `prune: true`.
+* `allowBracketMove: true`:
+  * The bracket-change check is skipped; all bracket changes are allowed.
+  * Brackets are rendered with `prune: false`, so Argo CD does **not** automatically delete the resources
+    left behind by a move.
+  * **Operator action required:** after a bracket move you must manually trigger a "re-render everything"
+    for the affected environment to remove the app from its old bracket. Argo CD will not prune it
```
sven-urbanski-freiheit-com: this seems rather redundant now with the values.yaml.
I would suggest that here you just mention that this option exists for a special case of using brackets (brackets enabled), and allowBracketMove.
The rest could then be details only described in values.yaml.

pkg/db/db_brackets_history.go
sven-urbanski-freiheit-com: I think renaming would be a decent option

services/cd-service/pkg/cmd/server.go
```diff
@@ -152,6 +153,7 @@ func parseEnvVars() (_ *Config, err error) {
 
 	c.AllowLongAppNames = valid.ReadEnvVarBoolWithDefault("KUBERPULT_ALLOW_LONG_APP_NAMES", false)
 	c.ArgoCdGenerateFiles = valid.ReadEnvVarBoolWithDefault("KUBERPULT_ARGO_CD_GENERATE_FILES", true)
+	c.AllowBracketMove = valid.ReadEnvVarBoolWithDefault("KUBERPULT_ALLOW_BRACKET_MOVE", false)
```
sven-urbanski-freiheit-com: nitpick: Plural would be the proper option here, so "AllowBracketMoves". This would also be consistent with `AllowLongAppNames` for example.

services/cd-service/pkg/repository/transformer.go
sven-urbanski-freiheit-com: I think we should do the `DBSelectApp` once at the beginning of the transformer

services/manifest-repo-export-service/pkg/argocd/render_test.go
```diff
@@ -241,6 +242,56 @@ spec:
...
+    name: foo
+  project: dev
+  sources:
+  - path: environments/dev/applications/app1/manifests
+    repoURL: example.com/github
+    targetRevision: main
+  syncPolicy:
+    automated:
+      allowEmpty: true
+      prune: false
```
sven-urbanski-freiheit-com, line 283: I'm pretty sure automated.prune=false will work just as well, but just so you're aware, in my test I used:
```
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
      allowEmpty: true
    syncOptions:
      - Prune=false
```
syncoptions.prune=false  here overwrites automated.prune=true.

services/cd-service/pkg/repository/transformer.go
ccreinartz11: I think if we touch it we have to go beyond and clean this up properly once. 
Just moving this up might also have side effects that we don't see now and I'd rather avoid this.
And considering our idea for maintenance mode for kubepult it might not be worth it at this point.

services/frontend-service/pkg/handler/release.go
ccreinartz11: that's already covered by the `jsonBlob`, which is passed through. that also matches the pattern for the other status codes.

ccreinartz11: But I did extend the original message a bit.

services/manifest-repo-export-service/pkg/argocd/render_test.go
ccreinartz11: I think automated.prune should be better because it leaves more option for manual intervention while ensuring the same functionality if only kuberpult is used.

sven-urbanski-freiheit-com (approved):

PR #2966, opened 2026-07-10
fix: resolve NPE warnings for go code

bqcuong:
Ref: SRX-O0CH51

### DIFF
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -455,6 +455,7 @@ func MeasureGitSyncStatus(unsyncedApps, syncFailedApps int) error {
 func (r *repository) notifyChangedApps(changes *TransformerResult) {
 	var changedAppNames []types.AppName
 	var seen = make(map[types.AppName]bool)
+	//nolint:nilaway
 	for _, app := range changes.ChangedApps {
 		if _, ok := seen[app.App]; !ok {
 			seen[app.App] = true
@@ -843,8 +844,10 @@ func (s *State) GetAllEnvironmentConfigsFromDB(ctx context.Context, transaction 
 		return nil, fmt.Errorf("unable to retrieve manifests for environments %v from the database, error: %w", dbAllEnvs, err)
 	}
 	ret := make(map[types.EnvName]config.EnvironmentConfig)
-	for _, env := range *envs {
-		ret[env.Name] = env.Config
+	if envs != nil {
+		for _, env := range *envs {
+			ret[env.Name] = env.Config
+		}
 	}
 	return ret, nil
 }
--- a/services/cd-service/pkg/repository/repository.go
+++ b/services/cd-service/pkg/repository/repository.go
@@ -453,9 +453,11 @@ func MeasureGitSyncStatus(unsyncedApps, syncFailedApps int) error {
 }
 
 func (r *repository) notifyChangedApps(changes *TransformerResult) {
+	if changes == nil {
+		return
+	}
 	var changedAppNames []types.AppName
 	var seen = make(map[types.AppName]bool)
-	//nolint:nilaway
 	for _, app := range changes.ChangedApps {
 		if _, ok := seen[app.App]; !ok {
 			seen[app.App] = true
--- a/services/cd-service/pkg/service/batch.go
+++ b/services/cd-service/pkg/service/batch.go
@@ -551,7 +551,7 @@ func (d *BatchServer) ProcessBatch(
 		} else {
 			transformerTag = "multi"
 		}
-		parentSpan.SetTag("transformers", transformerTag)
+		parentSpan.SetTag("transformers", transformerTag) //nolint:nilaway
 	}
 
 	if requiresIsolation {
--- a/services/cd-service/pkg/service/overview.go
+++ b/services/cd-service/pkg/service/overview.go
@@ -499,7 +499,7 @@ func (o *OverviewServiceServer) getOverview(
 			var groupName = mapper.DeriveGroupName(config, envName)
 			var envInGroup = getEnvironmentInGroup(result.EnvironmentGroups, groupName, envName)
 			if envInGroup == nil {
-				return nil, fmt.Errorf("unable to find environment %s in group %s", envName, groupName)
+				continue
 			}
 
 			var argocd api.ArgoCDEnvironmentConfiguration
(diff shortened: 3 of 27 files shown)


services/reposerver-service/Makefile
```diff
@@ -19,3 +19,7 @@
 MIN_COVERAGE=48.0
 MAX_DOCKER_SIZE_MB=104
 include ../../infrastructure/make/go/include.mk
+
+sast:
+	@echo "Skip SAST checks for reposerver-service"
```
bqcuong, line 24: Running `nilaway` on reposerver and rollout services fails because there is not enough RAM on GitHub Action machine. Currently, we're using `ubuntu-latest` with [16GB RAM](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories), and running SAST inside Docker.

`nilaway` consumes significant amount of RAM when building the call/dependency graphs for `reposerver-service`. It's interesting that `reposerver-service` codebase is very small compared to other services such as `cd-service`, `manifest-repo-export-service`, but it might have much more code complexity.

The only solution for now is to disable SAST on `reposerver-service` and `rollout-service` for CI/CD pipeline. The warnings found for these two services are already addressed in this PR.

On local dev with a good machine, it's still possible to run SAST for these two services.

services/cd-service/pkg/repository/repository.go
```diff
@@ -455,6 +455,7 @@ func MeasureGitSyncStatus(unsyncedApps, syncFailedApps int) error {
 func (r *repository) notifyChangedApps(changes *TransformerResult) {
 	var changedAppNames []types.AppName
 	var seen = make(map[types.AppName]bool)
+	//nolint:nilaway
 	for _, app := range changes.ChangedApps {
```
ccreinartz11: ❓ Why not add a nil-check at the beginning of the function?
```go
if changes == nil {
   return
}
```

services/cd-service/pkg/service/batch.go
```diff
@@ -551,7 +551,7 @@ func (d *BatchServer) ProcessBatch(
 		} else {
 			transformerTag = "multi"
 		}
-		parentSpan.SetTag("transformers", transformerTag)
+		parentSpan.SetTag("transformers", transformerTag) //nolint:nilaway
```
ccreinartz11, line 554: ❓ Why is this required here? Can `nil` not be passed safely?
Also, it would be easy to fix without the nolint directive by moving `var transformerTag string` out of the `if len(transformers...` block

services/cd-service/pkg/service/overview.go
```diff
@@ -498,6 +498,9 @@ func (o *OverviewServiceServer) getOverview(
 		for envName, config := range envs {
 			var groupName = mapper.DeriveGroupName(config, envName)
 			var envInGroup = getEnvironmentInGroup(result.EnvironmentGroups, groupName, envName)
+			if envInGroup == nil {
+				return nil, fmt.Errorf("unable to find environment %s in group %s", envName, groupName)
+			}
```
ccreinartz11, line 503: ❌ This is a behavioral change, because now, if one group envInGroup is empty, we return rather than looping through the rest.
This entire for loop looks very weird to be honest but unless we know exactly what should happen here now we should not adjust logic as part of the NPE cleanup. 

Suggestion:
`continue` isntead of `return`

services/cd-service/pkg/service/overview.go
```diff
@@ -987,7 +990,7 @@ func CalculateWarnings(appDeployments map[types.EnvName]db.Deployment, appLocks
 				// if the env has no upstream, there's nothing to warn about
 				continue
 			}
-			upstreamEnvName := env.Config.GetUpstream().Environment
+			upstreamEnvName := env.Config.GetUpstream().Environment // nolint:nilaway
```
ccreinartz11: 🔧 Fix instead of // nolint directive.

services/cd-service/pkg/service/release_train_prognosis.go
```diff
@@ -98,7 +98,7 @@ func (s *ReleaseTrainPrognosisServer) GetReleaseTrainPrognosis(ctx context.Conte
 						},
 					}
 				}
-				retEnvPrognosis.GetAppsPrognoses().Prognoses[string(appName)] = retAppPrognosis
+				retEnvPrognosis.GetAppsPrognoses().Prognoses[string(appName)] = retAppPrognosis //nolint:nilaway
```
ccreinartz11: 🔧 fix instead of nolint

services/frontend-service/pkg/cmd/server.go
```diff
@@ -782,7 +782,7 @@ func (p *GrpcProxy) ProcessBatch(
 	ctx context.Context,
 	in *api.BatchRequest) (*api.BatchResponse, error) {
 	for i := range in.Actions {
-		batchAction := in.GetActions()[i]
+		batchAction := in.GetActions()[i] //nolint:nilaway
```
ccreinartz11: fix

services/rollout-service/pkg/argo/argo.go
```diff
@@ -348,6 +348,7 @@ func (a *ArgoAppProcessor) ProcessArgoOverview(ctx context.Context, l *zap.Logge
 	// bracket was last spec-updated against. Read it once up-front.
 	bracketSnapshotEslId := a.lookupBracketSnapshotEslId(ctx, l)
 	for _, currentApp := range sorting.SortKeys(argoOv.AppDetails) {
+		//nolint:nilaway
```
ccreinartz11, line 354: 🔧 
handle the argoOv == nil case once at the top of the function.

ccreinartz11 (changes requested):

services/cd-service/pkg/service/batch.go
bqcuong: Not the `transformerTag` but `parentSpan` is detected as nill-able here.
This warning is false positive because we already checked for `parentSpanExisted` (i.e., `parentSpan != nil`) before go to this line.

services/frontend-service/pkg/cmd/server.go
```diff
@@ -782,7 +782,11 @@ func (p *GrpcProxy) ProcessBatch(
 	ctx context.Context,
 	in *api.BatchRequest) (*api.BatchResponse, error) {
 	for i := range in.Actions {
-		batchAction := in.GetActions()[i] //nolint:nilaway
+		actions := in.GetActions()
+		if actions == nil {
+			continue
+		}
+		batchAction := actions[i]
```
ccreinartz11: ❓ Now that I look at that again.
We're already iteration through `in.Actions`. Why do we also need to call `in.GetActions`?
Why can't swe directly do
```go
for i, action := range in.GetActions() {
    batchAction := action
}
```

ccreinartz11 (approved):

PR #2986, opened 2026-08-18
fix(db): do not insert duplicates in app teams

sven-urbanski-freiheit-com:
The apps_teams_history table was used inefficiently. On every team change, it stored team changes additionally, instead of removing old entries, leading to a significant size increase over time.

Now duplicates are removed in the two functions that read from and write to that table.

Ref: SRX-HMADWG

### DIFF
--- a/pkg/db/db_apps.go
+++ b/pkg/db/db_apps.go
@@ -22,8 +22,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"slices"
-	"strings"
 	"time"
 
 	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
@@ -77,7 +75,7 @@ by reading the latest row created_at <= a given timestamp.
 */
 const appsTeamsHistoryTable = "apps_teams_history"
 
-type AppTeamMap = map[string][]types.AppName
+type TeamToAppsMap = map[string][]types.AppName
 
 // SELECTS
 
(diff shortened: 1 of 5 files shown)


pkg/db/db_team_apps.go
```diff
@@ -0,0 +1,81 @@
...
+		return appTeams
+	}
+	appTeams = RemoveAppFromTeams(appTeams, appName)
+	appTeams[teamName] = append(appTeams[teamName], appName)
+	return appTeams
+}
+
+// appsTeamsFromMap is the inverse of appsTeamsToMap: one entry per app, sorted by app name.
+func appsTeamsFromMap(appTeams AppTeamMap) []AppWithTeam {
+	result := make([]AppWithTeam, 0, len(appTeams))
```
bqcuong, line 58: The correct capacity of this array is: numTeams x numApps.

This array capacity is pre-allocated with number of teams (map key), which might be a much smaller value compare to correct capacity, which causes multiple time re-allocation.

pkg/db/db_apps.go
```diff
@@ -386,7 +388,7 @@ func (h *DBHandler) DBSelectLatestAppsTeamsHistory(ctx context.Context, transact
 	return h.processAppsTeamsRow(rows, err)
 }
 
-func (h *DBHandler) DBSelectAppsTeamsHistoryAtTimestamp(ctx context.Context, transaction *sql.Tx, ts time.Time) (_ []AppWithTeam, err error) {
+func (h *DBHandler) DBSelectAppsTeamsHistoryAtTimestamp(ctx context.Context, transaction *sql.Tx, ts time.Time) (_ []AppWithTeam, _ AppTeamMap, err error) {
```
bqcuong: The returned value `_ []AppWithTeam` seems not be used anymore, could we remove it?

bqcuong (approved):

pkg/db/db_team_apps.go
sven-urbanski-freiheit-com: It's true that it needs to re-allocated, but the actual capacity its not `numTeams x numApps`. Each team can have 1-n apps.

pkg/db/db_apps.go
sven-urbanski-freiheit-com: done

PR #2991, opened 2026-08-25
feat: make sql migration timeout configurable

sven-urbanski-freiheit-com:
This allows operators to increase the timeout for sql migrations.

Ref: SRX-7EX4MK


charts/kuberpult/values.yaml
```diff
@@ -152,6 +152,12 @@ db:
   # SSL mode to be used by the database.
   # For all possible values see https://jdbc.postgresql.org/documentation/ssl/
   sslMode: verify-full
+  migrations:
+    # Note that this timeout is for ONE sql migration.
+    # This timeout is used during startup of the cd- and manifest-repo-export-service, in order to run the go-migrate based
+    # SQL migrations.
+    # If this is increased, you also have to increase the k8s probes of these 2 services.
+    timeoutPerFile: 10s
```
ccreinartz11, line 160: 🤔 Let's discuss the part with the k8s probes.

pkg/db/db.go
```diff
@@ -62,6 +63,13 @@ type DBConfig struct {
 	DatadogServiceName string
 }
 
+func (db *DBConfig) getMigrationsTimeout() time.Duration {
+	if db.MigrationsTimeout == nil {
+		return time.Second * 10
+	}
+	return *db.MigrationsTimeout
+}
```
ccreinartz11, line 71: 🤔 This does not seem to be the right place for a default. 
Set the default in the environment configuration or fail if this is not set.

ccreinartz11 (changes requested):

ccreinartz11 (approved):

PR #2997, opened 2026-08-27
fix: query to get brackets in manifest-export

sven-urbanski-freiheit-com:
This bug only appears when using brackets in the manifest-export, in helm that is if manifestRepoExport.rendering.experimentalRenderBrackets=true.

The manifest-repo-export-service used a wrong query to get the bracket history. It insisted on one specific row while it should also allow any row before that.

Note that the repo-server uses the same query. However, there it is correct, because the repo-server has the precise esl ID to query exactly.

This also fixes the issue that the manifest-export attempted to render brackets, despite pointToBrackets==false. This issue happens when an app already got a bracket and pointToBrackets==false.

Ref: SRX-7EX4MK


pkg/db/db_brackets_history.go
```diff
@@ -236,6 +236,23 @@ func DBSelectBracketHistoryPrevious(ctx context.Context, h *DBHandler, tx *sql.T
 	return executeSelectQuery(ctx, tx, selectQuery)
 }
 
+func DBSelectBracketHistoryAtOrBeforeId(ctx context.Context, h *DBHandler, tx *sql.Tx, maxEslVersion TransformerID) (result *BracketRow, err error) {
```
ccreinartz11, line 242: :nit: exported function should have a comment.

services/manifest-repo-export-service/pkg/repository/repository.go
```diff
@@ -1124,9 +1126,16 @@ func collectArgoAppDataForEnv(
...
+	var allBrackets *db.BracketRow
+	if pointToBrackets {
+		// Bracket names are only relevant for rendering if pointToBracket==true.
+		// By only getting the brackets conditionally, we prevent CalculateAppDataWithBrackets
+		// from accidentally using bracket data when it's not wanted.
+		// This is only an issue if pointsToBracket==false && there are bracket apps.
+		allBrackets, err = db.DBSelectBracketHistoryAtOrBeforeId(ctx, dbHandler, transaction, eslVersion)
+		if err != nil {
+			return nil, fmt.Errorf("could not find bracket at %v: %w", eslVersion, err)
+		}
```
ccreinartz11: 🤔 I don't fully get it. Is it a performance improvement or a fix for a rendering bug? 
If the latter, why is it not part of the rendering function?

sven-urbanski-freiheit-com: it's a rendering bugfix.

I put it here, because loading DBSelectBracketHistoryAtOrBeforeId is pointless when we don't even render brackets.

sven-urbanski-freiheit-com: But I think you are right, I will move the check to where it actually matters.

sven-urbanski-freiheit-com: The issue was that if
pointToBrackets==false && bracketMap had entries
then it would switch to rendering apps instead of brackets.

This is now fixed by checking "if !pointsToBrackets"

ccreinartz11 (approved):

PR #2998, opened 2026-09-01
fix: team name rendering in manifest-export

sven-urbanski-freiheit-com:
This fixes an issue where the manifest-export rendered a label that has a value starting with "_" which is invalid according to kubernetes.

Ref: SRX-UG5WLO

### DIFF
--- a/pkg/valid/validations.go
+++ b/pkg/valid/validations.go
@@ -42,6 +42,9 @@ const (
 	ArgoBracketNameRegExp    = AppNameRegExp
 	MaxArgoBracketNameLength = 63
 	MinArgoBracketNameLength = 1
+
+	MaxKubernetesLabelValueLength = 63
+	KubernetesLabelValueRegExp    = `\A[A-Za-z0-9]([-A-Za-z0-9_.]*[A-Za-z0-9])?\z`
 )
 
 var (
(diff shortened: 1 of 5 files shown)


pkg/valid/validations.go
```diff
@@ -52,6 +55,8 @@ var (
 	ArgoBracketNameRx = regexp.MustCompile(ArgoBracketNameRegExp)
 	commitIDPrefixRx  = regexp.MustCompile(commitIDPrefixRegExp)
 	MaxAppNameLen     = setupMaxAppNameLen()
+
+	kubernetesLabelValueRx = regexp.MustCompile(KubernetesLabelValueRegExp)
```
ccreinartz11, line 59: 🤔 Usually we avoid regexp in our projects. But I see that we're already doing it in multple places here, so at least this does not establish a new precedent.

services/manifest-repo-export-service/pkg/argocd/render_test.go
```diff
@@ -758,11 +749,10 @@ metadata:
     com.freiheit.kuberpult/aa-parent-environment: test-env
     com.freiheit.kuberpult/application: app1
     com.freiheit.kuberpult/environment: test-env
-    com.freiheit.kuberpult/teams: team1_team2
   finalizers:
   - resources-finalizer.argocd.argoproj.io
   labels:
-    com.freiheit.kuberpult/teams: team1_team2
+    com.freiheit.kuberpult/team: team1_team2
```
ccreinartz11: ❓Are there any potential side effects to this change?
Also, `teams` seems to be more precise, especially if it can contain multiple teams.
Not sure why you changed this.

ccreinartz11 (changes requested):

services/manifest-repo-export-service/pkg/argocd/render_test.go
sven-urbanski-freiheit-com: I reverted this part

ccreinartz11 (approved):

PR #3001, opened 2026-09-03
fix: optimize DBSelectAppsWithDeploymentInEnvAtTimestamp

sven-urbanski-freiheit-com:
This query plays an important role in determining "what apps will be deployed".
Now we only query the absolutely required columns, so the json parsing is also gone.

This change optimizes the query and adds an appropriate index on top for the table deployments_history.

Ref: SRX-LJHTO5

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -24,6 +24,9 @@ import (
 	"strings"
 )
 
+type ReleaseVersion = *uint64
+type Revision = uint64
+
 type GitTag string
 
 // EnvName is a type that helps us avoid mixing up envNames from other strings.
@@ -67,7 +70,7 @@ type ManifestLockID int64
 
 type ArgoProjectName string
 
-func EnvNamesToStrings(a []EnvName) []string {
+func NamesToStrings[T ~string](a []T) []string {
 	var result = make([]string, len(a))
 	for i := range a {
 		result[i] = string(a[i])
@@ -75,18 +78,26 @@ func EnvNamesToStrings(a []EnvName) []string {
 	return result
 }
 
-func StringsToEnvNames(a []string) []EnvName {
-	var result = make([]EnvName, len(a))
+func StringsToNames[T ~string](a []string) []T {
+	var result = make([]T, len(a))
 	for i := range a {
-		result[i] = EnvName(a[i])
+		result[i] = T(a[i])
 	}
 	return result
 }
 
-func Sort(a []EnvName) []EnvName {
-	s := EnvNamesToStrings(a)
+func EnvNamesToStrings(a []EnvName) []string {
+	return NamesToStrings(a)
+}
+
+func StringsToEnvNames(a []string) []EnvName {
+	return StringsToNames[EnvName](a)
+}
+
+func Sort[T ~string](a []T) []T {
+	s := NamesToStrings(a)
 	sort.Strings(s)
-	return StringsToEnvNames(s)
+	return StringsToNames[T](s)
 }
 
 func StringPtr(a EnvName) *string {
--- a/services/manifest-repo-export-service/pkg/db_history/db_history.go
+++ b/services/manifest-repo-export-service/pkg/db_history/db_history.go
@@ -19,101 +19,97 @@ package db_history
 import (
 	"context"
 	"database/sql"
-	"encoding/json"
-	"errors"
 	"fmt"
 	"time"
 
-	"go.uber.org/zap"
+	"github.com/lib/pq"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 
-	"github.com/freiheit-com/kuberpult/pkg/db"
-	"github.com/freiheit-com/kuberpult/pkg/logging"
 	"github.com/freiheit-com/kuberpult/pkg/types"
 )
 
-type DeploymentMap map[types.AppName]db.Deployment
+type DeploymentShort struct {
+	ReleaseVersion types.ReleaseVersion
+	Revision       types.Revision
+}
+type DeploymentMap map[types.AppName]DeploymentShort
 
 // DBSelectAppsWithDeploymentInEnvAtTimestamp returns all apps that had a deployment in the given env at the given timestamp:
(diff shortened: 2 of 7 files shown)


services/manifest-repo-export-service/pkg/db_history/db_history.go
```diff
@@ -19,52 +19,56 @@ package db_history
...
 	"github.com/freiheit-com/kuberpult/pkg/logging"
 	"github.com/freiheit-com/kuberpult/pkg/types"
 )
 
-type DeploymentMap map[types.AppName]db.Deployment
+type DeploymentShort struct {
+	ReleaseVersion *uint64
+	Revision       uint64
+}
+type DeploymentMap map[types.AppName]DeploymentShort
```
sven-urbanski-freiheit-com, line 35: note: the old `DeploymentMap` contained several fields that were completely unused, most notably the metadata that required json parsing (the json is small, but we have many rows of it)

pkg/types/types.go
```diff
@@ -24,6 +24,9 @@ import (
 	"strings"
 )
 
+type ReleaseVersion = *uint64
```
ccreinartz11: 🔧 Do not alias a pointer type. 
1. It's easy to miss nil-checks because now it's no longer obvious that `ReleaseVersion` is a pointer because it's hidden by the type.
2. You cannot define methods on pointer types or pointer type aliases.

pkg/types/types.go
```diff
@@ -67,26 +70,34 @@ type ManifestLockID int64
...
 
-func Sort(a []EnvName) []EnvName {
-	s := EnvNamesToStrings(a)
+func EnvNamesToStrings(a []EnvName) []string {
+	return NamesToStrings(a)
+}
+
+func StringsToEnvNames(a []string) []EnvName {
+	return StringsToNames[EnvName](a)
+}
```
ccreinartz11: 🤔 Not sure that you still need / want these now that you have the generic. 
The point of the generic should be that you no longer need to write these type-specific functions.

pkg/types/types.go
```diff
@@ -67,26 +70,34 @@ type ManifestLockID int64
...
 	var result = make([]string, len(a))
 	for i := range a {
 		result[i] = string(a[i])
 	}
 	return result
 }
 
-func StringsToEnvNames(a []string) []EnvName {
-	var result = make([]EnvName, len(a))
+func StringsToNames[T ~string](a []string) []T {
```
ccreinartz11: :nit: `StringsToNames` is no longer a good name. 
Maybe `ConvertToStrings` and `ConvertFromStrings` (or `Cast` instead of `Convert`)

services/manifest-repo-export-service/pkg/db_history/db_history.go
```diff
@@ -19,101 +19,97 @@ package db_history
...
-	"github.com/freiheit-com/kuberpult/pkg/db"
-	"github.com/freiheit-com/kuberpult/pkg/logging"
 	"github.com/freiheit-com/kuberpult/pkg/types"
 )
 
-type DeploymentMap map[types.AppName]db.Deployment
+type DeploymentShort struct {
+	ReleaseVersion types.ReleaseVersion
+	Revision       types.Revision
+}
```
ccreinartz11: :nit: `DeploymentShort` is not clear. 
Maybe `DeploymentReleaseInfo`?

services/manifest-repo-export-service/pkg/db_history/db_history.go
```diff
@@ -19,101 +19,97 @@ package db_history
...
+	span.SetTag("kuberpultEnvironment", envSelector)
+	span.SetTag("numApps", len(appNames))
+	selectQuery := `
+		 SELECT
+			  a.appname,
+			  d.releaseversion,
+			  d.revision
+		  FROM unnest($1::text[]) AS a(appname)  -- "unnest" simply converts our slice into a row
+		--  For manually using this query in psql, replace this line with:
+		--  FROM unnest((SELECT array_agg(appname::text) FROM apps)) AS a(appname)
```
ccreinartz11, line 52: ❓ Why is the psql query different?

services/manifest-repo-export-service/pkg/db_history/db_history.go
```diff
@@ -19,101 +19,97 @@ package db_history
...
+
+		--- Then we join lateral ("subquery") in order to get the latest deployment.
+	    --- This will skip apps that have now deployment - that's ok, we cannot deploy those anyway!
```
ccreinartz11:
```suggestion
	    --- This will skip apps that have no deployment - that's ok, we cannot deploy those anyway!
```

services/manifest-repo-export-service/pkg/db_history/db_history.go
```diff
@@ -19,101 +19,97 @@ package db_history
...
+		  JOIN LATERAL (
+			  SELECT
+				  releaseversion,
+				  revision
+			  FROM deployments_history
+			  WHERE appname = a.appname
+				  AND envname = $2
+				  AND created <= $3
+			  -- Note that we cannot filter "releaseVersion IS NULL" here, even though we later filter out null deployments,
+			  -- because that would not give us the LATEST deployment.
```
ccreinartz11, line 65: ❓ I don't get this. Let's discuss.

ccreinartz11 (changes requested):

services/manifest-repo-export-service/pkg/db_history/db_history.go
sven-urbanski-freiheit-com: This is just for convenience, if you want to try out the query in cloud sql.
I'm also fine with removing it.

services/manifest-repo-export-service/pkg/db_history/db_history.go
sven-urbanski-freiheit-com: releaseVersion=NULL carries a specific meaning that "there is no release". If there is no release, we cannot just use another release instead, we have to keep this "no-release" information.
This is because the table carries all historic deployments.

If the current releaseVersion is actually null, and we would skip NULLS right here, we would get the next release that isn't null, but that would be wrong.

ccreinartz11 (approved):

PR #3004, opened 2026-09-18
fix: do not panic when valid jwts do not include name and email fields

ccreinartz11:
Ref: SRX-LBGOID


services/frontend-service/pkg/interceptors/interceptors.go
```diff
@@ -57,10 +57,12 @@ func authorize(ctx context.Context, jwks *keyfunc.JWKS, clientId string, tenantI
...
 	if _, ok := claims["aud"]; ok && claims["aud"] == clientId {
+		email, _ := claims["email"].(string)
+		name, _ := claims["name"].(string)
 		u = &auth.User{
 			DexAuthContext: nil,
-			Email:          claims["email"].(string),
-			Name:           claims["name"].(string),
+			Email:          email,
+			Name:           name,
 		}
```
ccreinartz11, line 66: ❓ Does this make sense in general if email / name are empty or do we fail in this case? 
If this is a failure condition, it should go into validation.

sven-urbanski-freiheit-com: We already have some logic around this elsewhere:
* If the name is empty, that's accepted, and we use the email as name.
* If both are empty, then we have no identifiable string, so counting this as a success would be wrong. If we allow both empty, then we essentially do not know the user. If there is neither name nor email, then we need to act the same as if there is no jwt.

ccreinartz11: If the name is empty, that's accepted, and we use the email as name.
-> Why? we already have email in email. why not just leave 'name' empty then?
Also, I checked the code and the behaviour you describe is not consistently applied

sven-urbanski-freiheit-com (approved):

PR #3005, opened 2026-09-21
feat: add devcontainer

ccreinartz11:
Ref: SRX-LBGOID


.devcontainer/load-env.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+# Loads this project's environment variables into the CURRENT shell's
+# environment, so devcontainer.json's ${localEnv:...} substitutions (in
+# "remoteEnv") see them when you run `devcontainer up` / `devcontainer exec`
+# next.
+#
+# Two files are loaded, in order:
+#   - project.env, alongside this script: non-secret, project-scoped defaults
+#     (e.g. the Revolution workspace/product IDs). Committed to the repo, so
+#     every checkout gets the same values without per-user setup.
```
sven-urbanski-freiheit-com, line 10: sounds good but the committed file contains "PROJECT_SECRETS_FILE="${HOME}/project-secrets/kuberpult/secrets.env.local", that sounds wrong to me, because it's based on absolute path that is different on our machines. Shouldn't that be part of the per-user file?

.devcontainer/mise.toml
```diff
@@ -0,0 +1,21 @@
+# The development toolchain of the devcontainer.
+[tools]
```
sven-urbanski-freiheit-com, line 3: we need at least a comment here that these versions need to be aligned with nix:

```suggestion
# The versions used here should always be aligned with the version in flake.nix
[tools]
```

.devcontainer/project.env
```diff
@@ -0,0 +1,12 @@
+# Project-scoped, non-secret environment variables -- committed to the repo,
+# unlike the per-user secrets file (see load-env.sh). Loaded automatically by
+# load-env.sh alongside that file.
+
+# Copying .devcontainer/ into a new project? Change this to that project's
+# own path. It's the per-user secrets file load-env.sh reads -- left as-is, a
```
sven-urbanski-freiheit-com, line 6: does this have to be an absolute path?

.devcontainer/load-env.sh
ccreinartz11: That's part of the readme. I would actually argue that it's fine if all members of a project place secrets in uniform locations. It's more uniform.
Having this file at all is bad though and has to be removed in the future. For now, I'd leave it there.
We can improve it later.

.devcontainer/project.env
ccreinartz11: it does not have to be but I think an absolute path is preferable. we don't want the path to be part of the repo and we don't want it to break if we move the repo somewhere else in the file system.

.devcontainer/README.md
```diff
@@ -0,0 +1,87 @@
...
+If you find limitations while working with this setup, please open a PR that documents them
+briefly in this section of the readme.
+
+### Agent access to secrets
+Since the secrets are available in the container as environment variables, a running 
+AI agent has access to them. 
+Keep this in mind when deciding which secrets should be exposed within the container.
+Ideally, only expose secrets with minimal blast radius and short lifespans.
+
+Status: The SRE team is actively working on a solution for this.
```
sven-urbanski-freiheit-com, line 66: Status here is not relevant

sven-urbanski-freiheit-com (approved):

PR #3006, opened 2026-09-23
fix: optimize manifest-exports deployment transformer

sven-urbanski-freiheit-com:
We already load the deployments in the releasetrain (and CreateApplicationVersion), so there is no reason to load it in the Deployment Transformer again. We only have to load it for the case that an individual Deployment is written to the esl events, because in this case the transformer is called without the "DeploymentPreloaded" field.

The loading (DBSelectLatestDeployment) was also wrong, because it looked at latest, not at the timestamp. It was unlikely to cause real harm, but still wrong.

Also removed the Cleanup, since it is not required anymore, as well as the "queuedVersionFile" which also does not exist anymore. Both of these were handling files in the manifest repo that have not been needed in a long time.

This improvement is particularly useful for release trains with many deployments on the manifest-export service.

Ref: SRX-LJHTO5


ccreinartz11 (approved):

pkg/db/db_deployments.go
```diff
@@ -478,6 +478,28 @@ func (h *DBHandler) DBSelectDeploymentsByTransformerID(ctx context.Context, tx *
...
+		transformerID,
+		appSelector,
+		envSelector,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("could not select deployment for app %s on env %s for transformer %v from DB. Error: %w", appSelector, envSelector, transformerID, err)
+	}
+	defer closeRowsAndLog(rows, ctx, "DBSelectDeploymentByTransformerIDAppAndEnv")
+	return processDeployment(rows)
+}
```
ccreinartz11: ❓ How does `deploymentPreloaded` get fetched from the database? Could we not use the same query that we use for the preloaded deployment  again? 
A bit confused as to why we need a new query here.

